PR #55, opened 2016-01-19
Nuclear decay behaves not as expected

lukasmerten:
The current implementation of the relativistic gamma factor in the decay rate leads to a strange behavior. Particles with higher energies and therefore higher gamma factors decay faster. The opposite behavior is expected, e.g atmospheric myons which can be detected at the Earth's surface.
Or do I see something terribly wrong? 

I created a python script notebook where I tried to explain this problem (see below). 
[Decay.txt](https://github.com/CRPropa/CRPropa3/files/95931/Decay.txt)

This pull request leads to the expected behavior.


DavidWalz:
Added fix manually because one of the 4 commits fails in travis.
Thanks for finding!!

PR #59, opened 2016-02-01
Wrong unit comparison in PropagationCK::process

lukasmerten:
bug fix: h [time] was compared to minStep [length]. Fixed units.


DavidWalz:
Thanks for catching!

PR #65, opened 2016-03-18
Creation point of secondaries and PhotonOutputThreshold1D module

cheiter:
Set creation point of secondaries uniformly distributed within propagation stepsize bin.
Added PhotonOutput1D module with energy threshold condition.


geromueller:
Hi,

i have 3 remarks:
1. please check indentation
2. extract the randomPoisiton function to the Random class as generic function to interpolate between two vectors. Random::interpolate(const Vector3 &a, const Vector3 &b)
3. Why another outputmodule? as far as i can see, the existing ones could be used!?

PR #81, opened 2016-06-02
Bug fix

lukasmerten:
The calculation of the poloidal component of the JF12 field lead to "nan" values at the z-axis. Since rXc/rc = rp/r the code is mathematically the same but  the computer is not forced to calculate the quotient of 0./0. at the z-axis.


DavidWalz:
In the constant elevation region rp/r is not the same as rXc/rc, or am I mistaken?

lukasmerten:
That's  correct David. I fixed it.

DavidWalz:
Ok merged. Thanks for the bug fix!

PR #82, opened 2016-06-02
Fixes in SynchrotronRadition and NuclearDecay modules

cheiter:
- Added missing boost to lab frame for photon energy in NuclearDecay module.
- Added set and get functions to define an energy threshold for production of synchrotron photons


DavidWalz:
Thanks! Merged with 032b290403da2b0d9cb0c152e51f2b041bbf0bbe

PR #119, opened 2017-04-11
Add source class: SourceSNRDistribution

lukasmerten:
Add a simple source class for drawing source positions from the Galactic SNR distribution. For details of the distribution see G. Case and D. Bhattacharya, _Revisiting the galactic supernova remnant distribution_, Astron.& Astroph. Suppl. **120** (Dec., 1996)

### DIFF
--- a/src/Source.cpp
+++ b/src/Source.cpp
@@ -335,6 +335,87 @@ void SourceUniformCylinder::setDescription() {
 	description = ss.str();
 }
 
+// ---------------------------------------------------------------------------
+SourceSNRDistribution::SourceSNRDistribution() :
+    R_earth(8.5*kpc), beta(3.53), Zg(0.3*kpc) {
+	set_frMax(8.5*kpc, 3.53);
+	set_fzMax(0.3*kpc);
+}
+
+SourceSNRDistribution::SourceSNRDistribution(double R_earth, double beta, double Zg) :
+    R_earth(R_earth), beta(beta), Zg(Zg) {
+	set_frMax(R_earth, beta);
+	set_fzMax(Zg);
+}
+
+void SourceSNRDistribution::prepareParticle(ParticleState& particle) const {
+  	Random &random = Random::instance();
+	double RPos;
+	while (true){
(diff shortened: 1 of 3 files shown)


src/Source.cpp
```diff
@@ -335,6 +335,87 @@ void SourceUniformCylinder::setDescription() {
...
+    R_earth(R_earth), beta(beta), Zg(Zg) {
+	set_frMax(R_earth, beta);
+	set_fzMax(Zg);
+}
+
+void SourceSNRDistribution::prepareParticle(ParticleState& particle) const {
+  	Random &random = Random::instance();
+	double RPos;
+	while (true){
+		RPos = random.rand()*20.*kpc;
```
adundovi: Is there any reason why R_max is hard-coded to 20*kpc? At least, for clarity, I would put it in the header file and here would put R_max. The same comment goes for ZPos a few lines below.

src/Source.cpp
```diff
@@ -335,6 +335,87 @@ void SourceUniformCylinder::setDescription() {
...
+
+double SourceSNRDistribution::get_frMax() {
+	return frMax;
+}
+
+double SourceSNRDistribution::get_fzMax() {
+	return fzMax;
+}
+
+void SourceSNRDistribution::f_r(double r, double &fr) const{
```
DavidWalz: Why not return fr(r) by value? Then its easier to use it from python. Same for f_z.
```
double SourceSNRDistribution::f_r(double r) const{
    ...
    return fr
}
```

src/Source.cpp
lukasmerten: There is no physics reason. It is only for efficiency using the JF12 field which is cut to zero at R=20 kpc. Probably it is better to just set the maximum values by default and make them variable. In that way the source class could be reused for other magnetic fields. 
I will fix that tomorrow.

src/Source.cpp
lukasmerten: When this is not a problem for the thread safeness we can change the behaviour of fr(r) and fz(z).

PR #122, opened 2017-04-27
SourcePulsarDistribution

lukasmerten:
A new source class using the Pulsar distribution given in Faucher‐Giguere, Claude‐Andre
Kaspi, Victoria M., ApJ **643** (2012) 332-355.

If anyone has a good idea for a meaningful test case I will be happy to include it.

### DIFF
--- a/include/crpropa/Source.h
+++ b/include/crpropa/Source.h
@@ -264,38 +264,52 @@ public:
 };
 /**
 @class SourcePulsurDistribution
-@brief Source distribution following the Galactic pulsur distribution
+@brief Source distribution following the Galactic pulsar distribution
 
 Four major arms are used. The details of the distribution can be found in
 Blasi (2012)
 */
-
-class SourcePulsurDistribution: public SourceFeature {
+const double r_0[] = {3.48, 3.48, 4.9, 4.9};
+const double theta_0[] = {0., 3.14, 2.52, -0.62};
+const double k_0[] = {4.25, 4.25, 4.89, 4.89};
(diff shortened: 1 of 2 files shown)


lukasmerten:
pulsardistribution

That is what the distribution looks like for 100,000 source positions.

adundovi:
I don't think that test for this distribution model is necessary. It is provided "as is" and directly implemented from the paper (btw the year in your comment above should be 2006, right?).

I am only concerned with the code organization: maybe we should put the code of sources in separate folder and put them in independent files if we continue to add source distributions like this. But that would be a future task.

lukasmerten:
Yes , 2006 is the correct year. I changed that in case anyone is looking into this conversation. 

A restructuring of the Source code sounds like a good idea.

DavidWalz:
Nice plot!
Andrej what do you have in mind?

adundovi:
Source.cpp has almost 1k lines, and it is becoming hard to manage. I also have several source features of my own, and if I include them in the main branch (which I am planning to do at some point), it will be definitely overcrowded. Maybe categorizing source features in separate files based on their role in the source object, for example, if a feature is setting initial positions it would be SourcePositionFeatures, for initial directions SourceDirectionFeatures, and so on (composition, energy). That would be logic which is easy to follow, right?

DavidWalz:
Simply moving stuff between source files doesn't affect users at all, so we can easily change back and forth. Since you have a number of additional features you want to commit at some point, you could in the process see whatever suits best. If you split into multiple files I wouldn't mind a subfolder like for magnetic fields.

PR #145, opened 2017-11-05
add weights

rafaelab:
Add weights as Candidate property.
Also adds the weights to the output when setEnableWeights(True) is called.


rafaelab:
Weights included as Candidate property.
They *do not* show up in the default output (only with enableAll) as most applications don't require them. 
See [this commit](https://github.com/CRPropa/CRPropa3/commit/73ee7fb860de681300ce2bcabcb7010b58b7a48b).

PR #169, opened 2018-02-01
Allow or parallelization of less than 1000 candidates.

lukasmerten:
No significant performance loss was found for simulation with many short trajectories.

### DIFF
--- a/src/ModuleList.cpp
+++ b/src/ModuleList.cpp
@@ -111,7 +111,7 @@ void ModuleList::run(candidate_vector_t &candidates, bool recursive, bool second
 	sighandler_t old_sigterm_handler = ::signal(SIGTERM,
 			g_cancel_signal_callback);
 
-#pragma omp parallel for schedule(static, 1000)
+#pragma omp parallel for schedule(static, 1)
 	for (size_t i = 0; i < count; i++) {
 		if (g_cancel_signal_flag != 0)
 			continue;
@@ -153,7 +153,7 @@ void ModuleList::run(SourceInterface *source, size_t count, bool recursive, bool
 	sighandler_t old_sigterm_handler = ::signal(SIGTERM,
 			g_cancel_signal_callback);
 
-#pragma omp parallel for schedule(static, 1000)
+#pragma omp parallel for schedule(static, 1)
 	for (size_t i = 0; i < count; i++) {
 		if (g_cancel_signal_flag)
 			continue;


adundovi:
For example:
```
#if _OPENMP
#include <omp.h>
#define OPM_SCHEDULE_KIND static
#define OPM_SCHEDULE_CHUNKS 1
#endif
```
#pragma omp parallel for schedule(OPM_SCHEDULE_KIND, OPM_SCHEDULE_CHUNKS)

lukasmerten:
Sounds good to me.
We should then maybe also but some information (in the wiki) on parallelization with the above link. So people know how and when to change the default OMP parameters.

adundovi:
Agree. Maybe this one too:
https://software.intel.com/en-us/articles/openmp-loop-scheduling

adundovi:
BTW when I read the description of these OPM methods I agree that 1000 is maybe too much, but 1 could introduce problems with strong magnetic fields where particles can be trapped and their propagation prolonged in the context of CPU time, which means that the static schedule will not be efficient any more if there is no "averaging" - maybe 10 would work as a compromise, i.e., for 10 particles their joint execution time will be somehow similar?

adundovi:
Of course, I didn't test this proposal so if performance tests are showing a different outcome, I'll go with what is the best in practice.

geromueller:
Hi,

the scheduler is overwritten and configurable from commandline:
http://docs.cray.com/books/S-2179-52/html-S-2179-52/z1050587246oswald.html

Gero

Am 01.02.2018 um 22:24 schrieb Andrej Dundovic:

lukasmerten:
For very strong magnetic fields you might be right.
And a chunksize of 10 should not restrict most of the simulations.

In case of very different trajectories even a "dynamic" scheduler might be best.

lukasmerten:
@geromueller: But the commandline overwrite works only for schedule type 'runtime', don't they? At least I do not see an impact if I write:
export OMP_SCHEDULE="static, 500" 
before starting a crpropa simulation

adundovi:
As Travis here, I also have segfaults locally if I go below 100 - don't know why.
```
crpropa::ModuleList: Number of Threads: 4
Run ModuleList
*** Error in `./testModuleList': double free or corruption (fasttop): 0x0000000000e0bfb0 ***
Segmentation fault (core dumped)
```

adundovi:
> OMP_SCHEDULE applies only to for and parallel for directives that have the schedule type runtime.
http://docs.cray.com/books/S-2179-52/html-S-2179-52/z1050587246oswald.html

lukasmerten:
So what about setting the schedule type to runtime?
`#pragma omp parallel for schedule(runtime)`

This allows then to specify the schedule type on runtime via:
`export OMP_SCHEDULE="TYPE,CHUNKSIZE"`

And, at least for me, falls back to `static` if no enviroment variable is defined?

lukasmerten:
Also I cannot reproduce the Travis error

adundovi:
Unfortunately, I think we cannot merge this until we resolve the seqfault problem. I've tested it on my laptop (Fedora 27, gcc 7.2.1, kernel 4.14.13) and it does not work with gcc, but it works with clang (!). It crashes for "runtime" or for "static, <100". What Linux distribution do you use and which version?

BTW I suggest that we use ```#define OMP_SCHEDULE static,1000``` instead of two variables that I suggested above because one can change this to ```runtime``` without editing the code below (where the second variable should be removed since the runtime cannot have chunks).

lukasmerten:
I agree with you that we should not merge a failing pull request.
But I thought this is a good way to test the implementation since it produces not error on my system:
* Debian: 8.10
* gcc: 4.9.2-10
* Kernel: 3.16.0-4-amd64

I will try it also test it on other systems.

Your suggestion to implement `define OMP_SCHEDULE static,1000` is a good a idea if we cannot make `runtime` work.

lukasmerten:
Now also checked and is working on:
* Debian: 9.3
* gcc: 6.3.0
* Kernel: 4.9.0.4-amd64

TobiasWinchen:
Could you merge the recent changes I made on master to the ModuleList? These might solve this issue.

lukasmerten:
With a rebase on the latest master it is failing now with the same error for me too.

adundovi:
Problem was in my commit bf5205144c64034390433cb76c1ad3ab47ecf385.

PR #187, opened 2018-05-07
Geometry system

TobiasWinchen:
I added a small geometry system that is mainly useful for the acceleration simulations, but can also be used in other parts of CRPropa.  I introduced an abstract base class `Surface` with implementations `Sphere`, `Plane` and `ParaxialBox`. Each `Surface` has a method `distance` that returns the closest distance of a point to the surface. For bounded surfaces the distance is
negative on the inside.

This is used in a new Module `RestrictToRegion` that restricts module executions to certain regions e.g.
```
sim.add( RestricToRegion( SecondOrderFermiAcceleration(), Sphere( Vector3d(0,0,0), 10) ))
```
This is used for the implementation of the acceleration simulation.

I also introduced a new `ObserverFeature`, `ObserverSurface`, that detects a candidate crossing the surface. It can be used e.g. as follows:
```
obs = Observer()
obs.add( ObserverSurface( Plane( Vector3d(0, 0, 0), Vector3d(0, 0, 1) ) ))
```
to define an observer in the xy plane (normal vector 0,0,1).

This potentially deprecates the `ObserverSmallSphere` and `ObserverLargeSphere` as
```
obs.add( ObserverSurface( Sphere( Vector3d(0,0,0), 10 )))
```

can be used for both depending on the source position. I already added according deprecation warnings. Also the tracking observer can be updated in the future to work with arbitrary geometries.

### DIFF
--- a/include/crpropa/Geometry.h
+++ b/include/crpropa/Geometry.h
@@ -1,0 +1,99 @@
+#ifndef GEOMETRY_H
+#define GEOMETRY_H
+
+#include <vector>
+
+#include "crpropa/Candidate.h"
+#include "crpropa/Vector3.h"
+
+namespace crpropa
+{
+/**
+ * \addtogroup Core
+ * @{
+ */
+
+/**
+ @class Surface
+ @brief A geometrical surface
+
+ Defines a surface. Can be queried if the candidate has crossed the surface in the last step.
+ */
+
+class Surface
+{
+	public:
+
+	/**
+		Returns the distance of a point to the surface. Negative on the one side,
+		positive on the other. For closed surfaces it is negative on the inside.
+	 */
+    virtual double distance(const Vector3d& point) const = 0;
+};
+
+/**
+ @class Plane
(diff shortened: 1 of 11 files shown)


include/crpropa/Geometry.h
```diff
@@ -0,0 +1,105 @@
...
+		Vector3d center;
+		double radius;
+	public:
+		Sphere(const Vector3d& _center, double _radius);
+    virtual double distance(const Vector3d &point) const;
+		virtual std::string getDescription() const;
+};
+
+
+///**
```
lukasmerten: Could you exclude the arbitrary box until its ready to prevent confusion?

include/crpropa/Geometry.h
```diff
@@ -0,0 +1,105 @@
...
+//		std::vector<Plane> facets;
+//		double A,B,C;
+//	public:
+//		Box(const Vector3d& _corner, const Vector3d& _x1,const Vector3d& _x2, const Vector3d& _x3);
+//    virtual double distance(const Vector3d &point) const;
+//};
+
+
+/**
+ @class ParaxialBox
```
lukasmerten: Description should be more specific on the ParaxialBox

Issue #190, opened 2018-05-29
Store/display random number seeds by default


adundovi:
Due to easier debugging, i.e., reproducing possible strange behaviours in CRPropa we want to store or display the random number seed(s) so that users can immediately report them to the issue tracker together with their steering cards and the produced output.

TobiasWinchen:
I briefly looked into this. Technically this is no problem. However, the seed of the RNG is an array of 625 uint32, i.e. ~2.5 kb of data - per thread. It is for sure no problem to store this per default in the hdf files, but we probably don't want to store NThreads * 42 lines of random characters in the TextOutput. Also displaying this by default is not an option, so I propose to store this in the hdf files and display it only when an error occurs. Alternatively, we could possibly also reduce the space of possible random sequences and shorten the seed to 127 bit per thread, thus moving to [TinyMT](https://en.wikipedia.org/wiki/Mersenne_Twister). I don't think that this would be a problem for our use case.

rafaelab:
Hi!
I think it is a good idea to display when an error occurs. Maybe we could also have a output.enableSeed(function), to be called if the user wishes to store this information as well.

lukasmerten:
I agree with Rafael: An opt-in for the txt-files would be nice.

PR #194, opened 2018-06-19
Store random seed, Closes: #190

TobiasWinchen:
These changes handle storing the random seed with aim to ease reproducibility of rare errors
* Seed is stored in hdf5 files per default
* Seed is stored in TextOutput if enabled
* Seed is printed in error stream on runtime_error in some interactions


TobiasWinchen:
If there are no further comments, I will merge this pull-request tomorrow.

PR #202, opened 2018-09-21
Add a modified, smoothed and solenoidal JF12 Galactic magnetic field

TimoSchorlepp:
Add a modified JF12 Galactic Magnetic field model, which is detailed in Kleimann, Schorlepp, Merten and Becker Tjus, "Solenoidal improvements for the JF12 Galactic magnetic field model", [arXiv:1809.07528](https://arxiv.org/abs/1809.07528v1), as a new MagneticField called "JF12Solenoidal". This model includes both a smooth redistribution of magnetic flux of the spiral disk field (to avoid cut off fieldlines and thus violations of magnetic flux conservation), and a smooth X field with parabolic fieldlines near the z=0 plane. The current field implementation was used for the numerical tests in the article to investigate whether the smoothed field leads to an improved performance of the DiffusionSDE module.

### DIFF
--- a/include/crpropa/magneticField/JF12FieldSolenoidal.h
+++ b/include/crpropa/magneticField/JF12FieldSolenoidal.h
@@ -1,7 +1,7 @@
 #ifndef CRPROPA_JF12FIELDSOLENOIDAL_H
 #define CRPROPA_JF12FIELDSOLENOIDAL_H
 
-#include "crpropa/magneticField/MagneticField.h"
+#include "crpropa/magneticField/JF12Field.h"
 #include "crpropa/Grid.h"
 #include <crpropa/Units.h>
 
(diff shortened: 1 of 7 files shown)


lukasmerten:
Hey all, 
please have a look at this implementation for smoother version of the JF12 field. We don't yet want to replace the old implementation but e.g. for the DiffusionSDE module this new version works better.

It can also be used as a starting point for other simpler magentic fields, as the user can turn on and off different parts of the magnetic field (x-field, disc-field, etc.)

As Timo already said: Details can be found in the arXiv paper which is currently under review.

MRCWirtz:
Hi Lukas & Timo,
great work! Can you provide some information about the differences in the deflections for UHECRs? Does it make any difference above energies of EeV (I guess not)? I would still go with a new field called "JF12Solenoidal" as proposed by Timo and not replace the old "JF12" field (this way other groups can still crosscheck their implementations with ours).
As far as I can judge, the code looks fine. However, I find a bit unphysical that the field lines close at 20 kpc suddenly (cf Fig 3, right). I would rather expect a model where the field lines close smoothly to infinity. Maybe one could think about to enable an option to not include this artifacts at 20kpc.

lukasmerten:
Hi Markus, 
we have not specifically checked for the differences in the deflection of UHECRs but is definitely something to do in the future. 
The reason for closing the field lines at r=20kpc was to have a solenoidal field also in a finite simulation volume. However, an option to change to a different boundary behavior sound reasonable. Nevertheless,  I would like to keep the code as it is until the paper is accepted to have a clear reference available.

If there are no further comments I will merge this branch tomorrow.
Best, Lukas

include/crpropa/magneticField/JF12FieldSolenoidal.h
```diff
@@ -0,0 +1,188 @@
...
+	bool isUsingTurbulent();
+	bool isUsingDisk();
+	bool isUsingToroidalHalo();
+	bool isUsingX();
+	
+	double getDelta() const;
+	double getZs() const;
+	
+	// scaling function for disk and toroidal halo field; same as in initial JF12
+	double logisticFunction(const double x, const double x0, const double w) const;
```
TobiasWinchen: const double without & makes no sense

avvliet:
Hi Lukas and Timo,

Thank you for the nice work, and apologies for my late reply. It looks good, but before merging I still have one small comment:
- I would suggest to be a bit clearer with the naming. Names of classes/methods/variables should be rather longer and clearer than shorter and unclear what they stand for. For example, from the names of 'setUseX', 'setDelta', 'setZs', 'zS', 'r1', 'r2', 'r1s', 'r2s', etc., it is not immediately clear what they stand for.

include/crpropa/magneticField/JF12FieldSolenoidal.h
```diff
@@ -0,0 +1,188 @@
...
+	double bDiskTurb5;   // field strength at r<5kpc
+	double zDiskTurb;	 // Gaussian scale height of disk
+	// halo
+	double bHaloTurb; // halo field strength
+	double rHaloTurb; // exponential scale length
+	double zHaloTurb; // Gaussian scale height
+
+public:
+
+	JF12FieldSolenoidal(double delta=(3*kpc), double zs=(0.5*kpc));
```
TobiasWinchen: remove () and insert spaces between operators

TobiasWinchen:
Thank you very much for the improvements. Inheritance is not the best solution, but IMO OK here as the new field is a JF12 field with some modification.

However, I just noticed some additional things. Sorry for not mentioning them earlier:
*  the naming of `double PhiIntegralH` does not follow our convention, this should be `getHPhiIntegral`, judging from the comment.
* The comments should be a little more extensive, e.g. what is Phi in PhiIntegralH, which aprameter ranges are allowed, etc.
* Comments should follow doxygen standard.
* There is still code duplication in the constructor. This should be resolved by calling the parents constructor from the child and set only the additional variables. 
* Why do you need an additional array bDiskCyclicClosure? Would it not be more easier to modify the array bDisk of the base class. 
* Regarding Arjens comment, good naming is in particular important for the interface. Internally this can always change without breaking user code. But in the interface, the users will have a field instance in the python shell and will then wonder what is `p' for a method. So I would suggest:

`setDelta -> setTransitionWidth`
`setZs -> setScaleHeightOfXField`
`q -> getTransitionPolynomial`
`p -> getTransitionPolynomialDerivative`

and add verbose comments, e.g.:
```
/// get transition polynomial p_delta(r) for the spiral field at distance r
double getTransitionPolynomial(const double &r) const;
```

TimoSchorlepp:
@TobiasWinchen I renamed a few methods and variables now and added comments in the cpp file which should make things a bit clearer (however, some technical details are hard to explain without a sketch, like the indexing of the phi0Arms array). I also tried to add Doxygen comments for the methods which are not self explanatory; I hope I did it correctly since I've never used Doxygen before.

TobiasWinchen:
You can check for yourself by  using `make doc` in the build directory,  if doxygen is installed of course. Then a folder html is generated that contains the webpage. Open th file `index.html' in a browser and check how the documentation looks. On first sight the documentation looks OK. I will check the rest and merge this week. Thank you for making the changes and contributing!

TobiasWinchen:
Thanks - I just merged (and fixed some more whitespace)

MRCWirtz:
> @TobiasWinchen I renamed a few methods and variables now and added comments in the cpp file which should make things a bit clearer (however, some technical details are hard to explain without a sketch, like the indexing of the phi0Arms array). I also tried to add Doxygen comments for the methods which are not self explanatory; I hope I did it correctly since I've never used Doxygen before.

Hi Timo, I don't quite understand why you renamed function names like "setUseRegular", "setUseStriated" and "setUseTurbulent", which prevents backwards-compatibility. These function calls occur probably in hundreds of scripts... For me this would only make sense either when we release a new version of CRPropa or when there is a major issue with the naming until now, which does not seem to be the case.

TimoSchorlepp:
Ah, so the idea was to choose a uniform naming convention for the 'setUse...'-commands, since I had to name them 'setUseXField' etc. for the individual regular field components instead of 'setUseX'. You are right, of course, it would have been better not to change those, sorry!

TobiasWinchen:
The consistency of the names is improved. We are working towards a release of CRProap 3.2.0. This is also a minor change in the API that can be fixed using a quite simple find&replace via sed if it is really hundreds of files. So I would be in favor of keeping it.

MRCWirtz:
Okay, but anyways we could keep the old function names additionally (with user warning) until the release of CRPropa 3.2. At least users are prepared to change their code then

MRCWirtz:
For people that are interesting in the differences that arise for propagation of UHECRs, here 10 EV rigidity, I am attaching the angular separation after backtracking through the JF12 field and the new JF12Solenoidal field for the respective pixel (averaged over 10 cosmic rays per pixel). I am surprised about the impact of correcting for divergence free fields, which is of the order of 30 degrees close to the disc.

[angular_diff_backtrack.pdf](https://github.com/CRPropa/CRPropa3/files/2495730/angular_diff_backtrack.pdf)

PR #210, opened 2018-11-06
Massdistribution

JulienDoerner:
Add a new structure for massdistribution. This includes a new virtual base class for target-densities and some individual models for the Milky-Way. 
The density base class contains flags for the target-type (HI, HII, H2), which is included and get functions for the type-density, the total density and the nucleon density. 
Now implemented models for the Milky Way are: 
* Ferrière: contains HI, HII, H2; in two regions (R < 3kpc: arXiv:astro-ph/0702532 ; R > 
3kpc: ApJ 497:759)
* Cordes:  contains HII (Nature 354, pages 121–124)
* Nakanishi: contains HI (arXiv:astro-ph/0304338) and H2 (arXiv:astro-ph/0610769), implemented fit in (arXiv:1607.07886 Appendix C)

Additional there are general structures for a constant density (each type seperately adjustable) and two superposition types for densities:
* CustomDensity for a superposition of one HI, one HII and one H2 type of different models 
* DensityList for the superpositon of more than one density model


include/crpropa/massdistribution/ConstantDensity.h
```diff
@@ -0,0 +1,65 @@
...
+	double getHIDensity(const Vector3d &position) const;
+	double getHIIDensity(const Vector3d &position) const;
+	double getH2Density(const Vector3d &position) const;
+	double getNucleonDensity(const Vector3d &position) const;
+	
+	bool getisforHI();
+	bool getisforHII();
+	bool getisforH2();
+	
+	void setHI(bool activate, double densitynumber);
```
adundovi: I must say that I don't understand why we need to activate a specific density and do all the connected steps and checks with it if a not-used density can be just 0 and be always activated? Or am I missing something?

src/massdistribution/Cordes.cpp
```diff
@@ -0,0 +1,50 @@
...
+	
+	double n=0;
+	
+	double z=position.z;
+	double R = sqrt(position.x*position.x+position.y*position.y);	//radius in galactic disk
+	
+	n += 0.025*exp(-fabs(z)/(1*kpc))*exp(-pow(R/(20*kpc),2));	//galactocentric component
+	n += 0.2*exp(-fabs(z)/(0.15*kpc))*exp(-pow((R-4*kpc)/(2*kpc),2));	//anular component
+
+	return n/ccm;
```
adundovi: I find this as undesirable behaviour. If we use internally SI units everywhere, why this explicit conversion of units? Can it be avoided?

lukasmerten:
Furthermore, the densityTest should be updated to exclude the test of the Pohl distribution for now.

include/crpropa/massdistribution/Density.h
```diff
@@ -0,0 +1,47 @@
...
+namespace crpropa {
+/**
+ @class Density
+ @brief Abstract base class for targetdensity
+ */
+class Density: public Referenced {
+ public:
+  virtual ~Density() {
+  }
+  virtual double getDensity(const Vector3d &position) const { 	// sum of all densities
```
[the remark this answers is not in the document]
JulienDoerner: This could cause an error, because not all models have each hydrogyn component (e.g. Cordes). 
If we define the pure virtual functions as "=0", it is necessary to define in each model a "return 0" for not used components. I think it is easier to define the standard virtual funktion with "return 0" in the base class and only overwrite the functions, which are implemented in the models.

include/crpropa/massdistribution/ConstantDensity.h
JulienDoerner: If you call the getDensity or the getNucleonDensity function, only active parts are summed up. 
If you call the get\<Type>Density function, the (constant)Densitynumber for this type is returned. 
So in the case that you want to ignore the ionised componend for the interaction you can deactivate this component. But if other modules (which we don't have jet) needs the charged part of the massdistribution, it is possible to call the getHIIDensity. So you don't need to have two densities for this case.

include/crpropa/massdistribution/Density.h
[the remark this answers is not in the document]
adundovi: OK. That is a valid reason. However, it still looks a bit complicated solution, but since I don't have any better suggestion for now, this can go. When we will have the wider picture and more practical experience we can re-implement it.

src/massDistribution/Ferriere.cpp
```diff
@@ -0,0 +1,330 @@
...
+
+double Ferriere::getHIDensity(const Vector3d &position) const {
+	
+	double n = 0;
+	double R = sqrt(position.x*position.x+position.y*position.y);	
+	bool inner = R<3*kpc;
+	
+	if(inner == true) 
+	{
+		double nCMZ = 0;	//density center in m^-3
```
TobiasWinchen: avoid declaring this as empty first, but later directly write

double nCMZ = 8.8 ....

this makes the code more readable

src/massDistribution/Ferriere.cpp
```diff
@@ -0,0 +1,330 @@
...
+		n+=getHIIDensity(position);
+	}
+	if(isforH2){
+		n+=getH2Density(position);
+	}
+	
+	//check if any density is activ and give warning if not
+	bool anyDensityActive = isforHI||isforHII||isforH2;
+
+	if(anyDensityActive == false){
```
TobiasWinchen: avoid declaring variable if not needed outside, i.e. write directly if (isforH1 | isforHII | ... }.

However, is the warning here necessary? This introduces a check (and potential output) on every get which can be a a lot I guess.

test/testDensity.cpp
```diff
@@ -0,0 +1,288 @@
...
+	
+	EXPECT_NEAR(n.getHIIDensity(p), 184500.,1);	// output in m^-3 ; uncertainty of 1e-6 cm^-3 
+	EXPECT_NEAR(n.getDensity(p), 184500.,1);	
+	EXPECT_NEAR(n.getNucleonDensity(p), 184500,1);	// only HII component -> no differenz between density and nucleon density
+	p.z=30*pc;			// invariant density for +/- z
+	EXPECT_NEAR(n.getDensity(p),184500,1); 
+	
+}
+
+TEST(testNakanishi, SimpleTest) {
```
TobiasWinchen: SimpleTest is ot a very descriptive name - how about: checkValueAtCertainPonints.

TobiasWinchen:
@lukasmerten - yes, having some examples for the code would be really helpful to understand the design choices. Currently I don't see why the design with several boolean flags for all individual components is necessary and cannot be handled e.g. in the way I sketched above that would be IMO simpler and more flexible.

JulienDoerner:
I created a pull request with the examples for the density in the notebook repository.

PR #211, opened 2018-11-09
Added PhotonOutput1D option to regular output 

avvliet:
Added PhotonOutput1D option to regular output and adjusted DINT and Eleca propagation to accept this output (the order of output quantities changed). Also created a test for this output and adjusted the warning in 'PhotonOutput1D.cpp'.

### DIFF
--- a/src/module/PhotonOutput1D.cpp
+++ b/src/module/PhotonOutput1D.cpp
@@ -18,16 +18,16 @@ using namespace std;
 namespace crpropa {
 
 PhotonOutput1D::PhotonOutput1D() : out(&std::cout) {
-	KISS_LOG_WARNING << "PhotonOutput1d is deprecated and will be removed in the future. Replace with TextOutput or HDF5Output with features ObserverNucleusVeto + ObserverDetectAll";
+	KISS_LOG_WARNING << "PhotonOutput1D is deprecated and will be removed in the future. Replace with TextOutput or HDF5Output with features ObserverNucleusVeto + PhotonOutput1D";
 }
 
 PhotonOutput1D::PhotonOutput1D(std::ostream &out) : out(&out) {
-	KISS_LOG_WARNING << "PhotonOutput1d is deprecated and will be removed in the future. Replace with TextOutput or HDF5Output with features ObserverNucleusVeto + ObserverDetectAll";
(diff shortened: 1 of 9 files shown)


src/module/PhotonOutput1D.cpp
```diff
@@ -18,16 +18,16 @@ using namespace std;
 namespace crpropa {
 
 PhotonOutput1D::PhotonOutput1D() : out(&std::cout) {
-	KISS_LOG_WARNING << "PhotonOutput1d is deprecated and will be removed in the future. Replace with TextOutput or HDF5Output with features ObserverNucleusVeto + ObserverDetectAll";
+	KISS_LOG_WARNING << "PhotonOutput1D is deprecated and will be removed in the future. Replace with TextOutput or HDF5Output with features ObserverNucleusVeto + PhotonOutput1D";
```
TobiasWinchen: Wouldn't you still need the ObserverDetectAll?

avvliet: No, something like this would work now:

```
# observer
obs1 = Observer() # proton output
obs1.add( ObserverPoint() )
obs1.add( ObserverPhotonVeto() ) # we don't want photons here
obs1.onDetection( TextOutput('proton_output.txt', Output.Event1D) )
m.add(obs1)
obs2 = Observer() # photon output
obs2.add( ObserverPoint() )
obs2.add( ObserverNucleusVeto() ) # we don't want hadrons here
obs2.onDetection( TextOutput('photon_output.txt', Output.PhotonOutput1D) )
m.add(obs2)
```

avvliet:
> Would it be a solution to use Event1D and enable the created columns in the Examples?

Yes, that would be possible as well I guess. I just thought it would be easier for the user to have a specific photon output that contains the same things as the original PhotonOutput1D and that can be loaded right away with DINT and Eleca. Because those two only work if you have exactly the correct output specified.

src/module/PhotonOutput1D.cpp
avvliet: Yes, you're right. Using ObserverDetectAll would probably be clearer.

avvliet:
I removed the PhotonOutput1D option again from the regular output, but the changes to PhotonPropagation.cpp are still necessary to handle the new output. Once this is merged I will update the photon example to use this output.

PR #215, opened 2018-12-07
Vector spacing for Grid

JulienDoerner:
I changed the spacing of a grid from one double to a Vector3d. Now are grids with different spacing along the axis suported.


adundovi:
Could you please give us a bit of background for this change? Do you plane to introduce some structurally anisotropic magnetic field models which would require such a feature? If the usage of it would be too specific, maybe just forking CRPropa for this investigation would be sufficient?

JulienDoerner:
This feature is used for density grids of the Milky Way. In many cases the resolution of the grid in the plane can be lower than the resolution on the z-axis.

PR #218, opened 2019-01-01
custom photon fields

Froehliche-Kernschmelze:
Dear all,

this pull request aims to introduce the following changes to CRPropa:

1) fully customizable photon fields:
    a) in addition to previous photon fields (CMB, IRBs) there are 8 open slots PF1...PF8 for custom fields
    b) each photon field satisfies a certain format encoding energy- and redshift dependence
    c) the tabulated files needed by the interaction modules can be generated by a central external file
    d) the field slots PF1...PF8 may be modified at will after CRPropa has been installed
    e) the photon field file itself and the tabulated files are stored in the ~/share/crpropa directory
    f) field photons can now be sampled by a class provided in PhotonBackground.cpp
    g) SOPHIA had to undergo significant changes in this process

2) space- and time dependence
   a) interaction modules support space- and time dependence of photon fields
   b) for this the ScalarGrid4d type has been added which generalizes ScalarGrid to 4 dimensions
   c) interaction modules take a ScalarGrid4d encoding space- and time dependence
   d) ScalarGrid4d comes with its most-needed tools e.g. interpolation, loadGridFromTxt, etc.

3) discretized version of SOPHIA
   a) a SOPHIA event generator based on histogram data may be used in the PhotoPionProduction
   b) this may be controlled via a new flag in PhotoPionProduction: useTabulatedData = true/false
   c) see issue #213 for details of this implementation
   d) NOTE: this tightly relates to the modified version of SOPHIA that came with 1g)

Looking forward to receiving feedback from you!
All the best,
Mario

Issue #213, opened 2018-11-19
discretize SOPHIA


Froehliche-Kernschmelze:
currently discretizing the SOPHIA code, aiming to fix the saturation of the PhotoPionProduction beyond 8 cores.

adundovi:
I once had started to do something similar, but never finished. Maybe the following will help.

This is ```Makefile``` that generates a python interface for SOPHIA:
```makefile
F2PY=f2py

all: sophia.so

sophia.so: sophia.pyf sophia_interface.f
        $(F2PY) -c sophia_interface.f sophia.pyf

clean:
        rm sophia.so
```
And this is my naive approach to generate tables of events
```python
from sophia import *
import numpy as np
import sys

import itertools as it
from multiprocessing import Pool

# parallelize the work
max_cores = 4

# input
nature = [0,1] # proton, neutron
Ein_div = 10000
Ein = {'CMB': 3.72*np.logspace(9, 13, num=Ein_div),
       'IRB': 5.83*np.logspace(6, 12, num=Ein_div) }  # Gev

redshift = np.array([0, 0.01, 0.03, 0.05, 0.1, 0.2, 0.3, 0.4,
                     0.5, 0.6, 0.8, 1.0, 1.25, 1.5, 2.0, 2.5,
                     3.0, 4.0, 5.0, 6.0]) # usual redshifts
backgrounds = { 'CMB': 1,'IRB': 2 }
maxRedshift = 100 # hardcoded in CRPropa's source

def sevent( n, E, z, bg ):
    """Wrapper for ``sophiaevent``.
       Parameters
       ----------
       nature : input int
       ein : input float
       outpart : in/output rank-2 array('d') with bounds (2000,5)
       outparttype : in/output rank-1 array('i') with bounds (2000)
       nboutpart : in/output rank-0 array(int,'i')
       z_particle : input float
       bgflag : input int
       zmax_irb : input float
       en_data : input rank-1 array('d') with bounds (idatamax)
       flux_data : input rank-1 array('d') with bounds (idatamax)
    """

    # output variables
    particleList = np.zeros( 2000, dtype=np.int32 )
    momentaList = np.zeros( (2000,5), dtype=np.float_, order='F' )
    nParticles = np.array(0) # number of particles, scalar field

    # dummy variables
    dummy1 = np.int_()
    dummy2 = np.zeros(10, dtype=np.float_)
    dummy3 = np.zeros(10, dtype=np.float_)

[...]

PR #220, opened 2019-01-29
custom photon fields / SOPHIA update

Froehliche-Kernschmelze:
Dear all,

with this pull request I want to provide the first of three parts that I wish to contribute to CRPropa. This approach follows the comments made in #218 

**Content:**

This extension is meant to introduce variable photon fields to CRPropa. Previously, there has only been the CMB and several IRB models. This photon field extension basically follows three concepts:

The photon field may

1) have any shape, i.e. any dependence of photon energy and field density
2) have any spacial and temporal behaviour (which is part of a future pull request)
3) be modified even after CRPropa has been installed

**Changelog (most severe changes):**

~/src/PhotonBackground.cpp
~/include/PhotonBackground.h

There are now eight new slots PF1..PF8 that may be included in the simulation. A class has been added to draw a random photon from the photon field that takes the primary, the nucleon/gamma crossection and the photon field's shape into account. This method and all those that it depends on have been inspired by SOPHIA.

Photon fields are now contained in photon field files in ~/share/crpropa/Scaling/ . For detailed information on the format and use of these fields, there is a [Wiki page](https://github.com/Froehliche-Kernschmelze/CRPropa3/wiki/How-to-create-and-use-custom-photon-fields) in my forked repository.

~/libs/sophia/sophia_interface.f

The main point of these changes to SOPHIA is to run the code on other photon fields than the CMB and IRB Kneiske (as only these two are implemented so far in SOPHIA). SOPHIA samples photons from these fields with its own photon sampling routine which in this version has been removed and replaced by the sampling method mentioned in the section above. For this, the input and output parameters of SOPHIA changed from
```
sophiaevent(nature, Ein, OutPart, OutPartType, NbOutPart, z_particle, bgFlag, Zmax_IRB, idatamax, en_data, flux_data)
```
to
```
sophiaevent(nature, Ein, OutPart, OutPartType, NbOut, eps).
```
Furthermore, SOPHIA has recieved:
- a removal of all parts of the program that are never used under any circumstances
- a removal of all GOTO pointers which have been replaced by safer control structures
- a unified indentation to ehance the readability of the code
- a re-naming of several variables and methods from cryptic abbreviations to full identifiers
- docstrings in most functions that as far as tracable describe their functions

~/src/module/PhotonPionProduction.cpp
~/include/module/PhotonPionProduction.h

The flag "haveRedshiftDependence" and all its associated functions have been removed as now all photon fields are represented on grids that are redshift-dependent. The signature of SOPHIA has been updated. The treatment of secondary particles being produced in SOPHIA has been rewritten in a more intuitively readable way. Also, this particle treatment did not allow for more than one hadron to be added among all secondaries produced by SOPHIA. This is valid for the CMB and IRB models as not eanough energy in the center of mass system would be present to produce nucleon/anti-nucleon pairs. However, with more energetic photon fields, these particles may occur. I would like to stress that no kind of anti-nucleon can be produced in the current implementation, even if it were produced.

**Tests**

As photon sampling is removed from SOPHIA and now provided within CRPropa, SOPHIA's sampling routines and its CRPropa counterparts are expected to produce similar results. Therefore, the following tests have been run according to their hierarchy in the procedure of sampling photons:

Hierarchy of sampling functions (highest to lowest):
```
sample_eps -> prob_eps -> get_photonDensity, gaussInt -> functs -> crossection
```
Reference simulation setup:
```
SimplePropagation(10 kpc, 100 kpc)
PhotoPionProduction(field=CMB, neutrinos=True)
Source @ (100 Mpc, 0, 0)										
Injected: 100.000 monoenergetic protons with 1e12 GeV
```
1) crossection is the deepest layer to be tested. Note that the SOPHIA function that returns the nucleon-gamma xsection does not depend on the nucleon's energy itself but rather on the fact whether or not it deals with a proton or a neutron. 
The reference simulation called the crossection function roughly 500.000 times. The sampled photon's energies have been injected into the newly written C++ version of crossection producing another set of ca 500.000 crossection values. These two samples have been inserted into the scipy.stats.ks_2samp Kolmogorov-Smirnov test. A result of D=0.0 and p=1.0 corresponds to total identity in this implementation of the Kolmogorov-Smirnov test. The test achieved a result of D=0.0011, p=0.9183. Note that the functions Ef, Pl and breitwigner are only used by crossection.

2) In an analogous way, the function gaussInt has been tested, which integrates the function functs over all center-of-mass energies. the function functs is only called by gaussint, whereas functs is the only function calling crossection. The KS test yielded a result of D=9.61e-5, p=0.9999999999 between the SOPHIA output and the C++ version.

3) prob_eps calculates the non-normalized probability to encounter a photon of the given photon field. Normalization is achieved by dividing the integral of prob_eps over its entire photon energy range. prob_eps is the only function calling gaussInt and get_photonDensity. For better comparison, get_photonDensity has been replaced by an analytical expression returning the photon density of the CMB, i.e. a blackbody with a temprature of 2.73 K. The results of SOPHIA vs C++ yield D=9.61e-5, p=0.9999999999. Results on grid data (and runtimes) are discussed in a later section.

4) Finally, sample_eps calls prob_eps to sample a photon. This happens in two steps: a) calculate the normalization and the photon energy with which one most likely gets an interaction. b) monte-carlo sample a photon from the field, use norm factor and max probability for this. For comparison, the SOPHIA output has been tested vs the C++ analytical expression for the photon density and vs the grid readout. The KS test yields:
```
SOPHIA vs C++ (analytic):     D=0.0224, p=0.0114
SOPHIA vs C++ (grid):         D=0.0225, p=0.0110
C++ (analytic) vs C++ (grid): D=0.0001, p=0.9999
```
Insights: the grid version and the analytic version are able to produce nearly identical outputs, however less similarity is achieved compared to SOPHIA. In return, SOPHIA operates a dedicated sampling routine that only works for the CMB this making several assumptions about the shape of the field in beforehand. The C++ version do not require any information that is not encoded in the photon field file. One might be interested in the fact that SOPHIA uses its own random number generator whereas the C++ routines have been run with std::srand().

5) Totally finally, the sampling method is implemented into CRPropa and tested on observables. For this, the neutrino output (energies) of the reference simulation has been chosen. The KS test result of a neutrino the neutrino outputs acquired by SOPHIA vs C++ sampling have been tested in three ways fore comparison: a) using no grid i.e. the analytical density readout b) density readout of a CMB grid with 101 points of resolution c) density readout of a CMB grid with 1001 points of resolution
```
SOPHIA vs C++ (analytic):  1:57 | 2:20 @ D=0.0035, p=0.3514
SOPHIA vs C++ (Grid_101):  1:57 | 2:20 @ D=0.0037, p=0.2716
SOPHIA vs C++ (Grid_1001): 1:57 | 5:58 @ D=0.0035, p=0.3514
```
Insights: The C++ small grid / analytic versions perform roughly 20% slower than SOPHIA. A larger grid may reach the analytic version at the expense of runtime. All C++ versions perform similarly but do not reach identity with the SOPHIA version. The two points coming to my mind are:
a) the C++ version of sampling has been done on CRPropa's random number generator, SOPHIA uses its own one
b) SOPHIA operates a dedicated sampling method that only works on the CMB (yet another one with IRB_Kneiske) making assumptions about the CMB shape in beforehand. The C++ version does not rely on such information.

Important notes: 
1) all the tests have been run on the CMB as this allows for tests against analytical expressions of the photon field and SOPHIA. The other only photon field implemented in SOPHIA is IRB_Kneiske. First tests did not yield a correlation between SOPHIA and C++. Before going deeper here, I would like to invoke your feedback on the current implementation.
2) The code does not tell CRPropa to build the photon field files stored in ~/share/crpropa/Scaling upon installation. Also, the tabulated files of the PhotoPionProduction have changed. Do you have any advice in how to include this?
3) The photon field URB_Protheroe96 could not be converted into a photon field file as there is no sufficient photon field data in the CRPropa-data repository for this. Hence, this field is currently unavailable.

Thanks in advance and best wishes,
Mario

### DIFF
--- a/include/crpropa/PhotonBackground.h
+++ b/include/crpropa/PhotonBackground.h
@@ -1,7 +1,8 @@
 #ifndef CRPROPA_PHOTONBACKGROUND_H
 #define CRPROPA_PHOTONBACKGROUND_H
 
-#include <string>
+#include <crpropa/Vector3.h>
+#include <crpropa/Grid.h>
 
 namespace crpropa {
 
@@ -11,6 +12,7 @@ namespace crpropa {
  */
 // Photon fields
 // The default IRB model is that of Kneiske et al. 2004
+// The slots PF1 to PF8 may be used for custom photon fields
 enum PhotonField {
 	CMB,
 	IRB,  // same as IRB_Kneiske04
@@ -22,7 +24,9 @@ enum PhotonField {
 	IRB_Gilmore12,
 	IRB_Stecker16_upper,
 	IRB_Stecker16_lower,
-	URB_Protheroe96
+	URB_Protheroe96,
+	PF1, PF2, PF3, PF4,  // customizable
+	PF5, PF6, PF7, PF8,  // field slots
 };
 
 // Returns overall comoving scaling factor
@@ -30,6 +34,56 @@ double photonFieldScaling(PhotonField photonField, double z);
 
 // Returns a string representation of the field
 std::string photonFieldName(PhotonField photonField);
+
+/** 
+ @class CustomPhotonField
+ @brief Handler class for photon fields. Provides the sampleEps method.
+
+ sampleEps draws a photon from a given photon background. This method 
+ and all methods it depends on have been inspired by the SOPHIA code.
+ */
+class CustomPhotonField {
+public:
+	/** Constructor for photon field data
+	 @param fieldPath  path/to/photonField.txt
+	 */
+	explicit CustomPhotonField(std::string fieldPath);
+
+	/* Empty constructor to ease initialization in some modules
+	 */
+	CustomPhotonField();
+
+	/** Draws a photon from the photon background
+	 @param onProton  true=proton, false=neutron
+	 @param Ein       energy of primary
+	 @param zIn       redshift of primary
+	 */
+	double sampleEps(bool onProton, double Ein, double zIn) const;
+
+	/** Returns the photon field density in 1/(Jm³).
+		Multiply by h*nu for physical photon density.
+	 @param eps   photon energy in eV
+	 @param zIn   redshift
+	*/
+	double getPhotonDensity(double eps, double z) const;
+
+	/** Returns the crossection of p-gamma interaction
+	 @param eps       photon energy
+	 @param onProton  true=proton, false=neutron
+	*/
+	double SOPHIA_crossection(double eps, bool onProton) const;
+
+protected:
+	std::vector<double> photonEnergy;
+	std::vector<double> photonRedshift;
+	std::vector<double> photonDensity;
+	void init(std::string fieldPath);
+	double SOPHIA_probEps(double eps, bool onProton, double Ein, double zIn) const;
+	double SOPHIA_pl(double x, double xth, double xmax, double alpha) const;
+	double SOPHIA_ef(double x, double th, double w) const;
+	double SOPHIA_breitwigner(double sigma_0, double Gamma, double DMM, double epsPrime, bool onProton) const;
+	double SOPHIA_functs(double s, bool onProton) const;
+};
 
 /** @}*/
 } // namespace crpropa
--- a/src/PhotonBackground.cpp
+++ b/src/PhotonBackground.cpp
@@ -1,11 +1,16 @@
 #include "crpropa/PhotonBackground.h"
 #include "crpropa/Common.h"
+#include "crpropa/Random.h"
+#include "crpropa/Units.h"
 
 #include <vector>
 #include <fstream>
 #include <stdexcept>
 #include <limits>
 #include <cmath>
+#include <sstream>
+#include <iostream>
+#include <string>
 
 namespace crpropa {
 
@@ -64,8 +69,10 @@ static PhotonFieldScaling scalingStecker16_lower("IRB_Stecker16_lower");
 
 double photonFieldScaling(PhotonField photonField, double z) {
 	switch (photonField) {
-	case CMB:
-		return 1;  // constant comoving photon number density
+	case CMB: // constant comoving photon number density
+	case PF1: case PF2: case PF3: case PF4:
+	case PF5: case PF6: case PF7: case PF8:
+		return 1;
 	case IRB:
 	case IRB_Kneiske04:
 		return scalingKneiske04.scalingFactor(z);
@@ -84,12 +91,9 @@ double photonFieldScaling(PhotonField photonField, double z) {
 	case IRB_Stecker16_lower:
 		return scalingStecker16_lower.scalingFactor(z);
 	case URB_Protheroe96:
-		if (z < 0.8)
-			return 1;
-		if (z < 6)
-			return pow((1 + 0.8) / (1 + z), 4);
-		else
-			return 0;
+		if (z < 0.8) { return 1; }
+		if (z < 6) { return pow((1 + 0.8) / (1 + z), 4); }
+		else { return 0; }
 	default:
 		throw std::runtime_error("PhotonField: unknown photon background");
 	}
@@ -97,30 +101,295 @@ double photonFieldScaling(PhotonField photonField, double z) {
 
 std::string photonFieldName(PhotonField photonField) {
 	switch (photonField) {
-	case CMB:
-		return "CMB";
-	case IRB:
-	case IRB_Kneiske04:
-		return "IRB_Kneiske04";
-	case IRB_Stecker05:
-		return "IRB_Stecker05";
-	case IRB_Franceschini08:
-		return "IRB_Franceschini08";
-	case IRB_Finke10:
-		return "IRB_Finke10";
-	case IRB_Dominguez11:
-		return "IRB_Dominguez11";
-	case IRB_Gilmore12:
-		return "IRB_Gilmore12";
-	case IRB_Stecker16_upper:
-		return "IRB_Stecker16_upper";
-	case IRB_Stecker16_lower:
-		return "IRB_Stecker16_lower";
-	case URB_Protheroe96:
-		return "URB_Protheroe96";
-	default:
-		throw std::runtime_error("PhotonField: unknown photon background");
+		case CMB: return "CMB";
+		case PF1: return "PF1";
+		case PF2: return "PF2";
+		case PF3: return "PF3";
+		case PF4: return "PF4";
+		case PF5: return "PF5";
+		case PF6: return "PF6";
+		case PF7: return "PF7";
+		case PF8: return "PF8";
+		case IRB:
+		case IRB_Kneiske04: return "IRB_Kneiske04";
+		case IRB_Stecker05: return "IRB_Stecker05";
+		case IRB_Franceschini08: return "IRB_Franceschini08";
+		case IRB_Finke10: return "IRB_Finke10";
+		case IRB_Dominguez11: return "IRB_Dominguez11";
+		case IRB_Gilmore12: return "IRB_Gilmore12";
+		case IRB_Stecker16_upper: return "IRB_Stecker16_upper";
+		case IRB_Stecker16_lower: return "IRB_Stecker16_lower";
+		case URB_Protheroe96: return "URB_Protheroe96";
+		default:
+			throw std::runtime_error("PhotonField: unknown photon background");
 	}
+}
+
+CustomPhotonField::CustomPhotonField(std::string fieldPath) {
+	init(fieldPath);
+}
+
+CustomPhotonField::CustomPhotonField() {
+	// empty constructor for initialization in some modules
+}
+
+void CustomPhotonField::init(std::string filename) {
+	std::vector< std::vector<double> > dndeps;
+	std::ifstream infile(filename.c_str());
+	if (!infile.good())
+		throw std::runtime_error("PhotoPionProduction @ CustomPhotonField::init : could not open file " + filename);
+	std::string line;
+	int i = 0;
+	while (std::getline(infile, line)) {
+		if (line.at(0) == '#')
+			continue;
+		std::istringstream ss(line);
+		std::vector<double> vec;
+		double n;
+		while (ss >> n)
+			vec.push_back(n);
+		if (i == 0) {
+			photonEnergy = vec;
+			i++;
+			continue;
+		}
+		if (i == 1) {
+			photonRedshift = vec;
+			i++;
+			continue;
+		}
+		dndeps.push_back(vec);
+	}
+	for (int i = 0; i < photonEnergy.size(); ++i) {
+		for (int j = 0; j < photonRedshift.size(); ++j) {
+			photonDensity.push_back(dndeps[i][j]);
+		}
+	}
+	infile.close();
+}
+
+double CustomPhotonField::sampleEps(bool onProton, double Ein, double zIn) const {
+/*
+	- input: particle type with energy Ein at redshift z
+	- output: photon energy [eV] of random photon of photon field
+	- samples distribution of n(epsilon)/epsilon^2
+*/
+	const double zMax = photonRedshift[photonRedshift.size() - 1];
+	if (zIn > zMax)
+		return 0.;
+
+	Ein /= GeV;  // SOPHIA standard unit
+	// calculate pMax and its norm factor via maximum such that peps <= 1
+	double cnorm = 0.;
+	double pMax = 0.;
+	for (int i = 0; i < photonEnergy.size(); ++i) {
+		double prob = SOPHIA_probEps(photonEnergy[i], onProton, Ein, zIn);
+		cnorm += prob;
+		if (prob > pMax)
+			pMax = prob;
+	}
+	pMax /= cnorm;
+
+	// sample eps between epsMin ... epsMax
+	const double epsMin = photonEnergy[0];
+	const double epsMax = photonEnergy[photonEnergy.size() - 1];
+	double eps;
+	double peps;
+	Random &random = Random::instance();
+	do {
+		eps = epsMin + random.rand() * (epsMax - epsMin);
+		peps = SOPHIA_probEps(eps, onProton, Ein, zIn) / cnorm;
+	} while (random.rand() * pMax > peps);
+	return eps;
+}
+
+double CustomPhotonField::SOPHIA_probEps(double eps, bool onProton, double Ein, double zIn) const {
+/*
+	- input: eps [eV]
+	- output: probability to encounter photon of energy eps
+	- called by: sampleEps, gaussInt
+*/
+	const double mass = onProton? 0.93827 : 0.93947;  // Gev/c^2
+	double gamma = Ein / mass;
+	double beta = std::sqrt(1. - 1. / gamma / gamma);
+	double photonDensity = getPhotonDensity(eps * eV, zIn) / 6.2415091e24;  // 1/(Jm³) -> 1/(eVcm³)
+	if (photonDensity == 0.) {
+		return 0.;
+	} else {
+		double sMin = 1.1646;  // head-on collision
+		double sMax = std::max(sMin, mass * mass + 2. * eps / 1.e9 * Ein * (1. + beta));
+		const double x[8] = {.0950125098, .2816035507, .4580167776, .6178762444,
+							 .7554044083, .8656312023, .9445750230, .9894009349};
+		const double w[8] = {.1894506104, .1826034150, .1691565193, .1495959888,
+							 .1246289712, .0951585116, .0622535239, .0271524594};
+		double xm = 0.5 * (sMax + sMin);
+		double xr = 0.5 * (sMax - sMin);
+		double ss = 0.;
+		for (int i = 0; i < 8; ++i) {
+			double dx = xr * x[i];
+			ss += w[i] * (SOPHIA_functs(xm + dx, onProton) + SOPHIA_functs(xm - dx, onProton));
+		}
+		double sIntegral = xr * ss;
+		return photonDensity / eps / eps * sIntegral / 8. / beta / Ein / Ein * 1.e24;
+	}
+}
+
+double CustomPhotonField::getPhotonDensity(double eps, double z) const {
+/*
+	- input: photon energy, redshift
+	- output: dndeps(e,z) [#/(eV cm^3)] from input file
+	- called by: sampleEps
+*/
+	return interpolate2d(eps / eV, z, photonEnergy, photonRedshift, photonDensity) * 6.2415091e24;  // 1/(eVcm³)->1/(Jm³)
+}
+
+double CustomPhotonField::SOPHIA_crossection(double x, bool onProton) const {
+/* 
+	- input: photon energy [eV], specifier: 0=neutron, 1=proton
+	- output: SOPHIA_crossection of nucleon-photon-interaction [area]
+	- called by: SOPHIA_functs
+*/  
+	const double mass = onProton? 0.93827 : 0.93947;  // Gev/c^2
+	double sth = 1.1646;
+	double s = mass * mass + 2. * mass * x;
+	if (s < sth)
+		return 0.;
+	// upper: proton resonance masses
+	// lower: neutron resonance masses
+	const double AMRES[18] = {1.231, 1.440, 1.515, 1.525, 1.675, 1.680, 1.690, 1.895, 1.950,
+							  1.231, 1.440, 1.515, 1.525, 1.675, 1.675, 1.690, 1.895, 1.950};
+	const double BGAMMA[18] = {5.6, 0.5, 4.6, 2.5, 1., 2.1, 2., 0.2, 1.,
+							   6.1, 0.3, 4.0, 2.5, 0., 0.2, 2., 0.2, 1.};
+	const double WIDTH[18] = {0.11, 0.35, 0.11, 0.1, 0.16, 0.125, 0.29, 0.35, 0.3,
+							  0.11, 0.35, 0.11, 0.1, 0.16, 0.150, 0.29, 0.35, 0.3};
+	const double RATIOJ[18] = {1., 0.5, 1., 0.5, 0.5, 1.5, 1., 1.5, 2.,
+							   1., 0.5, 1., 0.5, 0.5, 1.5, 1., 1.5, 2.};
+	const double AM2[2] = {0.882792, 0.880351};
+
+	int idx = onProton? 0 : 9;
+	double SIG0[9];
+	for (int i = 0; i < 9; ++i) {
+		SIG0[i] = 4.893089117 / AM2[int(onProton)] * RATIOJ[i + idx] * BGAMMA[i + idx];
+	}
+	double cross_dir = 0.;
+	double cross_res = 0.;
+	if (x <= 10.) {
+		cross_res = SOPHIA_breitwigner(SIG0[0], WIDTH[0 + idx], AMRES[0 + idx], x, onProton)
+				  * SOPHIA_ef(x, 0.152, 0.17);
+		for (int i = 1; i < 9; ++i) {
+			cross_res = SOPHIA_breitwigner(SIG0[i], WIDTH[i + idx], AMRES[i + idx], x, onProton)
+					   * SOPHIA_ef(x, 0.15, 0.38);
+		}
+		// direct channel
+		double cross_dir1 = 0.;
+		if ( (x > 0.1) && (x < 0.6) ) {
+			cross_dir1 = 92.7 * SOPHIA_pl(x, 0.152, 0.25, 2.)  // single pion production
+					   + 40.0 * std::exp(-(x - 0.29) * (x - 0.29) / 0.002)
+					   - 15.0 * std::exp(-(x - 0.37) * (x - 0.37) / 0.002);
+		} else {
+			cross_dir1 = 92.7 * SOPHIA_pl(x, 0.152, 0.25, 2.);  // single pion production
+		}
+		double cross_dir2 = 37.7 * SOPHIA_pl(x, 0.4, 0.6, 2.);  // double pion production
+		cross_dir = cross_dir1 + cross_dir2;
+	}
+	// fragmentation 2:
+	double cross_frag2;
+	if (onProton) {
+		cross_frag2 = 80.3 * SOPHIA_ef(x, 0.5, 0.1) * std::pow(s, -0.34);
+	} else {
+		cross_frag2 = 60.2 * SOPHIA_ef(x, 0.5, 0.1) * std::pow(s, -0.34);
+	}
+	// multipion production/fragmentation 1 cross section
+	double cs_multidiff = 0.;
+	if (x > 0.85) {
+		double ss1 = (x - 0.85) / 0.69;
+		double ss2;
+		if (onProton) {
+			ss2 = 29.3 * std::pow(s, -0.34) + 59.3 * std::pow(s, 0.095);
+		} else {
+			ss2 = 26.4 * std::pow(s, -0.34) + 59.3 * std::pow(s, 0.095);
+		}
+		cs_multidiff = (1. - std::exp(-ss1)) * ss2;
+		// diffractive scattering:
+		double cross_diffr = 0.11 * cs_multidiff;
+		// **************************************
+		ss1 = std::pow((x - 0.85), 0.75) / 0.64;
+		ss2 = 74.1 * std::pow(x, -0.44) + 62. * std::pow(s, 0.08);
+		double cs_tmp = 0.96 * (1. - std::exp(-ss1)) * ss2;
+		double cross_diffr1 = 0.14 * cs_tmp;
+		double cross_diffr2 = 0.013 * cs_tmp;
+		double cs_delta = cross_frag2
+						- (cross_diffr1+cross_diffr2-cross_diffr);
+		double cs_multi = 0.89 * cs_multidiff;
+		if (cs_delta < 0.) {
+			cross_frag2 = 0.;
+			cs_multi += cs_delta;
+		} else {
+			cross_frag2 = cs_delta;
+		}
+		cross_diffr = cross_diffr1 + cross_diffr2;
+		cs_multidiff = cs_multi + cross_diffr;
+	}
+	return (cross_res + cross_dir + cs_multidiff + cross_frag2) * 1.e-34;  // µbarn to m²
+}
+
+double CustomPhotonField::SOPHIA_pl(double x, double xth, double xmax, double alpha) const {
+/*
+	- input: photon energy [eV], threshold [eV], max [eV], unknown [no unit]
+	- output: unknown [no unit]
+	- called by: SOPHIA_crossection
+*/
+	if (xth > x)
+		return 0.;
+	double a = alpha * xmax / xth;
+	double prod1 = std::pow((x - xth) / (xmax - xth), (a - alpha));
+	double prod2 = std::pow(x / xmax, -a);
+	return prod1 * prod2;
+}
+
+double CustomPhotonField::SOPHIA_ef(double x, double th, double w) const {
+/*
+	- input: photon energy [eV], threshold [eV], unknown [eV]
+	- output: unknown [no unit]
+	- called by: SOPHIA_crossection
+*/
+	double wth = w + th;
+	if (x <= th) {
+		return 0.;
+	} else if ((x > th) && (x < wth)) {
+		return (x - th) / w;
+	} else if (x >= wth) {
+		return 1.;
+	} else {
+		throw std::runtime_error("error in function PhotonBackground::SOPHIA_ef");
+	}
+}
+
+double CustomPhotonField::SOPHIA_breitwigner(double sigma_0, double Gamma, double DMM,
+	double epsPrime, bool onProton) const {
+/*
+	- input: cross section [µbarn], width [GeV], mass [GeV/c^2]
+	- output: Breit-Wigner SOPHIA_crossection of a resonance widh width Gamma
+	- called by: SOPHIA_crossection
+*/
+	const double mass = onProton? 0.93827 : 0.93947;  // Gev/c^2
+	double s = mass * mass + 2. * mass * epsPrime;
+	double gam2s = Gamma * Gamma * s;
+	return sigma_0 * (s / epsPrime / epsPrime) * gam2s
+				   / ((s - DMM*DMM) * (s - DMM * DMM) + gam2s);
+}
+
+double CustomPhotonField::SOPHIA_functs(double s, bool onProton) const {
+/*
+	- input: s [GeV^2] 
+	- output: (s-p^2)*sigma_(nucleon/gamma) [GeV^2*area]
+	- called by: sampleEps
+*/  
+	const double mass = onProton? 0.93827 : 0.93947;  // Gev/c^2
+	double factor = s - mass * mass;
+	double epsPrime = factor / 2. / mass;
+	double sigma_pg = SOPHIA_crossection(epsPrime, onProton) / 1.e-34;  // m² to µbarn
+	return factor * sigma_pg;
 }
 
 } // namespace crpropa


src/module/PhotoPionProduction.cpp
```diff
@@ -2,11 +2,14 @@
 #include "crpropa/Units.h"
 #include "crpropa/ParticleID.h"
 #include "crpropa/Random.h"
+#include "crpropa/PhotonBackground.h"
 
 #include <kiss/convert.h>
-#include <kiss/logger.h>
 #include "sophia.h"
+#include <stdlib.h>
+#include <math.h>
```
TobiasWinchen: Avoid double include, we already have cmath !

include/crpropa/CRPropa.h
```diff
@@ -0,0 +1,71 @@
+#include "crpropa/Clock.h"
```
TobiasWinchen: Wht this second conveniance header? Use include/CRPropa.h !

include/crpropa/PhotonBackground.h
```diff
@@ -1,7 +1,16 @@
 #ifndef CRPROPA_PHOTONBACKGROUND_H
 #define CRPROPA_PHOTONBACKGROUND_H
 
+#include <crpropa/Vector3.h>
+#include <crpropa/Grid.h>
+
+#include <iostream>  // cout, endl
```
TobiasWinchen: If not used in the header, additional includes should go to the cpp file

src/PhotonBackground.cpp
```diff
@@ -63,64 +66,377 @@ static PhotonFieldScaling scalingStecker16_upper("IRB_Stecker16_upper");
...
+*/  
+    const double mass = onProton? 0.93827 : 0.93947;  // Gev/c^2
+    double factor = s - mass*mass;
+    double epsPrime = factor/2./mass;
+    double sigma_pg = crossection(epsPrime, onProton);
+    return factor*sigma_pg;
+}  // end functs
+
+
+double Photon_Field::gaussInt(double A, double B, bool onProton) const {
```
TobiasWinchen: The gauss integration should go to some general implmenetation and not be hidden inside the field

src/PhotonBackground.cpp
```diff
@@ -63,64 +66,377 @@ static PhotonFieldScaling scalingStecker16_upper("IRB_Stecker16_upper");
...
+*/
+    const double mass = onProton? 0.93827 : 0.93947;  // Gev/c^2
+    double s = mass*mass + 2.*mass*epsPrime;
+    double gam2s = Gamma*Gamma*s;
+    return sigma_0 * (s/epsPrime/epsPrime)*gam2s
+                   / ((s-DMM*DMM)*(s-DMM*DMM)+gam2s);
+}  // end breitwigner
+
+
+double Photon_Field::functs(double s, bool onProton) const {
```
TobiasWinchen: Please use unit conventions everywhere

src/PhotonBackground.cpp
```diff
@@ -63,64 +66,377 @@ static PhotonFieldScaling scalingStecker16_upper("IRB_Stecker16_upper");
...
+    }
+    // fragmentation 2:
+    double cross_frag2;
+    if (onProton) {
+        cross_frag2 = 80.3*Ef(x, 0.5, 0.1) * std::pow(s, -0.34);
+    } else {
+        cross_frag2 = 60.2*Ef(x, 0.5, 0.1) * std::pow(s, -0.34);
+    }
+    // multipion production/fragmentation 1 cross section
+    double cs_multidiff;
```
TobiasWinchen: set directly to 0 and delete else branch

src/PhotonBackground.cpp
```diff
@@ -63,64 +66,377 @@ static PhotonFieldScaling scalingStecker16_upper("IRB_Stecker16_upper");
...
 }
 
+
+// ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+// custom photon field methods related to SAMPLING
+// These methods are taken from SOPHIA.
+// ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+
+Photon_Field::Photon_Field(std::string fieldPath) {
```
TobiasWinchen: Read CRPropa style guide and adhere to it: CamelCase isntead of _

Froehliche-Kernschmelze:
Thanks for your feedback, I have applied all your comments so far and therefore kindly ask for the next step of review. A few notes:

* naming specifications: for clarity, exceptions exist in PhotonBackground.h/.cpp to emphasize methods directly taken from SOPHIA (this may be changed in a later step)
* unit conventions: for all public methods that have been added, especially in the photon field files, the CRPropa unit convention and usage has been adapted, e.g. getPhotonDensity(1*eV). This is not entirely consistent with SOPHIA methods as their unit conversions and equations cannot be completely resolved (However, KS tests show identical output of the original methods and their C++ versions). SOPHIA operates with GeV as its standard unit. Workarounds exist and in this draft of the code the user would not be mistaken with a second unit system.
* data and helper files will be added to their respective repository in another step. **Question:** How does one modify which data files are being written to the /share/ directory?
* A test unit for photon fields has been added and runs successfully on my system. Externally run tests (Travis) fail where the new photon field format is used as their data files are not available for now. This should change once the CRPropa-data repository has merged a pull request with these features.
* GitHub indicates conflicts with the master branch ever since the Boris Push has been merged.

Thanks again, looking forward to hearing from you soon,
Mario

TobiasWinchen:
> Thanks for your feedback, I have applied all your comments so far and therefore kindly ask for the next step of review.

Your pull request is still extremely huge - it seems that you are changing every file and every line of the code. The pull request should touch only files  that need to be changed for the SOPHIA update. Maybe your IDE made some automatic changes to the code? Please fix this so that we can see what you changed and review these lines.

> A few notes:
[...]

>     * data and helper files will be added to their respective repository in another step. **Question:** How does one modify which data files are being written to the /share/ directory?

I never updated the data files myself. But as far as I understand it one needs to create a new data.tar.gz file by
	* checking out the data repository
	* executing all? scripts in the repository
	* creating a tgz of the data repository
	* Uploading it to the download location

>     * A test unit for photon fields has been added and runs successfully on my system. Externally run tests (Travis) fail where the new photon field format is used as their data files are not available for now. This should change once the CRPropa-data repository has merged a pull request with these features.

I don't see a pull request for the data repository. I assume that the changes there are not conflicting with the data already there. We thus can merge them first and verify that the code builds and runs on all test systems.


>     * GitHub indicates conflicts with the master branch ever since the Boris Push has been merged.

The merge conflict should be resolved.

Froehliche-Kernschmelze:
Now I see that the PR implies to make a change in 160+k lines! That is way not the scope in which I wish to contribute but at the same moment I do not see why this happens. PR in the data repository follows.

Froehliche-Kernschmelze:
Thanks for your feedback so far! All conflicts are solved and all points of feedback have been realized. Note that I have changed the CMakeLists.txt in a way that it downloads a modified version of the data tarball that temporarily can be found in a branch of [my CRPropa-data fork](https://github.com/Froehliche-Kernschmelze/CRPropa3-data/tree/implement_custom_photon_field). The access to these files fixed all errors that occurred with the travis-ci tests and in a later step should be changed to their [original download locations](https://www.desy.de/~crpropa/data/interaction_data/data.tar.gz).

src/PhotonBackground.cpp
Froehliche-Kernschmelze: All SOPHIA methods use GeV as their internal standard unit. As not few parts of the original code are very difficult to read and to interpret, all methods directly taken from the original code are marked `SOPHIA_<methodName>` and are being based on their own unit system. Nevertheless, all public methods relating to SOPHIA do use the CRPropa unit convention now - such that the user is not exposed to more than one unit convention.

adundovi:
> Thanks for your feedback so far! All conflicts are solved and all points of feedback have been realized. Note that I have changed the CMakeLists.txt in a way that it downloads a modified version of the data tarball that temporarily can be found in a branch of [my CRPropa3-data fork](https://github.com/Froehliche-Kernschmelze/CRPropa3-data/tree/implement_custom_photon_field). The access to these files fixed all errors that occurred with the travis-ci tests and in a later step should be changed to their [original download locations](https://www.desy.de/~crpropa/data/interaction_data/data.tar.gz).
> 
> Also, a pull request [#7](https://github.com/CRPropa/CRPropa3-data/pull/7) to the CRPropa3-data repository has been opened, featuring two files associated with custom photon fields

Thanks, Mario for your contribution, especially because it is for a component which nobody wanted to touch for years.

Regarding the data file - I can give you necessary credentials to upload it, but you should send me your public GPG key so that I can send it to you in an encrypted form. My e-mail is andrej.dundovic@desy.de.

My general concern with this PR is that it modifies one of the most opaque parts of CRPropa, SOPHIA, and thus any comparison of your contribution with the original implementation just by code inspection is hardly possible. Therefore, we have to test it on various scenarios. It is not that I don't believe that you performed a lot of tests by yourself to check the validity of your contribution, but I would like to have two implementations in parallel for some transitional period that we can quickly test the new modification. Maybe copy it as a totally separate module?

TobiasWinchen:
> 
> My general concern with this PR is that it modifies one of the most opaque parts of CRPropa, SOPHIA, and thus any comparison of your contribution with the original implementation just by code inspection is hardly possible. Therefore, we have to test it on various scenarios. It is not that I don't believe that you performed a lot of tests by yourself to check the validity of your contribution, but I would like to have two implementations in parallel for some transitional period that we can quickly test the new modification. Maybe copy it as a totally separate module?

I fully agree with Andrej and already remarked this on the original PR: The original SOPHIA code needs to be available in parallel and the user should have a flag to switch ebtween them, and the new version should NOT be the default in the beginning - Maybe, after more testing, we can switch to the new version with the release of 3.2

src/PhotonBackground.cpp
```diff
@@ -84,43 +91,305 @@ double photonFieldScaling(PhotonField photonField, double z) {
...
+	- called by: SOPHIA_functs
+*/  
+	const double mass = onProton? 0.93827 : 0.93947;  // Gev/c^2
+	double sth = 1.1646;
+	double s = mass * mass + 2. * mass * x;
+	if (s < sth)
+		return 0.;
+	// upper: proton resonance masses
+	// lower: neutron resonance masses
+	const double AMRES[18] = {1.231, 1.440, 1.515, 1.525, 1.675, 1.680, 1.690, 1.895, 1.950,
```
TobiasWinchen: make data blocks `static const`

src/module/PhotoPionProduction.cpp
```diff
@@ -2,47 +2,46 @@
 #include "crpropa/Units.h"
 #include "crpropa/ParticleID.h"
 #include "crpropa/Random.h"
+#include "crpropa/PhotonBackground.h"
 
 #include <kiss/convert.h>
 #include <kiss/logger.h>
 #include "sophia.h"
+#include <stdlib.h>
```
TobiasWinchen: prefer c++ header <cstdlib> over C header

src/PhotonBackground.cpp
```diff
@@ -1,11 +1,16 @@
...
+#include "crpropa/Random.h"
+#include "crpropa/Units.h"
 
 #include <vector>
 #include <fstream>
 #include <stdexcept>
 #include <limits>
 #include <cmath>
+#include <sstream>
+#include <iostream>
```
TobiasWinchen: The module should not output o the terminal but use the logger

Froehliche-Kernschmelze:
Thanks for your review so far! Yes, I fully agree about how careful we should be with SOPHIA and any modifications to it. With an eye on what I also intend to contribute prior to the 3.2 release which is

* space- & time-dependence of photon fields (of which there exists a working implementation already) and possibly [matter distributions](https://github.com/juliaebeling/CRPropa3/tree/master/src/module) as well provided that this feature by @juliaebeling will have been merged until then
* a discretized version of SOPHIA as described in #213 of which there exists an implementation already

I feel like it could be the best strategy to **not** implement and merge a temporary module within the current CRPropa3 master but rather to have both versions tested and (compared to) in separate installations. This would not only **1)** avoid quite an amount of implementation time for a module that is deemed to be re-set to the current version of my PR (provided all features operate correctly) but also **2)** ensure no unintended interactions between the two ways of handling photon fields in the two versions of CRPropa. 
It is a matter of fact that **3)** photon fields are wired deeply and diversely into CRPropa (and thus any changes to them as well):  Changes, among others, do not only occur in SOPHIA here but also to the `PhotonBackground`, severely in the `PhotoPionProduction` and the `data.tar.gz`file (kind of supporting point 2) here).

What do you think?

src/PhotonBackground.cpp
Froehliche-Kernschmelze: iostream was an unused import and has been removed. By logger use do you mean the use of `KISS_LOG_ERROR << "message";` on **each** occasion or are statements of the kind 

https://github.com/CRPropa/CRPropa3/blob/bd39f844217a02c7b5f4d3f6a67d88d22617650a/src/module/PhotoPionProduction.cpp#L76

allright as well?

src/module/PhotoPionProduction.cpp
Froehliche-Kernschmelze: unused import, got removed

Froehliche-Kernschmelze:
> The C reimplementation of the sophia functions should be moved to a separate class
> [...]
>  so that they can be called as `SOPHIA::propEps(...)` withjout need for an instance of that class.

I see a challenge in how these functions rely on each other:
<pre>
sampleEps ---> probEps ---> functs ---> crossection ---> breitwigner
    |             |                         |----------> pl
    |             |                         |----------> ef
    |             |
    |             |-----> getPhotonDensity ---> init
    |             |                               ^
    |             |------------------------------>|
    |-------------------------------------------->|
</pre>
Note that `init(/path/to/photonField.txt)` reads in a photon field file stored at `share/crpropa/Scaling`which provides all functions mentioned above with information on how to sample photons from a field. `sampleEps`, `probEps`, and `getPhotonDensity`rely on data being provided by `init` which is called once an instance of CustomPhotonField is created. Making each function independent of the class constructor by e.g. calling `init` prior to each function call appears not to be desirable in my eyes. Any thoughts?

adundovi:
Although I have to catch up with the other contributions that you have mentioned that should be merged before 3.2 to judge the broader context better, I still propose that we look into this PR first. The main obstacle which I see with your suggestion is that not many people will go and install two versions of CRPropa in parallel, and, hence, your modification of SOPHIA will never be thoroughly tested as it should be... Thus, bringing it to the master branch should be our priority.

If I understood correctly, you wrote in (2) and (3) that your changes in the photon field implementation would already cause some compatibility issues. I briefly inspect the diff, and I don't see many points of interference in this PR that would possibly change current results (in physics). Can you explicitly point to those possible unintended interactions?

I suggest that we first discuss modifications of the photon field interface that would be required for your contribution to land in the master. Those can be split in a separate PR and can be easily inspected, right? After we merge those in the master, we can focus on a new/modified PhotoPionProduction module.

Regarding the tabular version of SOPHIA ( #213 )... I always thought it was necessary to improve the overall photo-pion performance as nobody would like to go into the SOPHIA code to see what is going on there. But now, when you already started to dig into SOPHIA and modify it, together with the fact that you measured the improvement of the tabular version over the regular version with the factor of 2 (if I remember correctly, which is not much IMHO), it could easily happen that the tabular version becomes already obsolete and unnecessary. We can profile and find slow parts of SOPHIA and re-write them to achieve that 2x speed-up if we are already in the process of modifying it, no?

Froehliche-Kernschmelze:
> Although I have to catch up with the other contributions that you have mentioned that should be merged before 3.2 to judge the broader context better, I still propose that we look into this PR first. The main obstacle which I see with your suggestion is that not many people will go and install two versions of CRPropa in parallel, and, hence, your modification of SOPHIA will never be thoroughly tested as it should be... Thus, bringing it to the master branch should be our priority.

If this strategy implies that the whole user base of CRPropa would test the modified version of SOPHIA that would be a big advantage indeed. On my mind was before that we ran dedicated tests in a smaller group and then consent on release in the master with 3.2 upon test results for everyone  else to use.

**At this point:** Does anyone have a meaningful name for a boolean flag here?
 
> If I understood correctly, you wrote in (2) and (3) that your changes in the photon field implementation would already cause some compatibility issues. I briefly inspect the diff, and I don't see many points of interference in this PR that would possibly change current results (in physics). Can you explicitly point to those possible unintended interactions?

Thanks for your positive feedback on this. In fact I see a risk in how the PhotoPionProduction would access shared files found at

https://github.com/CRPropa/CRPropa3/blob/bd39f844217a02c7b5f4d3f6a67d88d22617650a/src/module/PhotoPionProduction.cpp#L28-L40

**current master:**
These data files are written to the `data.tar.gz`and basically IRB field MFPs in either redshift dependence (indicated by IRBz files e.g. IRBz_Kneiske04.txt) or non-redshift dependence (e.g. IRB_Kneiske04.txt) - and the CMB.
**pull request**
Abandon redshift dependence flag and always load redshift-dependent data. Therefore, the naming "z" becomes obsolete such that a file e.g. IRB_Kneiske04.txt would be loaded. For short: `IRBz_Kneiske04.txt(master) = IRB_Kneiske04.txt(pull request)`.

This structure has therefore to be represented in the corresponding data.tar.gz file (of which there is only one but requires two different structures). However, there may be a workaround for the phase between testing and final release.

> I suggest that we first discuss modifications of the photon field interface that would be required for your contribution to land in the master. Those can be split in a separate PR and can be easily inspected, right? After we merge those in the master, we can focus on a new/modified PhotoPionProduction module.

The modification to the PhotoPionProduction is a direct consequence of the changes to the photon field treatment whereas those are a direct consequence of the changes to SOPHIA. Over the last 15 months I could achieve a rough understanding about SOPHIAs current structure; the interlinks of the proposed changes becomes clear on the conceptual level:

**SOPHIA(master)**
takes 1) information about the incoming nucleon 2) information about the photon background from which a photon should be drawn (within SOPHIA) and then calculate the actual nucleon-photon interaction 3) return primary and secondary particles.
1) SOPHIA can **only** sample photons from the CMB or an early version of IRB_Keiske (for each there exist a sampling method within SOPHIA). If any user would run the PhotoPionProduction on a photon field other than these two, the results are likely to *not* represent these photon fields
2) The PhotoPionProduction is, as far as I can judge that, not able to add secondary nucleons as each time a proton or neutron is returned by SOPHIA, the current canditate's energy is overwritten but there is no additional nucleon being generated as a secondary (which happens as soon as an anti-nucleon is produced):

https://github.com/CRPropa/CRPropa3/blob/bd39f844217a02c7b5f4d3f6a67d88d22617650a/src/module/PhotoPionProduction.cpp#L233-L243

**SOPHIA(pull request)**
takes 1) information on the incoming nucleon 2) the energy of a photon with which the photo-hadronic unteraction should be undergone 3) return primary and secondaries.
1) This approach is **independent** of the photon field at hand. The photon is sampled from an instance of the CustomPhotonField class (such as proposed in the PhotoPionProduction) that uses methods formerly operated in the SOPHIA code, however applicable to any photon background.
2) The proposed form of the PhotoPionProduction is different in a way that it discriminates between whether there is a primary nucleon and adds secondary nucleons to the simulations if they occur.

**summary:** There is no way to separate all changes of SOPHIA and the PhotoPionProduction as they require each other.

> Regarding the tabular version of SOPHIA ( #213 )... I always thought it was necessary to improve the overall photo-pion performance as nobody would like to go into the SOPHIA code to see what is going on there. But now, when you already started to dig into SOPHIA and modify it, together with the fact that you measured the improvement of the tabular version over the regular version with the factor of 2 (if I remember correctly, which is not much IMHO), it could easily happen that the tabular version becomes already obsolete and unnecessary. We can profile and find slow parts of SOPHIA and re-write them to achieve that 2x speed-up if we are already in the process of modifying it, no?

This seems achievable from the outside and I thought that it would be an option to e.g. have a pure C++ version of SOPHIA a few months ago. I had to abandon this idea when the overall structure of SOPHIA became clear:

Roughly 1/3 of SOPHIA (as used in the current master branch) appears to be written by, according to the author's names in the docstrings of the SOPHIA methods, G.Sigl & A.Muecke, whereas these lines of code incorporate a larger code of 2/3 the document's size called JETSET v7.4

https://github.com/CRPropa/CRPropa3/blob/bd39f844217a02c7b5f4d3f6a67d88d22617650a/libs/sophia/sophia_interface.f#L3980-L4004

which in my eyes cannot be reconstructed in C++ in reasonable amounts of time and effort.

Reconstruction of the entire SOPHIA code in C++ appears to be unlikely, however SOPHIA cannot be run parallel as written in the code at

https://github.com/CRPropa/CRPropa3/blob/bd39f844217a02c7b5f4d3f6a67d88d22617650a/src/module/PhotoPionProduction.cpp#L228-L231

and pointed out by @lukasmerten and in #213. A discretized version of SOPHIA is able to be run entirely parallel and thus avoid parallelization saturation of the PhotoPionProduction beyond 8 cores. This is where the actual increase of performance is to be expected, however the factor of x2 is nice to have though.

src/PhotonBackground.cpp
TobiasWinchen: No, throwing an exception is OK.

Froehliche-Kernschmelze:
To help everyone understand the proposed changes I came up with a strategy proposal on how we could approach all changes in three smaller pull requests, please let me know if this could be an option for you and /or if you have additional ideas. This argumentation starts with the master branch "as is".


**step 0**


* problem 0:
    a) the PhotoPionProduction is faulty such that it cannot add more than one non-antinucleonic secondary
    b) instead of adding these particles, the primary's energy is being overwritten multiple times

* goal 0:
    a) introduce discrimination between primary and secondaries in output handling of PhotoPionProduction
    b) reduce entanglement of all changes proposed in this pull request

* solution 0:
    a) rewrite that part of the secondaries' treatment

* validation 0:
    a) test very-high-energy simulations for added nucleons


**step 1**


* problem 1:
    a) SOPHIA is difficult to read / understand
    b) pull request appears unreasonably large (11+k lines changed)

* goal 1:
    a) help SOPHIA become understandable
    b) reduce size of pull request by ~ 9k lines

* solution 1: 
    a) do not modifify anything in CRPropa except SOPHIA:		
    b) **only** remove unused parts of SOPHIA / JETSET
    c) introduce uniform indentation
    d) remove unhelpful docstrings
    e) add meaningful docstrings
    f) rename variables in a better readable way
    g) replace GOTO statements by for / if/else / while
    h) optional: provide documentation on structure?
    i) provide boolean switch in PPP for the two versions of SOPHIA (default: original SOPHIA)

* validation 1:
    a) show that really nothing changed within the numerics: KS tests, etc.


**step 2**


* problem 2:
    a) SOPHIA samples photons for interaction internally
    b) only the CMB and IRB_Kneiske04 is available

* goal 2:
    a) move photon sampling from SOPHIA to CRPropa (yet keep "old" version in parallel)
    b) provide photon sampling for **any** photon field
    c) only pass primary type / energy and photon energy to SOPHIA such that SOPHIA become independent from the photon background

* solution 2:
    a) re-implement SOPHIA sampling methods in C++
    b) generalize these methods to sample **any** photon field
    c) but these sampling methods in CustomPhotonField class
    d) provide unified photon field format as .txt file
    e) provide additional module PhotoPionProduction_modified in parallel to native PPP
    f) provide additional lib SOPHIA_modified in parallel to native SOPHIA
    g) remove boolean switch as PPP uses SOPHIA and PPP_modified uses SOPHIA_modified
    h) modify data.tar.gz file such that both versions of PPP are supported

* validation 2:
    a) KS tests of the various C++ vs SOPHIA methods (existing already)
    b) KS tests of output data samples: previous methods PPP vs PPP_modified (partially existing)

lukasmerten:
@Froehliche-Kernschmelze this looks like a good roadmap for me. Although it looks like a lot of extra work for you the only way to convince people that your reimplementation/extension works is to make all changes clear and transparent.

However, I was more involved into your developments and I am therefore probably biased. So maybe @TobiasWinchen @adundovi can comment on your plan.

adundovi:
@lukasmerten I am convinced that Mario is doing a great job and his contribution should definitely land in the master branch, and I will try to help with that. But before we do so, we have to be cautious how we perform this merge as it modifies one of the core components, especially when we know that there are already several published results which heavily relied on it. This abrupt in physical results should be clearly documented and explained to our users; otherwise, we lose credibility. The main issue here is not in Mario's contribution but in the portion of our existing code (SOPHIA) which is very tricky to change as we took it for granted - in the form of the black box. Now, since Mario wants to introduce changes, we have to understand it better. That is why I am conservative here.

@Froehliche-Kernschmelze If I understand this correctly your step 0 is actually a bug report for physics of photo-pion? In that case, if the current behaviour is faulty how do you expect that we validate the results? Against what we should compare it?

With other steps I generally agree, although I would recommend writing unit tests for SOPHIA explicitly so that we can to some extent be sure that the introduced changes are not changing its behaviour (this procedure is usually employed when one is refactoring code). Because of GOTOs, global variables, and other nasty techniques, the SOPHIA code is highly non-linear and thus changes to it can be particularity unpredictable.

Also, I would suggest that you don't tie your individual work with this PR. You have all the freedom to continue your work and publish physical results created with your fork of CRPropa if you like. It is ungrateful to work with other people in academia in the context when you depend on them since we all have different deadlines and goals, and our job contracts are generally rather short.

TobiasWinchen:
The proposed roadmap looks great and I hope that this is not too much work for you? Also I would suggest to *NOT* change the docstrings / variable names and replace the `gotos` with an other logic: this could be a huge source of errors and make it harder to understand the changes. IMO removing unused code + adding additional whitespace + comments in a unique way that marks the comment as your addition (e.g. `c Mario: .... `) should be enough. This would also simplify the validation of step 1.

However, the most important issue seems right now to understand the reported physics-bug with all the implications. How does it affect the results produced so far? Any additional information on this would be great.

Froehliche-Kernschmelze:
Thanks for your replies! No worries about invested time, this is a large PR and I am happy to contribute. Not worth mentioning that each step has to be entirely transparent. I will open an issue on what I think there is with the PPP with as many details as I have.

lukasmerten:
Here, the same problem occurs as in #226 and has to be resolved, too.

Froehliche-Kernschmelze:
Now that step 0 is taken, there is the agreement to introduce a second version of SOPHIA that includes my modifications in parallel to its original version. This version is agreed to be toggled by some boolean switch in the PhotoPionProduction, defaulting to original SOPHIA. Yet I am having trouble to actually link this second version of SOPHIA to CRPropa and after going through all sorts of attempts I am asking for a few thoughts of yours.

**Objective:**
Create a parallel instance of SOPHIA, called SOPHIAMOD. Have it imported and used in the PhotoPionProduction.

**State of progress**
I can establish that second version of SOPIHA and also import it in the PhotoPionProduction. However, I can not use it.

**How SOPHIA is implemented / linked**
The version of CRPropa as listed in the master branch of the repository should contain all information necessary for the programme to install and link SOPHIA properly. Mentions of SOPHIA occur in the following places as retrieved by `grep -ri "sophia" (comments in which SOPHIA is mentioned are cut out here):

(main CMakeLists)
CRPropa3/CMakeLists.txt:# SOPHIA (provided)
CRPropa3/CMakeLists.txt:add_subdirectory(libs/sophia)
CRPropa3/CMakeLists.txt:list(APPEND CRPROPA_EXTRA_LIBRARIES sophia gfortran)
CRPropa3/CMakeLists.txt:list(APPEND CRPROPA_EXTRA_INCLUDES libs/sophia)

(sophia.h)
CRPropa3/libs/sophia/sophia.h:#ifndef _SOPHIA_H
CRPropa3/libs/sophia/sophia.h:#define _SOPHIA_H
CRPropa3/libs/sophia/sophia.h:void sophiaevent_(int& channel, double& inputenergy, double momentum[][2000],

(sophia_interface.f)
CRPropa3/libs/sophia/sophia_interface.f:      subroutine sophiaevent(nature,Ein,OutPart,OutPartType,NbOut

(SOPHIA CMakeLists)
CRPropa3/libs/sophia/CMakeLists.txt:add_library(sophia STATIC
CRPropa3/libs/sophia/CMakeLists.txt:    sophia_interface.f
CRPropa3/libs/sophia/CMakeLists.txt:set_target_properties(sophia PROPERTIES COMPILE_FLAGS -fPIC)

(import / use in PPP)
CRPropa3/src/module/PhotoPionProduction.cpp:#include "sophia.h"
CRPropa3/src/module/PhotoPionProduction.cpp:		sophiaevent_(nature, Ein, momentaList, particleList, nParticles, z, background, maxRedshift, dummy1, dummy2, dummy2);

in all these files I have implemented additional entries of the same syntax yet mentioning an object called SOPHIAMOD instead, including a new directory called sophiaMod in the /libs folder. I added an import in the PPP `#include "sophiaMod.h"` and up to this point the code compiles flawlessly.

**Problem**
If I want to call the modified version of SOPHIA it returns "undefined reference to sophiaeventMod_" during compilation (yet this function exists at that point).

Do you have any educated guesses what the solution could be? Otherwise I would be happy to provide a more detailed description of this problem. There is a [branch](https://github.com/Froehliche-Kernschmelze/CRPropa3/tree/SOPHIA_update) in my repository that currently has these settings. Also, the PPP has got a version of sophiaeventMod (toggled out) that if enabled would produce the error described.

adundovi:
Give me one day - I'll write a CMake option for it from which we would be able to switch between sophia and sophiamod.

adundovi:
Sorry for the delay.

I somehow repeated the procedure you described and it worked for me.

Or to be more specific:
1. copy libs/sophia to libs/sophiamod
2. rename sophia.h to sophiamod.h in libs/sophiamod
3. rename sophia to sophiamod in CMakeLists.txt located in libs/sophiamod
4. copy PhotoPionProduction.h to PhotoPionProductionMod.h and PhotoPionProduction.cpp to PhotoPionProductionMod.cpp
5. sed -i "s/PhotoPionProduction/PhotoPionProductionMod/g" and  "s/PHOTOPIONPRODUCTION/PHOTOPIONPRODUCTIONMOD/g" in two of those
6. change `#include "sophia.h"` to `#include "sophiamod.h"` in PhotoPionProductionMod.cpp
7. add `#include "crpropa/module/PhotoPionProductionMod.h"` in include/CRPropa.h
8. in python/2_headers.i add `%include "crpropa/module/PhotoPionProductionMod.h"`
5. edit the main CMakeLists.txt by adding `src/module/PhotoPionProductionMod.cpp` in the `add_library(crpropa ...` list and insert:
```
# SOPHIAmod (provided)
add_subdirectory(libs/sophiamod)
list(APPEND CRPROPA_EXTRA_LIBRARIES sophiamod gfortran)
list(APPEND CRPROPA_EXTRA_INCLUDES libs/sophiamod)
```
[Here](https://github.com/adundovi/CRPropa3/commit/7fa017f7e3ba1d4cbf9268bb674cf0097466506c) you can see the diff in my branch.

I can import and use both modules in python in parallel after doing this.

Froehliche-Kernschmelze:
Thanks a real lot to @adundovi !

[A branch of mine](https://github.com/Froehliche-Kernschmelze/CRPropa3/tree/SOPHIA_update) now contains a modified version of SOPHIA following the suggestions of @TobiasWinchen where only unused parts of the code have been removed. No indentation / other style specifications / goto removal or new functions and the like have been realized.

This is to tackle step 1 of the agreed roadmap. It should be shown that nothing changed to SOPHIA by removing unused parts of the code.

This modified version can be used via a new temporary module called `PhotoPionProcuctionMod` that is entirely identical to the regular PPP except that it uses `SOPHIAmod`.

At this point, how shall we proceed? I could create a pull request with this modification or conduct further tests on the scientific identity of the two versions.

So far I have conducted Kolmogorov-Smirnov tests on each of SOPHIAs output particles between two samples 1) drawn from the original SOPHIA in the master branch (for comparison of the KS parameters) and 2) drawn from master and SOPHIAmod. Results suggest identity between all of the produced secondary output particles, i.e. photons, electrons and neutrinos (hadrons not yet tested).

I am happy to share any setup or tests I have made. Any opinions?

adundovi:
Tests are certainly welcome if not necessary. I suggest that you write them in the form of unit tests and place them in `test/testPhotoPionProduction.{py.cpp}` for instance. Test any aspects of PhotoPionProduction that you can think of being relevant for CRPropa usage but which are not time-consuming (not running longer than ~10s).

It would be great if we could control the SOPHIA random number generator, i.e. to fix its seed so that both modules (the old and the new) produce deterministic results for given inputs. In that way, we would avoid time and resource consuming unit tests which are based on statistical comparisons. Moreover, if you have some theoretical estimates which should be easy to reproduce with SOPHIA, they could be included, too. Both modules should pass all these tests.

After ending the basic testing step, we should proceed by including your PPPmod in the master repo following manually testing in different "large-scale scenarios". If everything goes well we should make PPPmod as the default PPP module, while the current one label as Legacy and after some time to put a deprecated notice and remove it from the source.

Froehliche-Kernschmelze:
There is progress with two aspects:

1) a) The RNG of SOPHIA is always deterministic as it will always use the same seed regardless of any setup in the PhotoPionProduction. This means, that if you called sophiaevent_(...) in lets say two separate instances of CRPropa, e.g. in two ipython terminal windows, the result of sophiaevent_(...) would always be exactly the same. Therefore, there can be conducted deterministic unit tests (which I will come up with eventually).
b) For this to test, I have implemented an external interface method for SOPHIA as a member of the PhotoPionProduciton in my current [working branch](https://github.com/Froehliche-Kernschmelze/CRPropa3/tree/SOPHIA_update). This method can be used for debugging and testing in unit tests as well. It returns a vector of the output particles' IDs (first half of vector) and energies (second half of vector).

2) I implemented your @adundovi suggested setup to have PhotoPionProductionMod include sophiaMod. However, although everything compiles just fine, there is a subtle yet critical bug as it seems that the different sophiaevent methods aren't yet properly linked. It seems like now there is always the modified version being called **regardless** of whether PhotoPionProduction or PhotoPionProductionMod call their versions of SOPHIA.
For this to verify I have pushed your suggested setup with two almost identical versions of SOPHIA with the difference that sophiaMod would print "mod" to the console. This is the code I have run in an ipython shell and should show that PhotoPionProduction actually calls sophiaMod:
```
In [1]: from crpropa import *
In [2]: ppp = PhotoPionProduction()
In [3]: out = ppp.sophiaEvent(True, 1e12*GeV)
 mod
In [4]: for x in out:
   ...:   print(x)
   ...:
1000000010.0
14.0
12.0
-11.0
-14.0
148.03603804212008
3.242941249811455
2.3102363906818386
1.075986525593002
5.552446491793585
```
Note that you should get exactly the same output as SOPHIAs RNG always starts with the same seed. I observe that if you renamed sophiaevent_ to sophiaeventMod_ in libs/sophiaMod/sophiaMod.h, libs/sophiaMod/sophiaMod_interface.f and include/crpropa/module/PhotoPionProductionMod.cpp, there will be an error upon compilation stating "libcrpropa.so: undefined reference to sophiaeventMod_". Do you have any thoughts on this?

lukasmerten:
> a) The RNG of SOPHIA is always deterministic as it will always use the same seed regardless of any setup in the PhotoPionProduction.

That might be problematic, as one usually assumes that the events are independent of each other. With this bug, to my understanding combining the results of two independent CRPropa runs does not reduce the MonteCarlo error for the combined results. 

Although this helps us now to test the code we should definetly change that behavior in the future, i.e. using the random seed from the CRPropa Random class also in the SOPHIA RNG or replace the SOPHIA RNG by the Random-class.

Froehliche-Kernschmelze:
So far, following your @adundovi approach, I can tell that the compiler randomly links to either sophia or sophiaMod. Hence, PhotoPionProduction and PhotoPionProductionMod **both** use the same version of either sophia or sophiaMod and there is no separate access.

antoniusf:
**Re: the linker problems**

I've only taken a brief look at this, so feel free to disregard what I'm saying if you've already thought these points through.

As far as I can tell, the functions that `sophia` and `sophiamod` define have the same names. Is this correct? If it is, then I'm pretty sure that's a problem.

This is because the linker only resolves functions by their name; it has no information on where a function is meant to come from. (Consider that `#include` literally just copy-pastes the header file into the source. After that, no part of the compiler cares where the header came from.) So if a function has the same name in both `sophia` and `sophiamod` the compiler just won't know which function you mean... does this explanation make sense?

(Even so, it is a little bit confusing that the compiler didn't report this as an error, but it seems problematic nonetheless.)

Froehliche-Kernschmelze:
Thanks a lot @antoniusf for looking into this. I indeed have considered this and e.g. renaming `"sophiaevent_" to "somethingElse"` would result in a compilation error saying that functions (all SOPHIA ones) would be defined in different places, i.e. in `/sophia/...` and `sophiaMod/...`. I think we need a way here to tell the framework that there are supposed to be two different libs. Any idea?

antoniusf:
> would result in a compilation error saying that functions (all SOPHIA ones) would be defined in different places

This is the error I would have expected to occur with unchanged function names... Did you change all names or just a few ones? (To get rid of this error, you would definitely have to change all of them, or at least all that are exported... which are quite a few, it appears??)

In any case, I think that renaming all of the exported functions (and constants) is probably the most technically simple option. Unfortunately, this appears to be quite a bit of work, so here's some other possibilities I can think of right now:

–  I believe that CRPropa currently uses static linking for sophia. In principle, it might be possible to split up the static linking process such that PhotoPionProduction/Mod are linked first against their respective versions of sophia (such that all function names are resolved), and only then linked together with the rest of CRPropa. However, I do not know how to make gcc or cmake do this – maybe you could build PhotoPionProduction in a separate folder, with its own CMakeLists, and then integrate that somehow?

 – The other option might be to manually load the correct version of sophia at runtime. While this would mean that you don't have to muck around with gcc, it is not platform-independent and requires you to write extra code to load each function. ([see here](https://stackoverflow.com/a/7626550) for what this would look like)

antoniusf:
Update on renaming the functions: Since only the `sophiaevent_` function appears to be needed by external code, it might be possible to set up the library to export only this symbol.

Currently, `libsophia.a` contains quite a lot of symbols, so that's probably where the problems come from. I've done a brief search to see if hiding these symbols is possible; it appears that C supports this through its `static` keyword, but I haven't been able to find anything comparable for Fortran.

However, before proceeding with any of the approaches above, it might be prudent to re-evaluate if the capability to switch between sophia and sophiaMod at run-time (as opposed to compile-time) is worth the extra effort and added complexity. Almost every approach I can think of either complicates the build process (step-by-step linking), causes issues with platform independence (dynamic linking, symbol hiding for static libraries), or introduces the possibility for (additional) bugs in the sophia code (renaming all exported symbols). This is in addition to the time necessary to make each of the approaches work, which shouldn't be neglected either.

adundovi:
Sorry for the long absence on this topic. I'm looking into it.

adundovi:
Hi all!
You are absolutely correct, I've encountered all the problems mentioned above. I'm sorry for not testing my "split" proposal before proposing it. Since then, I've been thinking about how to avoid both the problem with linking and the problem with unnecessary code duplication. Now I suggest the following:

1. split PPP into two classes as in the first proposal, so that the following example works:
```python
from crpropa import *

ppp = PhotoPionProduction()
c = Candidate()
c.current.setId(nucleusId(1, 1))
c.current.setEnergy(100 * EeV)
c.setCurrentStep(1000 * Mpc)
ppp.process(c)

pppmod = PhotoPionProductionMod()
c = Candidate()
c.current.setId(nucleusId(1, 1))
c.current.setEnergy(100 * EeV)
c.setCurrentStep(1000 * Mpc)
pppmod.process(c)
```

2. Make modifications within the sophia library in the form of new functions:
```c++
extern "C" {
void sophiaevent_(int& channel, double& inputenergy, double momentum[][2000],
                int id[], int& n, double& redshift, int& photonbackground, double& maxz,
                int&, double[], double[]);
}

extern "C" {
void sophiaeventmod_(int& channel, double& inputenergy, double momentum[][2000],
                int id[], int& n, double& redshift, int& photonbackground, double& maxz,
                int&, double[], double[]);
}
```

3. You shouldn't edit sophia_interface.f, but create a new file, for example, sophia_interface_mod.f, which include functions and routines from sophia_include.f (with the "include" command), and then either use header guards, or replace "sophia_interface.f" with "sophia_interface_mod.f" in CMakeLists.txt file in the sophia directory.

I tested this idea by putting different print output in sophiaevent_ and sophiaeventmod_, and the above python script gives:
```
$ python testppp.py
 sophiaevent: normal version
 sophiaevent: normal version
 sophiaevent: normal version
 sophiaevent: normal version
 sophiaevent: mod version
 sophiaevent: mod version
 sophiaevent: mod version
 sophiaevent: mod version
 sophiaevent: mod version
 sophiaevent: mod version
```

Froehliche-Kernschmelze:
Thanks a lot @adundovi for the proposal! Could you push a working example of what you propose? That would help me a lot!

adundovi:
Here it is: https://github.com/adundovi/CRPropa3/tree/newppp

Froehliche-Kernschmelze:
Thanks a lot @adundovi for pushing your example. Based on your changes, I am able / intend to continue with step 1 of the agreed [road map above](https://github.com/CRPropa/CRPropa3/pull/220#issuecomment-466701902) which is to clear up SOPHIA to introduce custom photon fields into CRPropa.

This sub-step is to demonstrate that, despite having an internal RNG:
* SOPHIAs event generator is strictly deterministic in its arguments
* SOPHIAs internal photon sampling is just as deterministic in its arguments
This step **is only supposed to point out aspects of SOPHIAs behaviour**, thus not proposing actual changes to be merged into the master branch (for now).

Once we agree on this step, further changes to SOPHIA are much easier to validate as we would not have to rely on statistical tests as @TobiasWinchen has pointed out.

The modifications can be found in [this branch](https://github.com/Froehliche-Kernschmelze/CRPropa3/tree/SOPHIA_update) of my repository.

This modification adds the following features:
1) the PhotoPionProduction has been given the ability to separately call SOPHIA without to propagate a candidate via a new class method
`std::vector<double> PhotoPionProduction::sophiaEvent(bool onProton, double Ein, double z=0.)`
which returns a vector containing SOPHIAs output particle IDs (PDG convention, in first half of the vector) and energies (SI, in second half of the vector).
2) The SOPHIA code itself has not been changed, except that the photon energy (in eV) that has been internally sampled, is printed.

How non-randomness can be shown:
* in an ipython terminal run
```
In [1]:  from crpropa import *
In [2]:  ppp = PhotoPionProduction()
In [3]:  for x in ppp.sophiaEvent(True, 1e11 * GeV):
   ...:      print(x)
```
The output should be:

 eps =    1.4145363205544225E-003
1000000010.0
14.0
12.0
-11.0
-14.0
12.135608149326126
0.11742056103982144
0.4679922135766474
3.0552110944588775
0.2455328515985278

* repeat this in a new console to get the same output.
* this works for every parameter constellation

**Can you verify this behaviour?**

If so, I suggest cleaning out unused parts of the SOPHIA code to make it way more readable and show that we will get the very same output after this. Then, we could move on to manipulating physics inside SOPHIA.

adundovi:
@Froehliche-Kernschmelze OK, I can confirm the behaviour with your git branch:
```
 eps =    1.4145363205544225E-003
1000000010.0
14.0
12.0
-11.0
-14.0
12.135608149326126
0.11742056103982144
0.4679922135766474
3.0552110944588775
0.2455328515985278
```

Froehliche-Kernschmelze:
Thanks for your confirmation. Now I suggest to either (or both!)
1) optional: (since we are at this point anyway) Clean out code in SOPHIA not used within CRPropa and proof it has not changed anything, i.e. reproducing the above mentioned output
2) Remove SOPHIAs internal photon sampling in sophiamod, thus leading to a signature where the photon energy eps now is an external parameter (functions providing eps follow in a next step)
```sophiaeventmod(bool onProton, double Ein, double eps, outputEnergy, outPartID, nOutPart)```

Which is the most viable way for you?

**Comment by adundovi** [comment]
If I understand you correctly, the first point is about cleaning/refactoring the FORTRAN code, and the second one is about to introduce new features / change the interface, right?

As far as I'm concerned, you can do whatever you want and what you find necessary in the modified PPP as long as we have both versions of PPP available in parallel (the new and the original). However, to ease testing of the modified/new version (when you go beyond cleaning the code, and start to introduce new features/change the interface), you should provide an additional interface/wrapper which, by construction, would use the same usage logic, and produce identical output as the old procedure. If this approach is possible, of course. If not, then we have to do the 1st and the 2nd point separately, in which you will have to wait with the inclusion of the 2nd until the 1st is well tested.

**Comment by Froehliche-Kernschmelze** [comment]
Exactly, the first would introduce changes to the original SOPHIA code by cutting away code that never gets used by CRPropa. This procedure may be lengthy time-wise but worth the review effort as this would reduce the code by approximately a factor of 2, i.e. ~8k lines of code. However, this step can still be done after having introduced the parallel version of SOPHIA.

Given that time is the scarcer resource, I would directly go to modifying the newly added interface for sophiamod (leaving the original completely unchanged and in parallel). The idea behind this step is to show that the photon sampling for SOPHIA can be done externally - and if done correctly, exactly reproduces the results of original SOPHIA.

Thus follows a detailed yet straight forward explanation of how I intend to do this:

* I have pushed [a version of CRPropa](https://github.com/Froehliche-Kernschmelze/CRPropa3/tree/SOPHIA_update) that introduces a new module called PhotoPionProductionMod that is identical to PhotoPionProduction except that it makes use of a new call to SOPHIA called sophiaeventmod (taken from and following @adundovi)
* I have modified the original SOPHIA, here written in `sophia_legacy.f` for purposes of demonstration such that it disregards which photon energy has been internally sampled but always selects eps=1e-3 eV. This value is being printed to the console for clarification (lines 16812 and 16813)
* PhotoPionProduction and PhotoPionProductionMod both contain a class method called "sophiaEvent". The signature of these methods are
`PhotoPionProduction::sophiaEvent(bool onProton, double Ein, double redshift=0.)`
and
`PhotoPionProductionMod::sophiaEvent(bool onProton, double Ein, double eps)`
* if PhotoPionProduction::sophiaEvent is called, the original SOPHIA code is called and calculates the interaction event for a photon energy of 1e-3 eV - if PhotoPionProductionMod::sophiaEvent is called, it calls sophiamod as written in "sophia_interface.f"

Here comes the critical part:
* sophiaeventmod (lines 13-103 in sophia_interface.f) is identical to sophiaevent (lines 16699 to 16846 in sophia_legacy.f) with the exception that there is no call to any photon sampling function in sophiaeventmod (as this is an external parameter now) as opposed to a call to the photon sampling routines in lines 16780 to 16803 in the original code. The signature of sophiaeventmod has changed accordingly.
* To proof that both approaches yield the same results, we make use of the fact that SOPHIA actually is deterministic in its arguments (at least seen from how it is integrated into CRPropa). Hence, in an ipython terminal run:

```
In [1]: from crpropa import *
In [2]: ppp = PhotoPionProduction()
In [3]: for x in ppp.sophiaEvent(True, 1e11 * GeV):
   ...:     print(x)
   ...:
```
the output should be

 eps =    1.0000000000000000E-003
1000000010.0
14.0
12.0
-11.0
-14.0
13.097485723026773
0.26277784161314893
0.852441829038488
0.7331417821599351
1.0759176941616528


In another terminal (to re-set the RNG counter as both versions call sophia_legacy.f in the end)
```
In [1]: from crpropa import *
In [2]: pppmod = PhotoPionProductionMod()
In [3]: for x in pppmod.sophiaEvent(True, 1e11 * GeV, 1e-3 * eV):
   ...:     print(x)
   ...:
```

which should yield

  sophiaeventmod version
1000000010.0
14.0
12.0
-11.0
-14.0
13.097485723026773
0.26277784161314893
0.852441829038488
0.7331417821599351
1.0759176941616528

This works for each combination of parameters that lie over the threshold CM energy for photo-pion production

Summary of this step:
* PhotoPionProduction and its mod version are identical except that they call either sophiaevent or sophiaeventmod, respectively.
* sophiaevent is the original version whereas sophiaeventmod does not call any photon sampling routine internal to SOPHIA but gets it externally
* This is to show that with this modification one can have both versions of SOPHIA in parallel whereas the modified version lets you choose the photon energy of a scattering photon. This energy can later be sampled in future methods intrinsic to the PhotoPionProduction and not SOPHIA, allowing for variable photon fields

Can this be verified by anyone?

**Comment by adundovi** [comment]
I've checked your changes. And it works as you presented, except one detail, you should probably change the data prefix in the mod version of PPP to use the same data files as the original PPP:
```c++
        if (haveRedshiftDependence)
                initRate(getDataPath("PhotoPionProduction/rate_" + fname.replace(0, 3, "IRBz") + ".txt"));
        else
                initRate(getDataPath("PhotoPionProduction/rate_" + fname + ".txt"));

```
BTW how do you plan to integrate `performInteraction` with `sophiaEvent` later?

Froehliche-Kernschmelze:
Thanks for looking into this!
...oops, that must have been a copy & paste gone wild. Changes to the data files have been made accordingly.

My suggestions on how to proceed / integrate are the following:
1) We either remove PPP::sophiaEvent again in the future as it was supposed to show behaviour of SOPHIA without input of CRPropa itself.
2) We keep and embed it into performInteraction, replacing the current way to call SOPHIA. The advantage I see doing this is that a user could use PPP(/mod)::sophiaEvent outside of a simulation as well.

Do your concerns evolve around how to provide eps? Because since we are at the same page up to here, this would be the next step(s). These could look like the following
1) Implement photon sampling into CRPropa (let's say to PhotonBackground.cpp), taking SOPHIA methods and translate them into C++ (already achieved). This could be done via a new class, e.g. CustomPhotonField, that provides a method like 
`sampleEps(bool onProton, double Ein, double z)` (already achieved)
2) This photon sampling gets verified against SOPHIA methods (already achieved)
3) make PPPMod call these sampling routines and pass this to sophiaEvent (already achieved)

These points are pushed in another branch but I am happy to provide this step-by-step. I can also elaborate on how custom photon sampling would work in CRPropa before continuing.

Do you have a preferred way to move on?

adundovi:
Ciao @Froehliche-Kernschmelze ,
I agree with the second decision - to keep sophiaEvent as a separate method in both classes (the new and the old). I'm also fine with the second part of your comment as long as it affects only PPPMod.

lukasmerten:
Does anyone of @adundovi, @MHoerbe or @TobiasWinchen know if there is still something in this PR that was not merged with different pull requests? 

I tried to quickly go through the discussion again but could come to a fast conclusion. So any hint would be appreciated.

MHoerbe:
> Does anyone of @adundovi, @MHoerbe or @TobiasWinchen know if there is still something in this PR that was not merged with different pull requests?
> 
> I tried to quickly go through the discussion again but could come to a fast conclusion. So any hint would be appreciated.

I remember that we kept this PR open but eventually agreed to never merge it. Our outline was:
1. the changes I proposed here are too big and deep
2. nonetheless, we wanted to have all changes (and a few future ones, see below) implemented which are proposed here
3. hence, we open, review, and merge new and smaller PRs until everything proposed here is in the main branch 

NOTE: over the years more features came to life and I had documented their implementation in my [PhD thesis](https://hss-opus.ub.ruhr-uni-bochum.de/opus4/frontdoor/index/index/year/2021/docId/8071).

Afaik the following changes are in fact implemented now:
1. cut out photon sampling from SOPHIA, streamline and generalize it to any photon field, and make it a feature of PhotonField
2. abolish PhotonField enums and introduce proper, tabulated PhotonField classes
3. enhance the API such that everyone may create custom photon fields (+ update the data repo, e.g. calc_* files)
4. interaction tags

Super good work so far; I am very happy to see this come to life in the main branch.

What remains, partly with respect to this PR, is
1. replace the Fortran version of SOPHIA by a [C++ version](https://github.com/CRPropa/sophia_next). It is operational and tested but I guess every further opinion helps
2. make the C++ version of SOPHIA threadsafe (which the version above is not yet). Alternatively, have the entire code discretised, I remember others at DESY(?) having approached this
3. introduce space- and time dependence on all interactions by introducing a 4D grid overlay. This is simply an extension to the domestic [Grid Class](https://github.com/CRPropa/CRPropa3/blob/5364e8af1b7d766085d0bbb697c185cb2a388047/include/crpropa/Grid.h#L137) to include time as well and would be passed to an interaction if desired.
4. hadronic interactions, yet I have seen that work is ongoing and I do not overview the entirety of progress here.

**TL;DR**
we can safely close this one if the 4 points above if:
1. the remaining proposed changes are discussed. Shall we opt out for all of these, this PR can be closed right ahead
2. if any of the above remaining changes are to be realised, they should go into a separate issue or even a milestone. Then we can close this PR.

lukasmerten:
Thanks @MHoerbe for the very comprehensive summary.

I like the idea to discuss open points of this PR and then move the interesting ones into issues. I will bring it up in the next CRPropa meeting and make the necessary changes afterwards.

PR #222, opened 2019-02-14
Boris push

reichherzerp:
Dear all,

With this Pull request, I would like to make the new PropagationBP available to everyone. This module is based on the Boris push algorithm, which is standard for particle pushing in plasma simulation codes. @GilbertWilbert started the project as documented in the commits below. 

PropagationBP has the possibility to use the adaptive step size adjustment known from PropagationCK as well as an optimized variant with fixed step size.

A detailed comparison between the PropagationBP and the PropagationCK is given in the example notebook presented in the Pull Request ([Propagation comparison of CK and BP #6](https://github.com/CRPropa/CRPropa3-notebooks/pull/6)). There, the great advantages of this algorithm are also demonstrated because of its energy-conserving properties togehter with its faster calculation.

Thanks in advance and best wishes,
Patrick


lukasmerten:
As promised in the last CRPropa meeting we found someone to finish the work of @GilbertWilbert on the Boris Push.  Please have a look on the code and especially also on the provided example. 

Using this new code might save you a significant amount of computation time. We, @reichherzerp and I, want to stress that in magnetic fields with large curvature radii (much larger than the step length) the conventional propagation using PropagationCK does not differ to much from results of the PropagationBP module. Meaning all results of lage scale extra-galactic simulations should be save. However, when CRPropa is used in environments where the magnetic field is strong enough to confine the particles effectively we strongly recommend to use the PropagationBP class in the future.

src/module/PropagationBP.cpp
```diff
@@ -0,0 +1,208 @@
...
+    }
+
+    // with a fixed step size
+    PropagationBP::PropagationBP(ref_ptr<MagneticField> field, double minStep) :
+            minStep(0) {
+        setField(field);
+        setTolerance(0.42);
+        setMaximumStep(minStep);
+        setMinimumStep(minStep);
+
```
TobiasWinchen: delete whitespace

src/module/PropagationBP.cpp
```diff
@@ -0,0 +1,208 @@
+#include "crpropa/module/PropagationBP.h"
+
+#include <sstream>
+#include <stdexcept>
+#include <vector>
+
+namespace crpropa {
+
+
```
TobiasWinchen: delete whitespace

src/module/PropagationBP.cpp
```diff
@@ -0,0 +1,208 @@
...
+            candidate->setCurrentStep(step);
+            candidate->setNextStep(maxStep);
+            return;
+        }
+
+        // further particle parameters
+        double z = candidate->getRedshift();
+        double m = current.getEnergy()/(c_light * c_light);
+
+
```
TobiasWinchen: delete whitespace

src/module/PropagationBP.cpp
```diff
@@ -0,0 +1,208 @@
...
+        // if minStep is the same as maxStep the adaptive algorithm with its error estimation is not needed and the computation time can be saved:
+        if (minStep == maxStep){
+            double step = minStep;
+            Vector3d pos = current.getPosition();
+            Vector3d dir = current.getDirection();
+            // half leap frog step in the position
+            Y yOut = dY(pos, dir, step, z, q, m);
+
+
```
TobiasWinchen: delete whitespace

src/module/PropagationBP.cpp
```diff
@@ -0,0 +1,208 @@
...
+        // check if field is valid and use the field vector at the
+        // position pos with the redshift z
+        if (field.valid())
+            B = field->getField(pos, z);
+
+        return B;
+    }
+
+    double PropagationBP::errorEstimation(const Vector3d x1, const Vector3d x2, double step) const {
+
```
TobiasWinchen: delete whitespace

test/testPropagation.cpp
```diff
@@ -92,6 +100,242 @@ TEST(testPropagationCK, neutron) {
...
+
+// Test the numerical results for parallel magnetic field lines along the z-axis
+TEST(testPropagationBP, gyration) {
+    PropagationBP propa(new UniformMagneticField(Vector3d(0, 0, 1 * nG)));
+
+    double step = 10. * Mpc;  // gyroradius is 108.1 Mpc
+    propa.setMaximumStep(step);
+    propa.setMinimumStep(step);
+
+
```
TobiasWinchen: delete whitespace

test/testPropagation.cpp
```diff
@@ -92,6 +100,242 @@ TEST(testPropagationCK, neutron) {
...
+	propa.process(&c);
+
+	EXPECT_DOUBLE_EQ(minStep, c.getCurrentStep());  // perform minimum step
+	EXPECT_DOUBLE_EQ(5 * minStep, c.getNextStep());  // acceleration by factor 5
+
+}
+
+TEST(testPropagationBP, exceptions)
+{
+
```
TobiasWinchen: delete whitespace

test/testPropagation.cpp
```diff
@@ -1,10 +1,17 @@
 #include "crpropa/Candidate.h"
 #include "crpropa/ParticleID.h"
 #include "crpropa/module/SimplePropagation.h"
+#include "crpropa/module/PropagationBP.h"
 #include "crpropa/module/PropagationCK.h"
 
 #include "gtest/gtest.h"
 
+#include <string>
```
TobiasWinchen: string and iostream are not needed.

test/testPropagation.cpp
```diff
@@ -92,6 +100,242 @@ TEST(testPropagationCK, neutron) {
...
+	p.setPosition(Vector3d(0, 0, 0));
+	p.setDirection(Vector3d(0, 1, 0));
+	Candidate c(p);
+	c.setNextStep(0);
+
+	propa.process(&c);
+
+	EXPECT_DOUBLE_EQ(minStep, c.getCurrentStep());  // perform minimum step
+	EXPECT_DOUBLE_EQ(5 * minStep, c.getNextStep());  // acceleration by factor 5
+
```
TobiasWinchen: delete whitespace

reichherzerp:
Thank you for the fast merge! Did I forget to do something or did something wrong because I'm not listed as a contributor for the Boris push?

TobiasWinchen:
I *assume* that the mail address you used for the git commits is not associated with your account. You can either add the address to your github account, or modify the .mailmap file in the repository to map your addresses.

reichherzerp:
Unfortunately, we (@antoniusf and me) have noticed that there could be a problem with the following commit:
https://github.com/CRPropa/CRPropa3/commit/646422dbbfc8909542419c986ade5182623ba7e0

Although this pull request should only have added +593 -1 lines as indicated above, the list of contributors includes GilbertWilbert with 10 commits 220,490 ++ 122 -- which can't be. I obviously didn't treat that right when merging Mike's Branch into mine. Even if the files should be correct, the statistics are incorrect. I'm sorry for this. What do you think about this?

TobiasWinchen:
There were many wrong files in this commit that were added by GilbertWilbert and removed in your merge - I didn't noticed as I didn't checked all individual commits but only the final request. For the future, we should squash the individual commits in merging a pull request. For now I removed the files from the git history using filter-branch.

reichherzerp:
Ok, thank you very much!

adundovi:
@reichherzerp @TobiasWinchen 
Any reason why the order of arguments in PropagationBP differs from PropagationCK?
These are constructors of both currently ([doc](http://crpropa.github.io/CRPropa3/group__Propagation.html)):

`PropagationCK (ref_ptr< MagneticField > field=NULL, double tolerance=1e-4, double minStep=(0.1 *kpc), double maxStep=(1 *Gpc))`

`PropagationBP (ref_ptr< MagneticField > field, double minStep, double maxStep, double tolerance=1e-4)`

I would like that that we can just change "CK" to "BP" and vice versa, and not also to change the order of arguments (like tolerance, min and max step). DiffusionSDE respects the PropagationCK pattern, and since PropagationBP is the newest newcomer to the CRPropa family (less python scripts will be broken), I suggest that we change the order in the BP. OK?

TobiasWinchen:
Good catch - I support this idea (and will add `check order of parameters' to my list of points for a code review)

adundovi:
Would you mind sharing the list for a code review (to e-mail, so that we don't pollute the topic) :-)

reichherzerp:
If I remember correctly, I changed the order of arguments for the original version when I had only one constructor for both variants (fixed and adaptive step size) so that either 2 (fixed), 3/4 (adaptive) arguments could be passed: 

PropagationBP (ref_ptr< MagneticField > field, double minStep, double maxStep=(1 *Gpc), double tolerance=1e-4)

This had the simple reason that no tolerance had to be specified for the fixed step size version.

One of the reasons why I chose two constructors (as it is implemented now) was that it could be called like PropgationCK. But I actually forgot to undo the order of the arguments...

Therefore many thanks for the accurate review!

Issue #225, opened 2019-02-25
lack of secondary protons or neutrons in the PhotoPionProduction?


Froehliche-Kernschmelze:
Dear all,

from the current implementation of the treatment of secondary particles in the `PhotoPionProduction` it seems that there would no secondary non-antinucleons be added shall they be created in SOPHIA. My concerns evolve around

https://github.com/CRPropa/CRPropa3/blob/bd39f844217a02c7b5f4d3f6a67d88d22617650a/src/module/PhotoPionProduction.cpp#L235-L269

**How this part of the code (currently) works:**

* `double nParticles`, `particleList[2000]`, and `double momentaList[5][2000]` are return values of SOPHIA. The first is the number of out-going particles from the SOPHIA event generator **including** the primary proton or neutron. The second contains the SOPHIA-intern particle IDs of these particles. The third contains each their x,y,z momentum, energy and rest mass. These arrays have the size of 2000 to account for very large secondary outputs as variable array sizes may not have been implementable when the SOPHIA/C++ junction was written.
* Each time SOPHIA was called, there is a loop of length `nParticles` over these lists checking in a switch for the corresponding particle ID. 13 and 14 correspond to protons and neutrons respectively whereas -13 and -14 denote their anti-particles
* in each branch of the switch that is not among (13,14,-13,-14), i.e. the secondary leptons and photons, there will the corresponding particle be added as a secondary unless the user opts these out.
* For hadrons this treatment discriminates between whether the incoming hadronic particle was a nucleus with mass number >1 or a single nucleon.

**what I think where the bug is (if any)**
For simplicity assume that there is an incoming very-high-energy proton undergoing photo-hadronic interactions in the PPP. The CM energy is sufficiently high to produce a single proton/anti-proton pair. The SOPHIA output therefore looks something like:
* `nParticles = 1337`
* `particleList = [13, 13, -13, (lots of non-hadrons)]` where the first 13 is the primary and the second belongs to the hadron pair.
* `momentaList = [ [px0, py0, pz0, E0, m], [px1, py1, pz1, E1, m], [px2, py2, pz2, E2, m], [all the rest] ]`
[...]

Issue #224, opened 2019-02-19
git history


lukasmerten:
Hi all, 
it seems that @TobiasWinchen's fix of amount of contributed code introduces some issues in the git history.

When I try to merge the current CRPropa/master branch into my local master branch I to get a bunch of merge conflicts. Since this fix my master and the CRPropa3/master are diverged by ~1300 commits.

This is probably also related to the huge number of changed files in Mario's pull request  #220.

At least @reichherzerp and @antoniusf can confirm this problem.

I think we should try to resolve this before further more people merging the current CRPropa3/master which might introduce new merge conflicts for future pull-requests.

Best, Lukas

antoniusf:
Hi all,

I've just taken a closer look at this, and it appears to me that the fix has rewritten a large part of the entire commit history (compare the 1300 commits divergence in most forks vs about 1500 commits total). The last non-rewritten commit appears to be at 6f4b17e3f123d636036b43fa056841dd178495aa.

To remove Mike's unnecessary contribution, I do feel like it should be sufficient to to rewrite only @reichherzerp's branch pre-merge, along with the merge commit onto master (e5de0d444542f72f2be3edbb8d2ab3d1101cc0d9), and the three master commits that followed it. While this might still cause problems, it would only affect those who have already pulled in that merge commit.

Of course, this would require to first undo the current fix, which I personally do think is advisable, but it is probably good to discuss this first. I should also say that I find `git filter-branch` to be quite scary; it was probably a lot easier for me to propose the above solution than it would be for someone to implement it.

 -- Antonius

TobiasWinchen:
I guess that rebasing the branches onto master should do the trick

lukasmerten:
I just created a new branch from my current master to test the rebase.
`git checkout -b TestMaster`
Then I tried to rebase on current CRPropa3/master
`git rebase upstream/master`, 
where upstream is pointing to https://github.com/CRPropa/CRPropa3

It did not work and creates a lot of merge conflicts, too.

TobiasWinchen:
@lukasmerten - I just tried to reproduce the problem and cloned your master from github and added CRPropa3/master as upstream:
 `$git remote add upstream https://github.com/CRPropa/CRPropa3 && git fetch upstream`
I then tried to merge 
`$git merge upstream/master`
yielding only few conflicts:

```$git status                                                 
On branch master
Your branch is up to date with 'origin/master'.

You have unmerged paths.
  (fix conflicts and run "git commit")
[...]

PR #226, opened 2019-03-19
bugfix PhotoPionProduction resolving issue #225

Froehliche-Kernschmelze:
This pull request aims to resolve issue #225.

Changes occur in the `PhotoPionProduction.cpp` file in the treatment of secondary particles after SOPHIA has been called. In order to reduce the amount of extra run time needed, the code makes the following approach:

1) The outgoing particles from SOPHIA are being looped over as before, yet each proton and neutron (particle IDs 13 and 14) are not being added as secondaries directly. Their type (proton or neutron) and energy are being stored in two separate vectors `pnType` and `pnEnergy` and no secondary is being generated in this case, nor is the canditate's energy being modified. Particles that are not protons or neutrons are being treated as they were before.
2) The maximum energy of all outgoing protons and neutrons is searched in `pnEnergy` as this is the agreed criterion of being the primary particle after the process of photo pion production in #225 by @adundovi and me. This discrimination may obscure cases in which the primary achieves lower energies than its secondaries, yet the total energy and hadron numbers will always be preserved.
3) A loop over these remaining protons and neutrons either modifies the candidate's energy (if primary) or adds secondary nucleons (if not primary particles).

The advantage of this approach is that there is no brute-force 2nd loop over all outgoing particles but only over all nucleons emerging from the interaction. The second loop and the two extra vectors used for the type end energies of the nucleons could be avoided as well if SOPHIA would not return the 5-momentum of its secondaries but rather only a list of their energies. This modification is possible but being disregarded at this point as in a future work the particles' directional information may be invoked as well.

### DIFF
--- a/src/module/PhotoPionProduction.cpp
+++ b/src/module/PhotoPionProduction.cpp
@@ -232,38 +232,17 @@ void PhotoPionProduction::performInteraction(Candidate *candidate, bool onProton
 
 	Random &random = Random::instance();
 	Vector3d pos = random.randomInterpolatedPosition(candidate->previous.getPosition(), candidate->current.getPosition());
+	std::vector<int> pnType;  // filled with either 13 (proton) or 14 (neutron)
+	std::vector<double> pnEnergy;  // corresponding energies of proton or neutron
 	for (int i = 0; i < nParticles; i++) { // loop over out-going particles
 		double Eout = momentaList[3][i] * GeV; // only the energy is used; could be changed for more detail
 		int pType = particleList[i];
 		switch (pType) {
 		case 13: // proton
 		case 14: // neutron
-			if (A == 1) {
-				// single interacting nucleon
-				candidate->current.setEnergy(Eout);
-				try
-				{
-					candidate->current.setId(sign * nucleusId(1, 14 - pType));
-				}
-				catch (std::runtime_error &e)
-				{
-					KISS_LOG_ERROR<< "Something went wrong in the PhotoPionProduction\n" << "Please report this error on https://github.com/CRPropa/CRPropa3/issues including your simulation setup and the following random seed:\n" << Random::instance().getSeed_base64();
-					throw;
-				}
-			} else {
-				// interacting nucleon is part of nucleus: it is emitted from the nucleus
-				candidate->current.setEnergy(E - EpA);
-				try
-				{
-					candidate->current.setId(sign * nucleusId(A - 1, Z - int(onProton)));
-					candidate->addSecondary(sign * nucleusId(1, 14 - pType), Eout, pos);
-				}
-				catch (std::runtime_error &e)
-				{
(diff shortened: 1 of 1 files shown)


adundovi:
@Froehliche-Kernschmelze since there is more than 1300 commits in your PR make a rebase locally and request PR again. We cannot merge this PR in the main repo otherwise.

Froehliche-Kernschmelze:
Yeah that issue... did the Authors of #224 @TobiasWinchen or @antoniusf provide git commands for others to run on their systems? Because "rebase" is too abstract to be clear for anyone not familiar with the process.

In my case a simple `git rebase origin/master" causes as many conflicts as there are commits to be rebased.

lukasmerten:
@Froehliche-Kernschmelze since also your origin/master has ~1300 diverged commits with CRPropa3/master I suggest to do the following:

1.) Update your master to be exactly equal to the CRPropa3/master. This could be done e.g. with `git reset --hard upstream/master`, when upstream is pointing to the CRPropa3 repo. Make sure to save local changes because they will be overwritten (see e.g. [here](https://stackoverflow.com/questions/1125968/how-do-i-force-git-pull-to-overwrite-local-files)).

2.) Checkout a new branch `git checkout -b <Fix225>`

3.) Add the all relevant files necessary for your pull request and create a commit

4.) Force push to the remote branch from which you created this pull request. (Make sure have a local copy before fore pushing. `git push -f <remote> <branch>`

I a ssume that a similar process has to be done for your other pull request and I am sorry for the extra work.

lukasmerten:
On the physics side, I do agree with @Froehliche-Kernschmelze proposed way to add the secondaries correctly.

Froehliche-Kernschmelze:
Thanks a lot @lukasmerten ! As far as I overlook the situation, this pull request is ready for review now.

adundovi:
I've reviewed the changes. No additional comments. Great work! I'll merge it.

lukasmerten:
@Froehliche-Kernschmelze Do you alreday have an idea how "relevant" this bug was? Can/Should we rerun some of the simulation of the CRPropa3 paper?

Froehliche-Kernschmelze:
> @Froehliche-Kernschmelze Do you alreday have an idea how "relevant" this bug was? Can/Should we rerun some of the simulation of the CRPropa3 paper?

This bug was **only relevant to setups where the PhotoPionProduction produces anti-nucleons** (for a detailed description of it's effects and scope see https://github.com/CRPropa/CRPropa3/issues/225#issue-414083923). If anti-nucleons were being produced, the bug usually led to faulty energy assignments to candidates in the order of several orders of magnitude and always produced only one instead of multiple secondary protons or neutrons. Hence both the energy and number of secondary protons / neutrons were affected.
Reading the [CRPropa3 paper](https://arxiv.org/pdf/1603.07142.pdf) there is a setup of 1 to 1000 EeV primary hadrons (sect. 4.1) and at these energies it is very likely that anti-nucleons had been produced in the PhotoPionProduction - hence potentially involving the bug. Yet it is out of my judgment how results have been affected, yet to entirely be on the safe side I recommend re-evaluating all simulations that could involve this issue.

Issue #246, opened 2019-09-30
Frequent confusion about the "discovery" of virtual dipoles


MRCWirtz:
As a lot of people are anyways thinking about CRPropa right now at the workshop, I have here some issue:

In recent times there was a lot of confusion in the community about the formation of anisotropies after forward tracking a seemingly isotropic distribution from outside the Galaxy to the observer (I know, for most of you this means more annoying discussions about Liouville's theorem ;-) ) in CRPropa3. The mistake was probably (at least in one case I know it) the following setup of the source:

```
source = crp.Source()
source.add(crp.SourceUniformShell(crp.Vector3d(0, 0, 0) * crp.kpc, 20 * crp.kpc))
source.add(crp.SourceIsotropicEmission())
```

The user assumed in this context that the "SourceIsotropicEmission()" set directions in a way as they would arrive (on the surface of the 20kpc Galactic sphere) after isotropic propagation in the Extragalactic space. Please note, that this is not the case as the arrival directions on each surface element of the Galactic sphere should be Lamberts distributed (https://en.wikipedia.org/wiki/Lambert%27s_cosine_law)... you know, flat angles have lower "hit probabilities". 

As a lot of research time was already dedicated to this issue, I was wondering if we should implement a new source model with an intuitive name like "SourceGalacticSphere" or "SourceGalacticArrival", which set up positions and directions as expected from isotropic extragalactic propagation. If there is already something like this implemented please let me know, otherwise I will work on that soon.

If you want to discover your own virtual dipole you can execute the following script:

```
import numpy as np
import matplotlib.pyplot as plt
import crpropa as crp

n = 100000000

output = crp.ParticleCollector()

# Simulation setup
sim = crp.ModuleList()
sim.add(crp.SimplePropagation())

# collect cosmic rays at Earth
obs = crp.Observer()
pos_earth = crp.Vector3d(-15, 0, 0) * crp.kpc
obs.add(crp.ObserverSurface(crp.Sphere(pos_earth, 0.1 * crp.kpc)))
obs.onDetection(output)
sim.add(obs)

# collect outwards going cosmic rays
obs_trash = crp.Observer()
obs_trash.add(crp.ObserverSurface(crp.Sphere(crp.Vector3d(0), 21 * crp.kpc)))
sim.add(obs_trash)

# source setup
source = crp.Source()
source.add(crp.SourceUniformShell(crp.Vector3d(0, 0, 0) * crp.kpc, 20 * crp.kpc))
[...]

PR #247, opened 2019-10-02
Implement Lamberts distributed vectors and an isotropic arrival scenario at the edge of the Galaxy

MRCWirtz:
Addresses issue #246, the user is now able to setup an isotropic galactic arrival scenario:
```
source = crp.Source()
source.add(crp.SourceIsotropicGalacticArrival(crp.Vector3d(0, 0, 0) * crp.kpc, 20 * crp.kpc))
```
You can test it by run the provided script in issue #246 and replace the 

```
source = crp.Source()
source.add(crp.SourceUniformShell(crp.Vector3d(0, 0, 0) * crp.kpc, 20 * crp.kpc))
source.add(crp.SourceIsotropicEmission())
```
by the above code block. I put the observer more outside (15 kpc instead of 8.5 kpc) of the center to make the anisotropies visible also with lower statistics. You should be able to run it on your local PC within 2 minutes. 

I hope the new function will prevent more future publications about the violation of Liouville...

### DIFF
--- a/include/crpropa/Random.h
+++ b/include/crpropa/Random.h
@@ -159,6 +159,10 @@ public:
 	Vector3d randFisherVector(const Vector3d &meanDirection, double kappa);
 	/// Uniform distributed random vector inside a cone
 	Vector3d randConeVector(const Vector3d &meanDirection, double angularRadius);
+	// Random lamberts distributed vector with theta distribution: sin(t) * cos(t)
+	Vector3d randVectorLamberts();
+	// Same as above but rotated to the respective normalVector of surface element
+	Vector3d randVectorLamberts(const Vector3d &normalVector);
 	///_Position vector uniformly distributed within propagation step size bin
 	Vector3d randomInterpolatedPosition(const Vector3d &a, const Vector3d &b);
 
--- a/include/crpropa/Random.h
+++ b/include/crpropa/Random.h
@@ -159,9 +159,11 @@ public:
 	Vector3d randFisherVector(const Vector3d &meanDirection, double kappa);
 	/// Uniform distributed random vector inside a cone
 	Vector3d randConeVector(const Vector3d &meanDirection, double angularRadius);
-	// Random lamberts distributed vector with theta distribution: sin(t) * cos(t)
+	/// Random lamberts distributed vector with theta distribution: sin(t) * cos(t),
+	/// aka cosine law (https://en.wikipedia.org/wiki/Lambert%27s_cosine_law),
+	/// for a surface element with normal vector pointing in positive z-axis (0, 0, 1)
 	Vector3d randVectorLamberts();
-	// Same as above but rotated to the respective normalVector of surface element
+	/// Same as above but rotated to the respective normalVector of surface element
 	Vector3d randVectorLamberts(const Vector3d &normalVector);
 	///_Position vector uniformly distributed within propagation step size bin
 	Vector3d randomInterpolatedPosition(const Vector3d &a, const Vector3d &b);
--- a/include/crpropa/Source.h
+++ b/include/crpropa/Source.h
@@ -301,8 +301,8 @@ class SourcePulsarDistribution: public SourceFeature {
 	double Zg; // exponential cut parameter in z direction
 	double frMax; // helper for efficient sampling
 	double fzMax; // helper for efficient sampling
-	double R_max;// maximum radial distance - default 22 kpc 
-	double Z_max; // maximum distance from galactic plane - default 5 kpc 
+	double R_max; // maximum radial distance - default 22 kpc 
+	double Z_max; // maximum distance from galactic plane - default 5 kpc
 	double r_blur; // relative smearing factor for the radius
 	double theta_blur; // smearing factor for the angle. Unit = [1/length]
 
@@ -406,7 +406,7 @@ class SourceLambertDistributionOnSphere: public SourceFeature {
 	double radius;
 	bool inward;
 public:
-	SourceLambertDistributionOnSphere(Vector3d center, double radius, bool inward);
+	SourceLambertDistributionOnSphere(const Vector3d &center, double radius, bool inward);
  void prepareParticle(ParticleState &particle) const;
  void setDescription();
 };
--- a/include/crpropa/Source.h
+++ b/include/crpropa/Source.h
@@ -393,6 +393,32 @@ public:
 };
 
 /**
+ @class SourceLambertsEmission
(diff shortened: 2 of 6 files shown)


TobiasWinchen:
Wouldn't be a feature SourceLambertDistributionOnSphere(center, radius) 
```
source = crp.Source()
source.add(crp.SourceUniformShell(crp.Vector3d(0, 0, 0) * crp.kpc, 20 * crp.kpc))
source.add(crp.SourceLambertOnSphere(crp.Vector3d(0, 0, 0) * crp.kpc, 20 * crp.kpc) )
```
be more consistent to avoid doing things twice? There probably should be a comment though, to note people that isotropic on a shell is probably not what they want.

MRCWirtz:
@TobiasWinchen I agree, that SourceLambertDistributionOnSphere() or SourceLambertOnSphere() are more general and consistent naming. Two things that are a bit confusing in this context:
   1) The directions as implemented are just for inwards emission, no general emission in both directions. I could change that, however, one would lose in both cases half of the statistics. Alternatively could be SourceLambertOnSphereInwards() and SourceLambertOnSphereOutwards().
   2) To keep the consistency of first setting the sources and then emitting from then, I need in the emission step the information of the source position relative to the center of the sphere (see code lines 746-748 in Source.cpp), the normalVector. Is their a way to access it in the later step again? Or are you in favor of just overwrite the old source positions? Then one could also only just write:

```
source = crp.Source()
source.add(crp.SourceLambertOnSphere(crp.Vector3d(0, 0, 0) * crp.kpc, 20 * crp.kpc) )
```
What do you think?

Also how would you propose the comment? Just in the source code or are you referring to the tutorial?

TobiasWinchen:
> @TobiasWinchen I agree, that SourceLambertDistributionOnSphere() or SourceLambertOnSphere() are more general and consistent naming. Two things that are a bit confusing in this context:
> 
>     1. The directions as implemented are just for inwards emission, no general emission in both directions. I could change that, however, one would lose in both cases half of the statistics. Alternatively could be SourceLambertOnSphereInwards() and SourceLambertOnSphereOutwards().

Instead of two modules you could use a flag inward=True in the constructor.

>     2. To keep the consistency of first setting the sources and then emitting from then, I need in the emission step the information of the source position relative to the center of the sphere (see code lines 746-748 in Source.cpp), the normalVector. Is their a way to access it in the later step again? Or are you in favor of just overwrite the old source positions? Then one could also only just write:
> 
> 
> ```
> source = crp.Source()
> source.add(crp.SourceLambertOnSphere(crp.Vector3d(0, 0, 0) * crp.kpc, 20 * crp.kpc) )
> ```
> 
> What do you think?

OK, maybe it would be least confusing for now to not separate the position and direction distribution in individual features. Ideally one would use the Geometry system to write write something like
```
surf = MySurface()  # s = Sphere()
source.add(SourceUniformSurface(s))
source.add(SourceLambertOnSurface(s))
```
but this would mean that we need to sample a point on an arbitrary surface which is not (yet?) possible -

> Also how would you propose the comment? Just in the source code or are you referring to the tutorial?

It should go to the doxygen part and if there is a turorial/example where this is used also there.

MRCWirtz:
@TobiasWinchen Thank you for the review, I addressed all the proposals that you made. In the CRPropa3-notebooks I will soon add a script "Galactic lensing - confirm liouvelle" or similar. Please confirm if you are fine with the changes.

include/crpropa/Random.h
```diff
@@ -159,6 +159,10 @@ class Random {
 	Vector3d randFisherVector(const Vector3d &meanDirection, double kappa);
 	/// Uniform distributed random vector inside a cone
 	Vector3d randConeVector(const Vector3d &meanDirection, double angularRadius);
+	// Random lamberts distributed vector with theta distribution: sin(t) * cos(t)
```
TobiasWinchen: Comment should be /// (3x/) to make it doxygen readable

include/crpropa/Random.h
```diff
@@ -159,6 +159,10 @@ class Random {
 	Vector3d randFisherVector(const Vector3d &meanDirection, double kappa);
 	/// Uniform distributed random vector inside a cone
 	Vector3d randConeVector(const Vector3d &meanDirection, double angularRadius);
+	// Random lamberts distributed vector with theta distribution: sin(t) * cos(t)
+	Vector3d randVectorLamberts();
+	// Same as above but rotated to the respective normalVector of surface element
```
TobiasWinchen: ///

src/Random.cpp
```diff
@@ -161,6 +161,27 @@ Vector3d Random::randConeVector(const Vector3d &meanDirection, double angularRad
         return randVectorAroundMean(meanDirection, theta);
 }
 
+Vector3d Random::randVectorLamberts() {
+	// random vector following Lamberts cosine law (https://en.wikipedia.org/wiki/Lambert%27s_cosine_law)
```
TobiasWinchen: This reference + explanation should go to the comment of the interface

include/crpropa/Source.h
```diff
@@ -393,6 +393,26 @@ class SourceIsotropicEmission: public SourceFeature {
...
+Galactic propagation with an isotropic arrival distribution at the Edge of our
+Galaxy. Note, that for simulation speed you should rather use the backtracking
+technique: see e.g. http://physik.rwth-aachen.de/parsec
+*/
+class SourceLambertDistributionOnSphere: public SourceFeature {
+	Vector3d center;
+	double radius;
+	bool inward;
+public:
+	SourceLambertDistributionOnSphere(Vector3d center, double radius, bool inward);
```
TobiasWinchen: cener should be const ref.

include/crpropa/Vector3.h
```diff
@@ -183,9 +183,9 @@ class Vector3 {
 		return (*this) - getParallelTo(v);
 	}
 
-	// rotate the vector around a given axis by a given a angle
+	// rotate the vector around a given axis by a given angle
 	Vector3<T> getRotated(const Vector3<T> &axis, T angle) const {
-		Vector3<T> u = axis;
+		Vector3<T> u = axis / axis.getR();
```
TobiasWinchen: I just noticed that this change apparently make the vector test fail - although I think that a normalized axis here should be reasonable

MRCWirtz: Alright, then we leave getRotated() unchanged for the moment and I change the normalization to the randVectorLamberts() function for now,

Issue #244, opened 2019-09-30
Keyword arguments are not always recognized


TobiasWinchen:
Reported by Jörg Rachen, CRPropa meeting October 2019. We should derive an example case when this happen and develop checks that all kwargs are passed correctly.

TobiasWinchen:
This is clearly a bug:
```
p = crpropa.PhotoDisintegration(havePhotons=True)

In [7]: p.getDescription()
Out[7]: 'PhotoDisintegration: CMB'

In [8]: p = crpropa.PhotoDisintegration(havePhotons=True, photonField=crpropa.IRB_Dominguez11)

In [9]: p.getDescription()
Out[9]: 'PhotoDisintegration: CMB'

In [10]: p = crpropa.PhotoDisintegration(photonField=crpropa.IRB_Dominguez11)

In [11]: p.getDescription()
Out[11]: 'PhotoDisintegration: CMB'

```

TobiasWinchen:
Apparently, swig cannot handle keyword arguments for overloaded functions. So they cannot be used here. However, sadly no error is created, even on
```
p = crp.PhotoDisintegration(nonExistingKeywordArguemntShouldRaiseException=True)
```

This needs to be fixed ASAP as this is highly dangerous.

TobiasWinchen:
I created a pull request #250 to disable swig -builtin option as a default. Without -builtin the code works as expected. As we already have a test, we can close this feature for now. The main issue will be that pure python extensions will be somewhat slower due to the additional proxy layer.

adundovi:
OK. However, as soon as the issue is solved in the upstream we should switch back to builtin.

PR #250, opened 2019-10-05
disable swig -builtin as default

TobiasWinchen:
This is a temporary workaround for issue #244: 

With this option kwargs are curerrntly silently ignored. This is
probably a SWIG bug https://github.com/swig/swig/issues/1595


adundovi:
OK. I agree that -builtin should be defaulted to "off" as long as this issue is not fixed. Although, without the "builtin" option we have longer loading time of the python library, and some other possible downsides ([python slots](http://www.swig.org/Doc4.0/Python.html#Python_builtin_overloads)?).

PR #255, opened 2019-11-18
Add a check for CMB field before deactivating the the redshift evolution

### LINKED ISSUE #254: Redshift evolution in PhotoPionProduction
**Description**

It seems that the redshift evolution is ignored in the current implementation of the PhotoPionProduction. Even for non CMB target fields the redshift dependence is turned off as for intended for the CMB. 

`void PhotoPionProduction::setPhotonField(PhotonField field) {
photonField = field;
if (haveRedshiftDependence) {
std::cout << "PhotoPionProduction ... " << std::endl;
haveRedshiftDependence == false;
}
...
}`

**To Reproduce**
from crpropa import *
EBL = IRB_Finke10
PPP = PhotoPionProduction(EBL, True, True, True, True, 0.1, True)

**Expected behavior**
No output

I have not checked any simulation results. However, the file rate_IRBz_Finke10.txt is not opened as I would have expected when including the redshift evolution.

lukasmerten:
Should resolve #254 
@Froehliche-Kernschmelze have you encountered this?

### DIFF
--- a/src/module/PhotoPionProduction.cpp
+++ b/src/module/PhotoPionProduction.cpp
@@ -27,7 +27,7 @@ PhotoPionProduction::PhotoPionProduction(PhotonField field, bool photons, bool n
 
 void PhotoPionProduction::setPhotonField(PhotonField field) {
 	photonField = field;
-	if (haveRedshiftDependence) {
+	if (field==CMB && haveRedshiftDependence) {
 		std::cout << "PhotoPionProduction: tabulated redshift dependence not needed for CMB, switching off" << std::endl;
 		haveRedshiftDependence = false;
 	}
--- a/src/module/PhotoPionProduction.cpp
+++ b/src/module/PhotoPionProduction.cpp
@@ -27,9 +27,14 @@ PhotoPionProduction::PhotoPionProduction(PhotonField field, bool photons, bool n
 
 void PhotoPionProduction::setPhotonField(PhotonField field) {
 	photonField = field;
-	if (field==CMB && haveRedshiftDependence) {
-		std::cout << "PhotoPionProduction: tabulated redshift dependence not needed for CMB, switching off" << std::endl;
-		haveRedshiftDependence = false;
+	if (haveRedshiftDependence) {
+		if (field==CMB){
+			std::cout << "PhotoPionProduction: tabulated redshift dependence not needed for CMB, switching off" << std::endl;
+			haveRedshiftDependence = false;
+		}
+		else {
+			KISS_LOG_WARNING << "PhotoPionProduction: You are using the 2-dimensional tabulated redshift evolution, which is not available for other interactions. To be consistent across all interactions you may deactivate this <setHaveRedshiftDependence(False)>.";
+		}
 	}
 	std::string fname = photonFieldName(field);
 	setDescription("PhotoPionProduction: " + fname);


Froehliche-Kernschmelze:
Yes, I have encountered this a while ago and this is clearly a bug and the fix is indeed as easy as suggested. The effect of this bug is that all redshift-dependence is switched off automatically regardless of the selected photon field. For the CMB this was obviously without consequences, yet for the IRB_Kneiske04 only the spectral shape at z=0 is used (see tabulated files for a comparison). Any other photon field that is not the CMB and IRB_Kneiske04, including its redshift dependence, is not taken into account anyways by the PhotoPionProduction as pointed out in #220

lukasmerten:
When I understand you correctly, @Froehliche-Kernschmelze, this will be fixed with #220. However, it won't hurt your pull request if we fix it -- at least for the Kneiske_04 field -- now?

Has anyone an example using redshift evolution in PPP, which can test with and without the fix. So we can estimate the influence of this. If not I'll try to find some time in the next days to look into this.

avvliet:
@lukasmerten I would indeed suggest first testing what happens with the fix before merging this pull request. The fix looks perfectly fine of course, but as the redshift evolution in PPP has basically been disabled before this fix it should be tested properly before this merge enables it.

rafaelab:
Hi all,
I think the commit is fine and this is a plain "bug" that hasn't been really affecting anything.
However, I agree with Arjen that it is more prudent to make sure nothing breaks (although this is unlikely).

lukasmerten:
Hi all

First of all I want to emphasize that this bug **does not mean that no scaling** of the IRB was implemented. The bug was just a silent fall back to the simple scaling also the user specified the 2-dimensional sampling.

As Arjen said I have made some tests, comparing the mean free path in the IRB field. Exemplary, the Kneiske 2004 field is attached. It is clearly visible that differences in the mean free path exist, comparing the two-dimensional interpolation (haveRedshiftDependence=True) and the simpler scaling of the field (haveRedshiftDependence=False) [IRB_Kneiske04.pdf](https://github.com/CRPropa/CRPropa3/files/3926911/IRB_Kneiske04.pdf).

In addition, I have also checked the influence of the different redshift scaling on actual simualtion output. In doing so, I have slightly adjusted the [example for neutrino production](https://github.com/CRPropa/CRPropa3-notebooks/blob/master/secondaries/neutrinos.v4.ipynb). I decreased the energy threshold to 1PeV and increased the statistics to 120,000 primaries. Furthermore, I deactivated all modules but the PhotoPionProduction (IRB_x). The difference between the two implementations becomes largest at the lowest energies 
[Kneiske04_neutrinoDif](https://github.com/CRPropa/CRPropa3/files/3926974/IRB_Kneiske04-neutrinos_Dif.pdf) . Also this may seem quite large, the differences between different EBL models are at least on the same order of magnitude. For other models (e.g. Finke10) the disagreement between the two redshift treatments can be much smaller (~percent level).

Notes:
* The differences between two-dimensional interpolation and scaling of the redshift have been one of the suggested reasons to explain differences between SimProp and CRPropa (see [arXiv:1901.01244](https://arxiv.org/abs/1901.01244)). This could be re-checked with the 2-dim sampling now available in CRPropa.

* At the moment it is not suggested to use the 2-dim sampling together with other loss modules (PhotoDisintegration, etc.). Since the 2-dim interpolation is only available for PPP an inconsistent treatment of the EBL would otherwise be introduced. The latest commit introduces a KISS_LOG_WARNING to inform the user.

* The 2-dim sampling should become the default treatment for all loss processes in the future.

Best, 
Lukas

PS: If want to see plots also for other IRB implementation please write me an email.

lukasmerten:
Thanks for giving feedback on this issue - internally and here. I am merging.

PR #256, opened 2019-11-19
Break condition: charge number

lukasmerten:
**What's new?**

Adds a new break condition based on the charge number of the candidates.

**Use case**
I was using this feature for a study on the propagation distance of different elements. I did not want to use the minimum energy break condition. With this break condition I was able to get rid of low charge number candidates that I was not interested in.

### DIFF
--- a/include/crpropa/module/BreakCondition.h
+++ b/include/crpropa/module/BreakCondition.h
@@ -83,6 +83,25 @@ public:
 };
 
 /**
+ @class MinimumChargeNumber
+ @brief Deactivates the candidate below a minimum number
+
+ This modules deactivates the candidate below a given minimum charge number.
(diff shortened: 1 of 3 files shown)


rafaelab:
Is this an essential addition to the CRPropa core?
What applications do you have in mind for this?
I think that if this is a useful feature for a very particular application, then it might be better to keep the code minimalist.
What do you think?

lukasmerten:
The application I used this for was the simulation of the energy loss length combining different loss modules. I found it was quiet handy as it allows to neclect a minimum energy cut and still reduces the simulation time significantly compared to runs without such a minimum charge number cut. 

It also does not really add much code and is in line with the other break conditions that are already available.

lukasmerten:
@rafaelab Tanks for approving. I'll keep it open until tomorrow for further comments and merge then.

PR #259, opened 2019-12-16
Implementation of Terral and Ferrière 2017 Galactic magnetic field

tfitoussi:
Dear all,

In this pull-request, I want to provide the implementation of the Galactic and halo magnetic field presented in [Terral & Ferrière 2017](https://ui.adsabs.harvard.edu/abs/2017A%26A...600A..29T/abstract).

I implemented all 6 models of the paper: disk (Ad1, Bd1, Dd1) + halo (C0, C1).

The models can be set as follow :
```python3
disk_model = "Ad1"
halo_model = "C0"
GMF = crpropa.TF17Field(disk_model,halo_model)
```
The field at a given position can be call as usual:
```python3
B = GMF.getField(crpropa.Vector3d(0,0,0))
```

All models have been checked by Katia Ferrière. As far as I know, the implementation is correct.

I would like to point out that from the few tests I made, the models have some problems. The Faraday rotation is indeed only fittted in the halo (excluding the Galactic plan). As a consequence, the Galactic models (Ad1, Bd1, Dd1) are poorly fitted. In particular really strong amplitudes of the magnetic field (higher than few hundreds microgauss) can appear locally.

Nevertheless, since it is implemented. I think it is worthy to include it in CRPropa.

Thank you for considering the pull-request,
Cheers,

Thomas F.

### DIFF
--- a/include/crpropa/magneticField/TF17Field.h
+++ b/include/crpropa/magneticField/TF17Field.h
@@ -14,14 +14,14 @@ using namespace std;
 
  */
 
+enum class TF17DiskModel {Ad1, Bd1, Dd1};
+enum class TF17HaloModel {C0, C1};
+
+
 class TF17Field: public MagneticField {
 private:
-    // model definition
-    bool C0;
-    bool C1;
-    bool Ad1;
-    bool Bd1;
-    bool Dd1;
+    TF17DiskModel disk_model; 
+    TF17HaloModel halo_model;
 
 	// disk parameters
 	bool useDiskField;
@@ -53,7 +53,7 @@ private:
 	void SetParams();
 
 public:
-	TF17Field(string disk_model="Ad1", string halo_model="C1");
+	TF17Field(TF17DiskModel disk_model_=TF17DiskModel::Ad1, TF17HaloModel halo_model_=TF17HaloModel::C1);
 
 	void setUseDiskField(bool use);
 	void setUseHaloField(bool use);
@@ -77,6 +77,9 @@ public:
     void set_Hp(const double H);
     void set_Lp(const double L);
     void set_p0(const double p0);
+
+    string getDiskModel() const;
+    string getHaloModel() const;
 
 	Vector3d getField(const Vector3d& pos) const;
 	Vector3d getDiskField(const double& r, const double& z, const double& phi, const double& sinPhi, const double& cosPhi) const;
--- a/include/crpropa/magneticField/TF17Field.h
+++ b/include/crpropa/magneticField/TF17Field.h
@@ -7,6 +7,9 @@ namespace crpropa {
 using namespace std;
 
 /**
+ @class TF17Field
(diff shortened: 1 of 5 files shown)


rafaelab:
Hi @tfitoussi
Do you have some material (slides, notes) showing the tests you performed?
I think it would also be interesting to have you present your implementation and your tests in our next meeting in January so that we can go over the details.

tfitoussi:
Hi @rafaelab 
Right now, I have essentially few figures that I used to check the implementation with Katia Ferrière. 
For a presentation, I need to check with my bosses when I come back from Holidays in January (if it's not too late).

rafaelab:
The presentation is not necessary, but some commented figures would be extremely helpful. 
In particular, if you could show us the tests you have done to make sure you implementation is correct, that would be great!

tfitoussi:
Here is [the document](https://github.com/CRPropa/CRPropa3/files/3969599/TF17.pdf) I used with Katia to check the implementation.
We check the structure of the lines of field (that are comparable to the one published) and the limited development needed when r or z tend to 0.

adundovi:
I see that a lot of work went into this one. The attached document will suffice to verify your implementation. Nice job @tfitoussi !

I could comment only that you should probably not use strings to choose between the sub-models, but an enumerator to enforce strict typing, and to avoid the last else (catch-all) case. I can explain in more details what I mean by that, if needed. Also, it would be helpful if you provide some automatic tests for the model (for example, values of the field at locations in the galaxy that can be easily read out from the original paper). Tests are not the requirement for the inclusion, because we don't have them for other models of GMF, but we should change this practice...

include/crpropa/magneticField/TF17Field.h
```diff
@@ -0,0 +1,93 @@
+#ifndef CRPROPA_TF17FIELD_H
+#define CRPROPA_TF17FIELD_H
+
+#include "crpropa/magneticField/MagneticField.h"
+
+namespace crpropa {
+using namespace std;
+
+/**
+ Implements the TF2017 galactic magnetic field model, consisting of large-scale regular disk and halo fields.
```
lukasmerten: @tfitoussi: Could you add some doxygen descriptors like, e.g. [here](https://github.com/CRPropa/CRPropa3/blob/master/include/crpropa/magneticField/JF12FieldSolenoidal.h)?

tfitoussi:
Hello everyone (and happy new year)

As suggested, I replaced the sub-models selection using strings by enumerators. Thanks @adundovi !
I also added Doxygen descriptors hoping that it will be sufficient for users to understand. 
And I merged the last commits from the main branch of CRPropa.

Can we proceed with this pull-request or should I start a new one cleaner ?

Here is the [last version of the document](https://github.com/CRPropa/CRPropa3/files/4047156/TF17.pdf) I already sent you, using the last implementation of the code.

Cheers,

PR #260, opened 2019-12-19
Separate photon sampling from SOPHIA

Froehliche-Kernschmelze:
Dear all,

this PR extracts the internal SOPHIA photon sampling methods and re-implements them in CRPropa.

This is a necessary preparatory step towards the implementation of variable photon fields as discussed in #220 (sometimes referred to as "step zero" in the agreed road map). The changes are:

- SOPHIA's sampling methods have been implemented in C++
- PhotoPionProduction now calls the C++ versions of SOPHIA's photon sampling methods
- This allowed for SOPHIA's signature to be changed to more meaningful parameters
- A direct SOPHIA interface has been added which could be used outside of a simulation

This PR comes with scripts which
- run minimal simulations of the PhotoPionProduction on a wider range of parameters
- analyse the similarity of the introduced changes vs. the master branch

To make a review easier:
- https://github.com/CRPropa/CRPropa3/commit/5c650ec386f1a82888fc9aad27800ef869f92ed4 provides the same functionalities as all later commits
- https://github.com/CRPropa/CRPropa3/commit/77d0f60240f6e280252aa00d6286976e70de2e8e deletes other parts of SOPHIA which are not used under any circumstances in CRPropa (Note that no non-linear behaviour of the code was touched with this)
- The tests above compare simulation data (i.e. at the highest hierarchy level). However, there are also tests available for each single new method which is introduced with this PR if this would become important at some point.

Important remarks:
- Please consider that the CRPropa naming style has not been met at each point to make comparisons to the SOPHIA methods feasible. Also the unit convention is broken internally in these methods. I would most eagerly recommend to amend the naming issues once the changes themselves have been reviewed.
- The scripts for testing are found in [testScripts.zip](https://github.com/CRPropa/CRPropa3/files/3983762/testScripts.zip). Note that the simulation script needs to be changed for each branch in which you are running the script (one for master, one for these changes, description provided).

Please do get in touch at any given point!


include/crpropa/module/PhotoPionProduction.h
```diff
@@ -60,6 +61,17 @@ class PhotoPionProduction: public Module {
 	 @param z		redshift
 	 */
 	double lossLength(int id, double gamma, double z = 0);
+
+	/**
+	 Direct SOPHIA interface.
+	 Output is a vector of length 2*N where N is the number of out-going particles.
```
TobiasWinchen: Is there a reason not to define a simple 
struct SophiaParticle{
int id;
double energy;
};
Storing the particle data in the double array is really ugly.

src/PhotonBackground.cpp
```diff
@@ -11,46 +13,46 @@ namespace crpropa {
...
-			infile.ignore(std::numeric_limits<std::streamsize>::max(), '\n');
-		}
-
-		infile.close();
-		initialized = true;
-	}
-
-	double scalingFactor(double z) {
-		if (!initialized) 
+    bool initialized;
```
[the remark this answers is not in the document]
Froehliche-Kernschmelze: weirdly Git sees a difference here but I haven't touched anything in PhotonFieldScaling and never intended to. Have changed initialized to isInitialized anyways.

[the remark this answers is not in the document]
Froehliche-Kernschmelze: ah my bad, a field command changed tabs to spaces. keeping the rename of isInitialized, though.

Froehliche-Kernschmelze:
Thanks lot @TobiasWinchen for the fast reply! I have implemented your feedback on all your points up to my best understanding.

Regarding the changes in SOPHIA: Sorry, I should have elaborated more on what actually changed.
- @adundovi and me concluded that, since we are modifying SOPHIA anyways, we should aim for the maximal amount of code reduction we are able to achieve given the change of the CRPropa/SOPHIA interface.
- the steps of modifying SOPHIA therefore evolve as follows:
1) Minimal solution: out-comment calls to internal sampling routines
2) Middle solution: delete all photon sampling routines from SOPHIA (as they are not used)
3) Maximal solution: delete all unused code from SOPHIA (checked for GOTO violations, comprehensive tests provided) and improve readability by proper indentations and styles

What you were seeing here was the maximal solution and I totally see how this is confusing at first glance. For the sake of a more efficient review, I will change sophia_interface.f to the minimal solution to ease this step of iteration while hoping to get to the maximal solution later on.

include/crpropa/module/PhotoPionProduction.h
```diff
@@ -12,13 +12,20 @@ namespace crpropa {
  * @{
  */
 
+struct SophiaEventOutput {
```
TobiasWinchen: OK, I meant a vector of structs. But for a temporary solution this will be ok.

include/crpropa/module/PhotoPionProduction.h
```diff
@@ -60,6 +67,17 @@ class PhotoPionProduction: public Module {
 	 @param z		redshift
 	 */
 	double lossLength(int id, double gamma, double z = 0);
+
+	/**
+	 Direct SOPHIA interface.
+	 Output is a vector of length 2*N where N is the number of out-going particles.
```
TobiasWinchen: Comment is outdated

src/PhotonBackground.cpp
```diff
@@ -123,4 +125,293 @@ std::string photonFieldName(PhotonField photonField) {
...
+	static const double X[8] = {.0950125098, .2816035507, .4580167776, .6178762444,
+						 .7554044083, .8656312023, .9445750230, .9894009349};
+	static const double W[8] = {.1894506104, .1826034150, .1691565193, .1495959888,
+						 .1246289712, .0951585116, .0622535239, .0271524594};
+	const double XM = 0.5 * (B + A);
+	const double XR = 0.5 * (B - A);
+	double SS = 0.;
+	for (int i = 0; i < 8; ++i) {
+		double DX = XR * X[i];
+		if (type == "functs") {
```
TobiasWinchen: Sorry that I missed this before, but the type-switch should not be  a string comparison, and certainly not in the inner loop! If you consider the gauss legendre integration temporary code, either use function pointers or simply create two methods.

If the integral will stay there, add a comment + file an issue (Feature request, gauss legendre integration) that this should go to a separate utility class, potentially from a external library. 

If only temporary, also add a comment that this is a temporary solution. Clearly marking temporaries is probably also good for the future reviews.

src/PhotonBackground.cpp
```diff
@@ -123,4 +125,293 @@ std::string photonFieldName(PhotonField photonField) {
...
+			i_rep++;
+			if (random.rand() < eps * eps * getPhotonDensity(eps, z_in) / rmax)
+				keepTrying = false;
+		} while (keepTrying);
+	}
+	return eps * eV;
+}
+
+double PhotonFieldSampling::prob_eps(double eps, bool onProton, double E_in, double z_in) const {
+	static const double mass = onProton? 0.93827 : 0.93947;  // Gev/c^2
```
TobiasWinchen: This should not be static (sorry if I marked this wrongly before)

src/PhotonBackground.cpp
```diff
@@ -123,4 +125,293 @@ std::string photonFieldName(PhotonField photonField) {
...
+	if (flag != 1 && flag != 2)
+		throw std::runtime_error("error: incorrect background flag. Must be 1 (CMB) or 2 (IRB_Kneiske04).");
+	bgFlag = flag;
+}
+
+double PhotonFieldSampling::sample_eps(bool onProton, double E_in, double z_in) const {
+	if (bgFlag == 0)
+		throw std::runtime_error("error: select photon field first: 1 (CMB) or 2 (IRB_Kneiske04)");
+
+	static const double mass = onProton? 0.93827 : 0.93947;  // Gev/c^2
```
TobiasWinchen: This should not be static!

include/crpropa/module/PhotoPionProduction.h
Froehliche-Kernschmelze: I could implement such a structure and it would compile properly and could be used in a terminal window as intended. However, I could not access the output data due to SWIG deeming the output data non-iterable.

src/PhotonBackground.cpp
adundovi: Here, and elsewhere in the code, we should use enum class since it increases readability of a code, and makes less chance for run-time errors. This is a good practice coming to C++ from programming languages with more advanced type systems.

For example:
in `PhotonBackground.h`:
```c++
enum class IntegrationTypeFunction { functs, prob_eps };
```
and the interface becomes:
```c++
gaussInt(IntegrationTypeFunction type, ...
```
so in `PhotonBackground.cpp`:
```c++
if (type == IntegrationTypeFunction::functs)
```
Finally, with this approach you don't need the second else as any other cases besides those defined in the enum class are impossible and caught during the compile time.

However, in this particular case, I would suggest splitting the code as Tobias wrote.

adundovi:
>     * @adundovi and me concluded that, since we are modifying SOPHIA anyways, we should aim for the maximal amount of code reduction we are able to achieve given the change of the CRPropa/SOPHIA interface.

@TobiasWinchen I confirm this. The idea is to clear the SOPHIA code as much as possible from unneeded parts, so that we can focus on what really matters and proceed with fixing and improving PPP. The current implementation of PPP with SOPHIA is not only slow (cannot be parallelized beyond 8+ threads) and unmaintainable (totally unreadable), but also lacking the sampling on other photon fields which some users consider as a bug (because it was not explicitly stated). Thus, Mario's work here is more like a bug fix, not a new feature implementation. Therefore, keeping the old state of the code in parallel in this case is not reasonable. After this commit, or how Mario calls it, the zeroth step, we can proceed with new contributions which can be checked systematically against the old one.

include/crpropa/module/PhotoPionProduction.h
```diff
@@ -12,13 +12,20 @@ namespace crpropa {
  * @{
  */
 
+struct SophiaEventOutput {
+	int nParticles;
+	std::vector<double> energy;
+	std::vector<int> id;
```
adundovi: Is the id variable SOPHIA-specific? If so, I suggest to go with a dedicated enum class in `sophia.h` where all particle ids are defined (like I wrote [here](https://github.com/CRPropa/CRPropa3/pull/260#discussion_r363662596)) and then this should be `std:vector<SophiaParticle>`.

```c++
enum class SophiaParticle : int {
   proton = 13,
   antiproton = -13,
   neutron = -14,
   ...
};
```
Instead of `int`s later, you use `SophiaParticle` instead, and if-comparisons are directly readable (no additional comments in the code are needed). For example, `if pType == SophiaParticle::proton`.

I don't intend to make this suggestion as a PR blocker - we can implement it later.

TobiasWinchen:
On Tuesday 7 January 2020 10:53:08 CET Andrej Dundovic wrote:
> >     * @adundovi and me concluded that, since we are modifying SOPHIA
> >     anyways, we should aim for the maximal amount of code reduction we
> >     are able to achieve given the change of the CRPropa/SOPHIA interface.
> @TobiasWinchen I confirm this. The idea is to clear the SOPHIA code as much
> as possible from unneeded parts, so that we can focus on what really
> matters and proceed with fixing and improving PPP. The current
> implementation of PPP with SOPHIA is not only slow (cannot be parallelized
> beyond 8+ threads) and unmaintainable (totally unreadable), but also
> lacking the sampling on other photon fields which some users consider as a
> bug (because it was not explicitly stated). Thus, Mario's work here is more
> like a bug fix, not a new feature implementation. Therefore, keeping the
> old state of the code in parallel in this case is not reasonable. After
> this commit, or how Mario calls it, the zeroth step, we can proceed with
> new contributions which can be checked systematically against the old one.

I agree that we should remove unused code, but changing e.g. capitalizations 
from print to PRINT at the same time just blows up the commit and makes the 
review harder.

adundovi:
> I agree that we should remove unused code, but changing e.g. capitalizations from print to PRINT at the same time just blows up the commit and makes the review harder.

Oh, sorry, I, of course, agree with that.

Froehliche-Kernschmelze:
Thanks all again for your useful comments. I am happy to resolve the integration part and consider enums for sophia particles. May I also emphasise that, based on your comments, the current version of the suggested SOPHIA code refrains from any cosmetics and only out-comments code which is not used as it is related to internal photon sampling (i.e. to be deleted later or now). The interface itself in `sophiaevent(...)` , of course, contains a very small amount of newly written code.

adundovi:
@Froehliche-Kernschmelze I have just tried to write you more code snippets for the special SophiaParticle type and realized that with the current state of the implementation where sophiaEvent() and performInteraction() methods are separated while doing the same PID conversion, it makes little sense to do what I suggested. So please, skip the suggestion for the moment. We'll do it later when we merge sophiaEvent() into performInteraction().

include/crpropa/module/PhotoPionProduction.h
TobiasWinchen: It would be clearer to change the ID values in sophia to the HEPID standard as in CRPropa - but this is an issue for another day.

src/PhotonBackground.cpp
Froehliche-Kernschmelze: I have implemented a suggestion of how the Gauss integration is moved to `.../Common.h` while accepting arbitrary functions over which to integrate. However, using explicit function pointers did not work for me as the functions over which to integrate do not have the same signature.
* at this point: is there a way to define a variable function type to pass to a function over which to integrate?
* My current attempt in a402b8eb2460793a0ed9f79266f50e7694b6b402 tries to solve this based on [this post](https://stackoverflow.com/questions/15959374/c-pass-function-with-arbitrary-number-of-parameters-as-a-parameter). However, the compiler throws an error where the Gauss integration is called mentioning "invalid use of non-static member functions". I could not fix this but maybe there is a trivial solution to this that I am not seeing? If yes then we would have resolved all concerns evolving around the Gauss integration.

adundovi: I would do it like standard integration methods in other languages are doing it (see for example, python's [scipy](https://docs.scipy.org/doc/scipy/reference/generated/scipy.integrate.quad.html)): wrap your functions within another function (like a delta function), and provide it to a simplified gaussInt() interface which accepts only: f(x), A, B:

```c++
#include <functional>
(...)
double gaussInt(std::function<double(double)> integrand, double A, double B);
```

The next step is to rewrite `prob_eps` and `functs` so that they are [referential transparent](https://en.wikipedia.org/wiki/Referential_transparency), but I think, after a brief lookup, that they already are. Then, you should be able to call the integrator with a function something like (I haven't compiled it):

```c++
gaussInt(std::bind(
     [this, onProton, E_in, z_in](double x) { return this->prob_eps(x, onProton, E_in, z_in); }),
     epsMin, epsMax)
(...)
gaussInt(std::bind([this, onProton](double x){ return this->functs(x, onProton); }), smin, smax)
```

I'll check exact pieces tomorrow.

adundovi: Sorry, calling goes without std::bind():
```c++
gaussInt(
     [this, onProton, E_in, z_in](double x) { return this->prob_eps(x, onProton, E_in, z_in); },
     epsMin, epsMax);
(...)
gaussInt(
     [this, onProton](double x){ return this->functs(x, onProton); }),
     smin, smax);

Move gaussInt() to `Common.cpp`:
```c+
double gaussInt(std::function<double(double)> integrand, double A, double B) {
        static const double X[8] = {.0950125098, .2816035507, .4580167776, .6178762444, .7554044083, .8656312023, .9445750230, .9894009349};
        static const double W[8] = {.1894506104, .1826034150, .1691565193, .1495959888, .1246289712, .0951585116, .0622535239, .0271524594};
        const double XM = 0.5 * (B + A);
        const double XR = 0.5 * (B - A);
        double SS = 0.;
        for (int i = 0; i < 8; ++i) {
                double DX = XR * X[i];
                SS += W[i] * (integrand(XM + DX) + integrand(XM - DX));
        }
        return XR * SS;
}

```
In `Common.h`:
```c++
#include <functional>
(...)
double gaussInt(std::function<double(double)> integrand, double A, double B);
```

Add some tests in `testCore.cpp`, for example:
```c++
TEST(common, gaussInt)
{
        EXPECT_NEAR(gaussInt(([](double x){ return x*x; }), 0, 10), 1000/3., 1e-4);
        EXPECT_NEAR(gaussInt(([](double x){ return sin(x)*sin(x); }), 0, M_PI), M_PI/2., 1e-4);
}
```

It compiles while the test is passing :-)

TobiasWinchen: The templates are IMO the correct way here in C++, so if this works we should go for it.But please add a test to check the integration for functions with different number parameters as this is a core feature here.

include/crpropa/Common.h
```diff
@@ -67,7 +67,21 @@ inline double pow_integer<0>(double base)
   return 1;
 }
 
-
+// - input:  function over which to integrate, integration limits A and B
+// - output: 8-points Gauß-Legendre integral
+template<typename Integrand>
+inline double gaussInt(Integrand&& integrand, double A, double B) {
+	static const double X[8] = {.0950125098, .2816035507, .4580167776, .6178762444, .7554044083, .8656312023, .9445750230, .9894009349};
```
TobiasWinchen: These integration constants should not be static inside the template but a static constant outside. Otherwise there will be a new static array every time the template is evaluated.

src/PhotonBackground.cpp
adundovi: Hi @TobiasWinchen, I'm curious, why the templates approach should be the preferred way? If due to performance reasons... OK, if called frequently, I could generally agree, although, it is still ugly.

Anyway, when calling, lambda should be preferred over std::bind (since lambda is a C++ core feature, while std::bind is a library, hence, compilers can usually optimize it better). @Froehliche-Kernschmelze, in Common.h, just replace std::function with the template version you already had (and return the function body from Common.cpp):

```c++
template <typename F>
double gaussInt(F&& integrand, double A, double B) {
```
It works for me just fine.

Froehliche-Kernschmelze: Thanks a real lot, everything compiles just fine now :)

Froehliche-Kernschmelze: @adundovi despite all code compiling fine and actually running I find each Gauss integral to return the value of zero (the bounds A and B are properly passed). Inside `gaussInt(...)` any call to `integrand(someDouble)` will return zero. May this be related to how functions are passed to this method?

Froehliche-Kernschmelze: Fixed the problem by myself (as it was a trivial implementation error by myself). Sorry for the spam. Happy to be reviewed again!

include/crpropa/Common.h
```diff
@@ -3,7 +3,7 @@
 
 #include <string>
 #include <vector>
-
+#include <iostream>
```
adundovi: I suppose that this header is not needed here any more.

include/crpropa/PhotonBackground.h
```diff
@@ -1,6 +1,9 @@
 #ifndef CRPROPA_PHOTONBACKGROUND_H
 #define CRPROPA_PHOTONBACKGROUND_H
 
+#include "crpropa/Common.h"
+
+#include <functional>
```
adundovi: Also, this one is obsolete if you dropped std::function in interfaces, right?

Froehliche-Kernschmelze:
Thanks again to you two. Before we merge, 
1) shall I summarise here what the KS tests yield on the respective observables or do you prefer running the [scripts from above](https://github.com/CRPropa/CRPropa3/files/3983762/testScripts.zip) on your systems as well first?
2) Should I remove the out-commented SOPHIA parts or leave them as "comments"?
3) If at all, would you like to have the rest of "unused SOPHIA" removed in a later step or another PR? This would reduce SOPHIA to roughly 9k lines. The same test scripts could show here that nothing changed, too

adundovi:
>     1. shall I summarise here what the KS tests yield on the respective observables or do you prefer running the [scripts from above](https://github.com/CRPropa/CRPropa3/files/3983762/testScripts.zip) on your systems as well first?

Are those the same tests you already presented in the last CRPropa call? If yes, this is OK, we don't need to go through these checks again.

>     2. Should I remove the out-commented SOPHIA parts or leave them as "comments"?

Yes, please. Just don't change "print" to "PRINT" or similar things that are complicating the review process.

>     3. If at all, would you like to have the rest of "unused SOPHIA" removed in a later step or another PR? This would reduce SOPHIA to roughly 9k lines. The same test scripts could show here that nothing changed, too

Is this the old sampling part? If yes, remove them too as we agreed. If this is something else, keep it for the next PR.

Froehliche-Kernschmelze:
> > ```
> > 1. shall I summarise here what the KS tests yield on the respective observables or do you prefer running the [scripts from above](https://github.com/CRPropa/CRPropa3/files/3983762/testScripts.zip) on your systems as well first?
> > ```
> 
> Are those the same tests you already presented in the last CRPropa call? If yes, this is OK, we don't need to go through these checks again.

The test scripts above are identical to the tests I had presented, however, on a wider selection of parameters. The proposed tests conduct tests on all combinations of:
* CMB, IRB_Kneiske04 (i.e. all photon fields)
* primary protons and neutrons (i.e. all input particle types)
* primary particle energies of log10(E / GeV) = {7, 8, 9, 10, 11, 12, 13, 14} (i.e. a selection of some energies from above the threshold up to quite high energies)

and compare the results of this PR to the master branch via KS tests.
> 
> > ```
> > 2. Should I remove the out-commented SOPHIA parts or leave them as "comments"?
> > ```
> 
> Yes, please. Just don't change "print" to "PRINT" or similar things that are complicating the review process.
> 
> > ```
> > 3. If at all, would you like to have the rest of "unused SOPHIA" removed in a later step or another PR? This would reduce SOPHIA to roughly 9k lines. The same test scripts could show here that nothing changed, too
> > ```
> 
> Is this the old sampling part? If yes, remove them too as we agreed. If this is something else, keep it for the next PR.

All right, I will delete the sampling-related parts in this case. Everything else will be subject to another PR.

adundovi:
@TobiasWinchen shell we merge it now?

Issue #116, opened 2017-02-17
Vector3 interoperability with numpy / Switch to C++11


TobiasWinchen:
To increase the interoperability with numpy, it would be great to implement the Array interface for the Vector3. This would allow to use numpy methods on Vector3 and in particular to e.g. add a Vctor3 to a numpy array. However, this requires the Vector3 to have the data in an array and not in separate variables x,y,z. The resulting major change to the interface can be avoided by switching to C++11 and have public references x,y,z, to the array data as in:
``
...
public:
  T data[3];
  T &x = data[0]; 
  T &y = data[1]; 
  T &z = data[2];
...
``

Any thoughts about moving to C++11 ?

adundovi:
I mainly use it on CentOS 7 or newer machines which should support c++11 via gcc 4.8 ( https://access.redhat.com/solutions/19458 ), therefore the migration is fine with me, but I doubt that it is so with others, like those on Scientific Linux 6 or similar.

But I would also like c++11 goodies :-)

Maybe tagging 3.1 as the last pre-c++11 version could save some confusion?

On 17 February 2017 11:52:40 CET, Tobias Winchen <notifications@github.com> wrote:
>To increase the interoperability with numpy, it would be great to
>implement the Array interface for the Vector3. This would allow to use
>numpy methods on Vector3 and in particular to e.g. add a Vctor3 to a
>numpy array. However, this requires the Vector3 to have the data in an
>array and not in separate variables x,y,z. The resulting major change
>to the interface can be avoided by switching to C++11 and have public
>references x,y,z, to the array data as in:
>``
>...
>public:
>  T data[3];
>  T &x = data[0]; 
>  T &y = data[1]; 
>  T &z = data[2];
>...
>``
>
>Any thoughts about moving to C++11 ?
>
>
>-- 
>You are receiving this because you are subscribed to this thread.
>Reply to this email directly or view it on GitHub:
>https://github.com/CRPropa/CRPropa3/issues/116

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

TobiasWinchen:
C++ 11 may be required after 3.2 release

adundovi:
[...]

PR #262, opened 2020-01-03
Vector3 array interface (Closes: #116)

TobiasWinchen:
This PR provides a fundamental change in Vector3: The data  is stored now in an array `data` instead of individual variables `x,y,z` . This enables interoperability interoperability with 3rd party geometry libraries and, in particular, is required for implementing the python array interface, also part of this PR. This allows inserting Vector3d + Vector3f into standard numpy functions.
Thanks to public references in C++11, the data can still be accessed via x,y,z references so changes in the code are small. There is no performance penalty for this (tested with gcc 8.3.0).


include/crpropa/Vector3.h
```diff
@@ -28,32 +28,63 @@ namespace crpropa {
...
 	template<typename U>
-	Vector3(const Vector3<U> &v) :
-			x(v.x), y(v.y), z(v.z) {
+	Vector3(const Vector3<U> &v) {
+		data[0] = v.x;
+		data[1] = v.y;
+		data[2] = v.z;
+	}
+
+	//Vector3(Vector3 &&v) noexcept {
```
adundovi: why the noexcept version is commented out?

TobiasWinchen: I started to implement a move constructor (rule-of-5 in C++11 instead of rule-of-3) - I thought this a reason for a segfault I had before - but this is not needed.

There is no change in the -builtin version here, as the issue is with [SWIG](https://github.com/swig/swig/issues/1595#issuecomment-541809053) we should probably create a test for the issue and a test for the switch version as soon as this is resolved. I will create an issue for this so that this is not forgotten.

PR #269, opened 2020-02-03
Clean out SOPHIA

Froehliche-Kernschmelze:
Dear everyone,

this PR aims to reduce SOPHIA by deleting parts which never get used the way the code is implemented in CRPropa. The code should be in the regime of ~9k lines (of formerly ~16k) if this were approved. Here are a few notes on what happened:

* no cosmetic changes have been made
* no GOTO logic has been broken
* deletions are put into separate commits to ease up review
* there are only deletions (please compare each commit in this PR to verify)
* Confusingly, the PR summary states 64 additions which is not true
* There are [test scripts](https://github.com/CRPropa/CRPropa3/files/4148908/testScripts.zip) which yield 100% identity of results of this PR with the current master branch
* Due to the nature of this PR, there is no change in the run time of SOPHIA

Please do get in touch I may assist in explaining any of the suggested changes as the amount of deleted lines is very high.

Cheers,
Mario


Froehliche-Kernschmelze:
Thanks a real lot for coming up with this list. My thoughts on the remaining functions are based on the fact that the only interface we have in CRPropa is via ```sophiaevent``` which determines (and very much constrains) the parameters we pass to SOPHIA. Hence, one can rather safely assume certain functions to not significantly contribute to the overall performance of the code. The following functions concern play a role in the suggestions made above:

* the functions ```lund_put``` (initialize JETSET input) , ```LUJOIN``` (tell JETSET which put particles interact), ```lund_frag``` (execute JETSET), and ```lund_get``` (read JETSET output) are the only functions accessing JETSET and are only called by ```gamma_h``` which in return is only called by ```dec_inter3``` which is only called by```eventgen``` which is only called by ```sophiaevent```. The pattern of handling JETSET in ```gamma_h``` is always:
1) init particle A via ```lund_put```
2) init particle B via ```lund_put```
3) tell JETSET that these particles are supposed to interact via ```LUJOIN```
4) execute JETSET via ```lund_frag```
5) read JETSET output via ```lund_get```

**Section 1: The role of LULIST / LULOGO / LUNAME**

* LULOGO is fairly straight-forward and prints the logo of JETSET to the console

* In ```lund_frag``` the value MSTU(12) is always set to zero
https://github.com/CRPropa/CRPropa3/blob/49356068b1df5ac73671a8b4c111a7c2c0aa93dd/libs/sophia/sophia_interface.f#L2998
where the int array MSTU of length 200 is a collection of switches to en/disable general behaviour of JETSET. MSTU(12) which can be either 0 or 1 can be interpreted in today's notation as ```bool printLogo```.

* LULIST is a function supposed to generate output of various types to the console. The output summarises the parameters of the event itself as well as (I suppose) debugging information or the JETSET logo:
https://github.com/CRPropa/CRPropa3/blob/49356068b1df5ac73671a8b4c111a7c2c0aa93dd/libs/sophia/sophia_interface.f#L10717-L10722
LULIST requires an input int MLIST which defines the type of console output the function will generate in a switch-like way:
```
if (MLIST == 0 or MSTU(12) == 1):
    print logo
    return
elif MLIST in (1, 2, 3, 11, 12, 13):
    print some other stuff
return
```
i.e. the possible values for MLIST can be (0, 1, 2, 3, 11, 12, 13) where stuff gets printed and returns silently otherwise. 

* All calls to LULIST are with hard-coded MLIST values and only values for MLIST of 0, 1, and 2 are called throughout entire SOPHIA.
a) There are calls to LULIST(0) by LU1ENT, LU2ENT, LU3ENT, LU4ENT, LUGIVE, LUUPDA, LUEEVT, LUONIA (all to be removed), and LUEXEC which only checks for for MSTU(12)==1 (i.e. printLogo == True) which is never initialised this way: https://github.com/CRPropa/CRPropa3/blob/49356068b1df5ac73671a8b4c111a7c2c0aa93dd/libs/sophia/sophia_interface.f#L4960
b) There is only one call to LULIST(1) in LUTEST which is to be removed https://github.com/CRPropa/CRPropa3/blob/49356068b1df5ac73671a8b4c111a7c2c0aa93dd/libs/sophia/sophia_interface.f#L15106
c) There is one call to LULIST(2) in LUTEST which is to be removed https://github.com/CRPropa/CRPropa3/blob/49356068b1df5ac73671a8b4c111a7c2c0aa93dd/libs/sophia/sophia_interface.f#L15104
and one in LUERRM https://github.com/CRPropa/CRPropa3/blob/49356068b1df5ac73671a8b4c111a7c2c0aa93dd/libs/sophia/sophia_interface.f#L10034-L10035
which is only called if an error occurred which is sufficiently critical to make SOPHIA stop immediately.
d) There are no other calls to LULIST.

* Effectively, only LULIST(2) is ever used potentially in LUERRM after which SOPHIA is stopped and fails with an error message and a print to the console. In the least case this allows for LULIST to be reduced by all parts which deal with MLIST != 2. In a more courageous case, one could look at what would actually happen if MLIST == 2: In this case, basically the mass, energy and Px, Py, and Pz data are given out to the console which in my eyes is pointless for any user as this happens very much within the depths of JETSET where no user is ever able to take action. Note that this case ends with the termination of the SOPHIA process anyways and this should be the only info which is given. Also note that this case has never been observed by anyone using or developing CRPropa and there are no issues relating to such an error message

* LUNAME is called 4x by LUTABU (to be removed) and 17x by LULIST. It returns string representations of particles.

Conclusion to Section 1: LULIST is only called as LULIST(2) once by LUERRM after which SOPHIA terminates immediately while printing some particle data. This has never been observed to happen and the user most likely would not need to know more than that there was an error in JETSET. I see two ways here:
1) We cut out all stuff from LULIST that deals with MLIST != 2.
2) We delete LULIST entirely and insert a console print in LUERRM just before SOPHIA terminates informing the user that there was a critical error in PPP / SOPHIA
In both cases we can safely remove LULOGO as this would have been important for MLIST=0.
3) Depending on how we proceed with LULIST, the function LUNAME could become obsolete.

**Section 2: The role of LUERRM / LUEXEC**

* LUERRM (a function not to be removed and potentially slightly to be modified) is described as
https://github.com/CRPropa/CRPropa3/blob/49356068b1df5ac73671a8b4c111a7c2c0aa93dd/libs/sophia/sophia_interface.f#L10009-L10012
and obviously is an error-message function which would also stop the entire program in case of a fatal error. It is called 93x throughout the code and calls no other function except LULIST once when
https://github.com/CRPropa/CRPropa3/blob/49356068b1df5ac73671a8b4c111a7c2c0aa93dd/libs/sophia/sophia_interface.f#L10034 the input variable ```int MERR``` can be understood as a switch for LUERRM to produce a certain, desired type of error handling behaviour. However, whenever LUERRM is called, the input variable MERR is never variably passed but always passed hard-coded in all 93 cases. My suggestion here is that we replace the call to LULIST(2) and replace it with an error message.

* LUEXEC (a function not to be removed but modified) is labelled as https://github.com/CRPropa/CRPropa3/blob/49356068b1df5ac73671a8b4c111a7c2c0aa93dd/libs/sophia/sophia_interface.f#L4947-L4960 Here, the only call to LULIST is triggered if MSTU(12) == 1 i.e. if the JETSET logo should be printed; hence the call to LULIST(0) in the very same line. We can safely remove this call to LULIST as the logo never is printed.

Conclusion to Section 2: There are minor changes we could make here. LUEXEC can safely be modified as suggested since there is no print of the logo of JETSET ever. LUERRM could be modified such that a critical error message is given out and the call to LULIST be deleted. This would make LULIST and all functions it only calls obsolete.

adundovi:
Hi @Froehliche-Kernschmelze, thanks for your detailed explanation! Although, if I may comment, maybe unnecessary verbose as I've been only interested if LULIST is ever called, and if it is called, in which cases, and you answered this, of course :-) But since you invested the effort, I followed your descriptions and check the code further, and gained better understanding of changes and the code. Yet, this digging just convinced me even more that SOPHIA/JETSET is a perfect example of spaghetti code.

Indeed, LULIST is called in all cases when MSTU(12) is set to 1, except in one. (LULOGO and LUNAME can be ignored since they are self-explanatory, and only called by LULIST as I documented earlier). The exception is, as you wrote, the following one
```FORTRAN
IF(MERR.NE.17) CALL LULIST(2) 
```
Hence, I would go with your suggestion that instead of deleting this line, we replace it with some appropriate error message. When you add this line, I'll merge your PR since everything else is fine. Great job, natürlich!

PR #275, opened 2020-03-29
Double precision grids

adundovi:
I want to introduce double-precision grids in CRPropa as an option. Corner cases do exist where single-precision is not enough (especially in simulations where coupling between magnetic fields and particles are "strong", i.e., in the diffusive regime) and that can affect physical results. Before, I did this as a special CMake option with which one can recompile the source and make all grids in double-precision. Now I think it is a wrong approach, on the top of the fact that it is ugly too (I let CMake to change source files, i.e., Grid.h, which makes the whole process of compiling more complex and harder to debug).

This PR removes the changes of the previous approach and introduces single- and double-precision grids that can be used in parallel. Moreover, no new options in CMake neither preprocessing source files by CMake are required. In later PR I'll present changes to GridTools and GridTurbulence which would operate on both types of grids.

The only doubt I have is in the naming: should we change VectorGrid (ScalarGrid) to VectorGridf (ScalarGridf) analogously to Vector3f and Vector3d, but what would, however, break users python scripts? Or should I just add VectorGridDouble next to VectorGrid (like in this PR)? This is inconsistent with the previous naming, yet, it doesn't break the interface, only extends it.

### DIFF
--- a/src/GridTurbulence.cpp
+++ b/src/GridTurbulence.cpp
@@ -20,7 +20,7 @@ double turbulentCorrelationLength(double lMin, double lMax, double alpha) {
 
 #ifdef CRPROPA_HAVE_FFTW3F
 
-std::vector<std::pair<int, float> > gridPowerSpectrum(ref_ptr<VectorGrid> grid) {
+std::vector<std::pair<int, float> > gridPowerSpectrum(ref_ptr<VectorGridf> grid) {
 	
 	double rms = rmsFieldStrength(grid);
 	size_t n = grid->getNx(); // size of array
@@ -106,7 +106,7 @@ std::vector<std::pair<int, float> > gridPowerSpectrum(ref_ptr<VectorGrid> grid) 
 /* Helper functions for synthetic turbulent field models */
 
 // Check the grid properties before the FFT procedure
-void checkGridRequirements(ref_ptr<VectorGrid> grid, double lMin, double lMax) {
+void checkGridRequirements(ref_ptr<VectorGridf> grid, double lMin, double lMax) {
 	size_t Nx = grid->getNx();
 	size_t Ny = grid->getNy();
 	size_t Nz = grid->getNz();
@@ -125,7 +125,7 @@ void checkGridRequirements(ref_ptr<VectorGrid> grid, double lMin, double lMax) {
 }
 
 // Execute inverse discrete FFT in-place for a 3D grid, from complex to real space
-void executeInverseFFTInplace(ref_ptr<VectorGrid> grid, fftwf_complex* Bkx, fftwf_complex* Bky, fftwf_complex* Bkz) {
+void executeInverseFFTInplace(ref_ptr<VectorGridf> grid, fftwf_complex* Bkx, fftwf_complex* Bky, fftwf_complex* Bkz) {
 
 	size_t n = grid->getNx(); // size of array
 	size_t n2 = (size_t) floor(n / 2) + 1; // size array in z-direction in configuration space
@@ -162,7 +162,7 @@ void executeInverseFFTInplace(ref_ptr<VectorGrid> grid, fftwf_complex* Bkx, fftw
 	}
 }
 
-void initTurbulence(ref_ptr<VectorGrid> grid, double Brms, double lMin, double lMax, double alpha, int seed) {
+void initTurbulence(ref_ptr<VectorGridf> grid, double Brms, double lMin, double lMax, double alpha, int seed) {
 
 	checkGridRequirements(grid, lMin, lMax);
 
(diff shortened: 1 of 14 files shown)


TobiasWinchen:
Sounds reasonable and the implementation is basically only the naming. here i would prefer using consistent names, i.e. VectorGridd, VectorGridf, ... . Most people will use this from python. Here we could add functions VectorGrid / ScalarGrid that mimic the existing constructors and return an VectorGrid / Scalargridf f after printing out a deprecation warning that VectorGrid will be removed in CRPropa 3.9 or so nd users should use VectorGridf / ScalarGridf instead.
I think we can assume that the few C++ users should quickly notice that there are just new typedefs for the templates they were using.

src/GridTurbulence.cpp
```diff
@@ -20,7 +20,7 @@ double turbulentCorrelationLength(double lMin, double lMax, double alpha) {
 
 #ifdef CRPROPA_HAVE_FFTW3F
```
TobiasWinchen: Probably we should have two functions then for both vector grids.

src/GridTurbulence.cpp
```diff
@@ -41,7 +41,7 @@ std::vector<std::pair<int, GridPrecision> > gridPowerSpectrum(ref_ptr<VectorGrid
 		for (size_t iy = 0; iy < n; iy++) {
 			for (size_t iz = 0; iz < n; iz++) {
 				i = ix * n * n + iy * n + iz;
```
TobiasWinchen: When modifying this code, you should reduce scope of i and declare it inside the inner loop. The compiler will optimize this - resulting code may be even better,as the variable is not required to exist at all.

src/GridTurbulence.cpp
```diff
@@ -63,8 +63,8 @@ std::vector<std::pair<int, GridPrecision> > gridPowerSpectrum(ref_ptr<VectorGrid
 	fftwf_execute(plan_z);
 	fftwf_destroy_plan(plan_z);
 
-	GridPrecision power;
-	std::map<size_t, std::pair<GridPrecision, int> > spectrum;
+	float power;
+	std::map<size_t, std::pair<float, int> > spectrum;
 	int k;
```
TobiasWinchen: Reduce also scope of k while at it.

adundovi:
Thanks Tobias for your always helpful review. I agree with the renaming for the sake of consistency. Did it now. Also, I've added a deprecation warning.

Regarding other changes in `src/GridTurbulence.cpp` (like corrections of the scope of variables) - yes, I'll do that, but as a major overhaul is needed in this file (and that is what I'm preparing in parallel), I would leave these changes for next PRs, OK?

src/GridTurbulence.cpp
adundovi: This is what bothers me: should we do code duplication in these cases or use templates? If we use templates, however, we have to move a lot of code inside of the header files, which is messy. The 3rd option is that we keep only the double version of functions (if the float one is not explicitly necessary), and rely on the C++ automatic casting of float to double, right?

TobiasWinchen: The most straight forward way would be to avoid code duplication and use templates for the functions.
To avoid clutter in the header files you could separate the template implementation into an additional subfolder (e.g. details) using forward declaration:
`// foo.h
template <typename T> foo(T f);

#include detail/foo.cpp;
`
`
// detail/foo.cpp
template <typename T> foo(T f)
{
// ...
}

 This should probably also be done for the functions in gridtools.

rafaelab: Sorry for the late reply.
I think keeping the simple ScalarGrid and VectorGrid in future versions makes everybody's life easier and the naming convention cleaner. Can't we just call ScalarGridf and VectorGridf for that?
For most applications float precisions is enough, anyways.
This is just an extra line in SWIG.

rafaelab: Just to be clear, this comment is a reply to Tobias' comment:  "deprecation warning that VectorGrid will be removed in CRPropa 3.9 or so nd users should use VectorGridf / ScalarGridf instead."

TobiasWinchen: The API should be easy to understand and there should be thus consistent names. To the user, and also future developers,  it is totally opaque why there is inconsistent naming. Sure, this one thing is not a big issue, but many small issues will make a mess at some point. Users already have to use Vector3d and Vector3f and thus think about the precision they want, so I don't think that we ask too much of the users here.

rafaelab: I agree that it this is not a lot to ask. But ScalarGridf is aesthetically very unpleasant. It seems like a typo. For Vector3f, the `f' appears after a number, which makes it look okay. If we just start appending f's and d's after objects, things will get ugly. 

I think one of the nice things about CRPropa is that most of the code looks nice and clean. 
If we are asking the users to adjust to a new notation anyways, can't we think of something nicer? 

Grid1f for ScalarGridf and Grid3f for VectorGridf? This would also make the implementation of grid for variable photon fields more natural: GridNf/GridNd (N is not a number is capitalised, so it is visually better).

adundovi: I agree with both of you in your arguments: VectorGridd really looks like a typo, but we should strive for consistency across the code. @rafaelab's suggestion meets both requirements, thus if there is no other suggestion that goes in the same direction, I would vote for Grid(dim)(d|f).

TobiasWinchen: Grid(dim)(d|f) sounds great.

adundovi:
@rafaelab @TobiasWinchen Now I did the renaming to Grid(dim)(d|f) as we agreed, and I would like to close this PR if there is no other feedback... I'll continue working on this by providing specialized functions for the double grid, but that I would leave for next PRs.

Issue #199, opened 2018-08-17
Absolute normalisation DINT


Mattver:
Hi,

I am trying to have an absolute normalisation of the photon flux as given by the DINT module.
However, I don't seem to get the correct result.

Given a local injection of Q = 10^46 erg Mpc^-3 yr^-1, an E^-2 spectrum and a redshift evolution rho(z), I need to convert the number of injected photons into the real number of photons. After interaction, this same number should rescale the output flux to the physical flux. For this, I use
c/4pi  Integral[dz rho(z) / (H(z) (1+z)^2) Q / ln(Emax/Emin) * 1/E^2]
(comoving volume * comoving density * flux from a source at D_L)
with H(z) the Hubble factor and Emin and Emax the minimum and maximum energy at the source.

However, this does not give the correct result, which I can find by comparing with the result without interactions, which is
E^2 dN/dE = c/4pi t_H Q/ln(Emax/Emin)
for no evolution and ignoring the subtleties at the edges.

But, if I replace (1+z)^2 by (1+z), the result is correct using a crpropa simulation without interactions, although I don't really get why. Am I double counting a redshift effect because I use the first formula together with what happens in crpropa?

On the other hand, if I turn off all interactions inside DINT, the total flux is still several orders of magnitude below the correct result or crpropa without interactions alone, even with this change. Ignoring the total normalisation, the relative change between 'without interactions' and 'with interactions' seems to be correct however.

Am I missing something?

Thanks!

rafaelab:
We have dropped the support for DINT and EleCa.  Deprecation warnings were added with #374 .
We encourage users to migrate to the CRPropa-native EM* modules (EMPairProduction, EMInverseComptonScattering, ...) to perform simulations of electromagnetic cascades.
Examples are available in the [documentation](https://crpropa.github.io/CRPropa3). A simple one that can be used to replace DINT/EleCa is [this](https://crpropa.github.io/CRPropa3/pages/example_notebooks/secondaries/photons.v4.html).

Issue #219, opened 2019-01-25
DINT energy conservation


Vossmax:
Hi,

I tried to reproduce the photon flux calculated by the ELMAG code with CRPropa using DINT propagation. Since ELMAG and DINT don"t share an identical EBL model, I implemented the model by Gilmore (2012) in the background.cpp file of DINT. In the following plot you see my result.
![cr-elm2](https://user-images.githubusercontent.com/47029650/51746175-82f07480-20a5-11e9-975c-7d18f554e2cf.png)
The Normalisation is not the same, so I just fitted the lines manually. However, the shapes of the fluxes differ significantly. Both are simulated without magnetic field.

Could I have made a mistake in the implementation or is there a general energy conservation issue in DINT?

Thanks!

rafaelab:
We dropped the support for DINT and EleCa in previous releases. Deprecation warnings were added with #374. 
We encourage users to migrate to the CRPropa-native EM* modules (EMPairProduction, EMInverseComptonScattering, ...) to perform simulations of electromagnetic cascades.
Examples are available in the documentation. A simple script that can be used to replace DINT/EleCa for extragalactic propagation is available [here](https://crpropa.github.io/CRPropa3/pages/example_notebooks/secondaries/photons.v4.html).

PR #277, opened 2020-04-16
EM thinning; DINT/EleCa removal

rafaelab:
This pull request introduces thinning for electromagnetic interactions.
It also removes EleCa and DINT to favour the usage of the CRPropa-native electromagnetic modules. These external libraries for electromagnetic studies are no longer supported nor available. 

This also closes issues #199 and #219 .


include/crpropa/module/EMDoublePairProduction.h
```diff
@@ -29,12 +30,14 @@ class EMDoublePairProduction: public Module {
 	EMDoublePairProduction(
 		PhotonField photonField = CMB, //!< target photon background
 		bool haveElectrons = false,    //!< switch to create the secondary electron pair
+		double thinning = 0,           //!< weighted sampling of secondaries
```
lukasmerten, line 33: Could we add a comment here that `thinning=0` means basically no thinning?

include/crpropa/module/EMInverseComptonScattering.h
```diff
@@ -34,12 +35,14 @@ class EMInverseComptonScattering: public Module {
 	EMInverseComptonScattering(
 		PhotonField photonField = CMB, //!< target photon background
 		bool havePhotons = false,      //!< switch to create secondary photon
+		double thinning = 0,           //!< weighted sampling of secondaries
```
lukasmerten, line 38: Could we add a comment here that `thinning=0` means basically no thinning?

include/crpropa/module/EMPairProduction.h
```diff
@@ -35,12 +36,14 @@ class EMPairProduction: public Module {
 	EMPairProduction(
 		PhotonField photonField = CMB, //!< target photon background
 		bool haveElectrons = false,    //!< switch to create secondary electron pair
+		double thinning = 0,           //!< weighted sampling of secondaries
```
lukasmerten, line 39: Could we add a comment here that `thinning=0` means basically no thinning?
And in all other modules, too.

src/module/EMDoublePairProduction.cpp
```diff
@@ -8,9 +8,10 @@
 
 namespace crpropa {
 
-EMDoublePairProduction::EMDoublePairProduction(PhotonField photonField, bool haveElectrons, double limit) {
+EMDoublePairProduction::EMDoublePairProduction(PhotonField photonField, bool haveElectrons, double thinning, double limit) {
```
lukasmerten, line 11: Do old scripts, using only photonField, haveElectrons, and limit as arguments still run?

include/crpropa/module/EMPairProduction.h
rafaelab: Good idea! I've also added a better description of the thinning to the comments in the header files.

src/module/EMDoublePairProduction.cpp
rafaelab: The old scripts run as usual for the photonField and haveElectrons arguments. But I had to make a decision concerning the limit argument. Since few people set the limit anyway, I added the thinning argument BEFORE the limit to make it more accessible. This would break scripts that explicitly use this argument.

rafaelab:
The travis is not building because make test is not working.
I have to make a few adjustments. I believe this is because there is an inconsistency in the limiting of step sizes.

src/module/EMDoublePairProduction.cpp
```diff
@@ -69,8 +75,19 @@ void EMDoublePairProduction::performInteraction(Candidate *candidate) const {
 	Random &random = Random::instance();
 	Vector3d pos = random.randomInterpolatedPosition(candidate->previous.getPosition(), candidate->current.getPosition());
 
-	candidate->addSecondary( 11, Ee, pos);
-	candidate->addSecondary(-11, Ee, pos);
+	double f = Ee / E;
+	double w0 = candidate->getWeight();
+
+	if (haveElectrons) {
```
lukasmerten, line 81: Does the difference in the two if conditions -- (1-f)^\alpha compared to f^alpha -- introduce a difference between particles and anti-particles or am I missing something?

When the energy fraction f is small then electrons are produced more often than positrons, aren't they?

rafaelab: It does. It increases efficiency, but since we are not tracking all the particles charge is not conserved.

src/module/EMDoublePairProduction.cpp
lukasmerten: I agree that most people do not set these arguments but would it be possible to add a meaningfull warning, when these arguments are set? 

If not the user will probably figure out fast that a new arguments was introduced and is able to adopt his code based on the default error message.

rafaelab: @lukasmerten : do you think this is a problem?
Do you use the limit argument?

src/module/EMPairProduction.cpp
```diff
@@ -190,11 +197,20 @@ void EMPairProduction::performInteraction(Candidate *candidate) const {
...
 	double Ep = E - Ee;
+	double f = Ep / E;
 
 	// sample random position along current step
 	Vector3d pos = random.randomInterpolatedPosition(candidate->previous.getPosition(), candidate->current.getPosition());
-	candidate->addSecondary(-11, Ee / (1 + z), pos);
-	candidate->addSecondary(11, Ep / (1 + z), pos);
+	double w0 = candidate->getWeight();
+	// apply sampling
+	if (random.rand() < pow(f, thinning)) {
```
lukasmerten, line 206: Here the probabilities are now the opposite for electrons and positrons compared to EMDoublePairProduction, right?

Is this by chance or is there a reason for it; making up for some of the charge imbalence?

rafaelab: The convention doesn't really matter. The total energy is fixed. If the electron takes a fraction f of the primary energy (E), then it's energy is Ee, such that Ee+Ep=E.
If we choose a random number r and accept electrons if r<f^thinning, positrons are accepted if r<(1-f)^thinning. The only important requirement is to keep this consistency that for one particle we have f, for the other 1-f.

src/module/EMDoublePairProduction.cpp
lukasmerten: No I don't think this is a problem and I am also not setting this values explicitly. So it is probably fine as it is.

src/module/EMPairProduction.cpp
rafaelab: Electrons or positrons are accepted based on random numbers, not on charge. The particle with the highest energy fraction tends to be added to the list. For thinning=1, if the electron takes 90% of the energy, it is likely that it will be added to the stack. In this case, the positron takes 10% of the energy, so it's less likely that it will be accepted.
If, on the other hand, the positron takes 90% of the energy and the electron 10%, it is the other way around.

src/module/EMDoublePairProduction.cpp
lukasmerten: Would it be possible to just dice the sign of the particle ID?
In this way the efficiency should stay the same but on average for constant energy fractions f the number of positrons and electrons should converge to the same value.

rafaelab: Take a look at my comment below. There is no bias towards either charge, since the ones that are kept are chosen randomly based solely on the fraction of the total energy they take. The more energetic ones are kept, and their weights increased, whereas the low-energy ones will rarely be added to the array of secondaries.
If electrons and positrons have equal probability of being produced, then two random numbers are drawn and, on average, the simulation will be charge-neutral.
In other words, the higher the thinning, the more the most energetic particles are valued, since there are only a few of them compared to the many that are produced at lower energies.

lukasmerten: I agree with you that is is true for the the other module but here I am not yet convinced.

Let's assume we have a maximal thinning `thinning=1` and the primary gamma-ray has an energy `E=4*m_e c^2 ` -- which might not be realistic due to production thresholds but makes my argument more clear.

In this case, the energy for positron and electron is `Ee=(E - 2*m_e c^2)/2=m_e c^2 ` and the energy fraction is `f=1/4`. 

Then we would have a 75%=(1-f) chance of producing an electron and a 25% chance for a positron. When I read the code correctly, these chances are constant for a given input energy E. 

This is different from the other module, where Ee!=Ep and Ee will be different in each call of the performInteraction function.

One last thought:
In reality it's possible that this treatment is not problematic at all.
Assuming that `E>>m_e c^2` --> f=1/2 which will lead to a equal production rate for positrons and electrons.

rafaelab: This is not exactly true.
This implementation follows Lee 1996 mentioned in the comments. He says:
"Therefore, I use the assumption where one pair of the two carries all the initial energy and two particles in the pair share the energy equally."
Therefore, f=1/2 (not 1/4) for one electron and positron, and 0 for the others.

This is weird, I agree. DPP in this case is just the usual PP (it only produces one pair, effectively), but the cross sections are different, hence it's importance. There might be better implementations.

I could have just set f=0.5 from the beginning. I kept it this way because in the future we might want to improve this implementation and the it will require less work.

And this treatment is okay exactly for the reason you said: f=1/2 at high energies, so regardless of the thinning we get + and - charges balanced.

lukasmerten: @rafaelab Thanks for you detailed explanations. I think I just misunderstood the implementation regardless of the thinning or not. I should have taken a look into the original paper (Lee 1996). I'll try to come up with a short comment that we might add to the code, to make the procedure more clear. But this should not block your PR.

lukasmerten:
In addition to all the smaller comments, I am wondering if we want to drop DINT and EleCa with this PR or wait until the 3.2 release. Here, I am actually quite open minded and would like to read your thoughts.

src/module/EMDoublePairProduction.cpp
rafaelab: But now that you mention, although they are the same in this case, one line should be Ee, and the other Ep, for consistency. I will change that.

rafaelab:
> In addition to all the smaller comments, I am wondering if we want to drop DINT and EleCa with this PR or wait until the 3.2 release. Here, I am actually quite open minded and would like to read your thoughts.

That's a good point. On one hand, we are no longer offering support for DINT, so it makes sense to provide an efficient alternative. On the other hand, it makes sense to do all this when 3.2 is ready. Therefore, it might be a good idea to wait to merge this pull request when releasing 3.2.

rafaelab:
This pull request will be closed tomorrow. It will be split in two: the thinning part, and the removal of DINT/EleCa.

PR #281, opened 2020-04-17
Thinning of electromagnetic processes

rafaelab:
Implementation of thinning. Supersedes #277.
DINT/EleCa removal will be done in another pull request.

I suggest we wait for synchrotron thinning before merging. But this may take some time.


lukasmerten:
@rafaelab, this looks good from my side now. It would be nice if you add an example when the thinning is integrated into the main branch. 

When you implemented the thinning for synchrotron radition I will check that too.

rafaelab:
Quick update:
Even with full thinning, synchrotron is still extremely slow and somewhat memory-consuming. I couldn't even finish testing the implementation within a reasonable time.
I'm working on a "second thinning" to speed things up.

rafaelab:
The new commits introduce thinning for synchrotron.
The procedure is in part the same one used for the other EM processes. But this doesn't improve the situation a lot (only for a few steps), so another type of procedure was devised. 

The module computes dE/dx. The energy lost in the step of length dx is dE. Synchrotron photons are emitted until dE~0. The additional sampling consists in assuming that we can sample only a few of these photons with energy Ephoton at each step, such that they add to dE1<dE. The loop is stopped after a given number of samples has been reached. If this sample is large enough and the step small enough, we can add (after the loop) secondary photons of energy Ephoton with weight dE/dE1, to conserve the total energy loss at each step.

src/module/SynchrotronRadiation.cpp
```diff
@@ -8,21 +8,23 @@
...
+SynchrotronRadiation::SynchrotronRadiation(ref_ptr<MagneticField> field, bool havePhotons, double thinning, int nSamples, double limit) {
 	setField(field);
+	setBrms(0);
 	initSpectrum();
-	this->havePhotons = havePhotons;
-	this->limit = limit;
-	secondaryThreshold = 1e7 * eV;
+	setHavePhotons(havePhotons);
+	setLimit(limit);
+	setSecondaryThreshold(1e6 * eV);
```
lukasmerten, line 17: What was the reason for changing the default value of the threshold?

src/module/SynchrotronRadiation.cpp
```diff
@@ -8,21 +8,23 @@
...
-	this->Brms = Brms;
+SynchrotronRadiation::SynchrotronRadiation(double Brms, bool havePhotons, double thinning, int nSamples, double limit) {
+	setBrms(Brms);
 	initSpectrum();
-	this->havePhotons = havePhotons;
-	this->limit = limit;
-	secondaryThreshold = 1e7 * eV;
+	setHavePhotons(havePhotons);
+	setLimit(limit);
+	setSecondaryThreshold(1e6 * eV);
```
lukasmerten, line 26: Same as above

src/module/SynchrotronRadiation.cpp
rafaelab: This choice was arbitrary. The previous value had no physical meaning, neither does this. 
It would make sense to choose either 1 GeV, which is roughly the minimum energy that we treat with CRPropa, or 1 keV, since X-ray energies are typical for synchrotron emission.

src/module/SynchrotronRadiation.cpp
```diff
@@ -129,25 +147,58 @@ void SynchrotronRadiation::process(Candidate *candidate) const {
...
+		dE -= Ephoton;
+
+		// counter for sampling break condition;
+		counter++;
+	}
+
+	// while loop before gave total energy which is just a fraction of the required
+	double w1 = 1;
+	if (maximumSamples > 0 && dE > 0)
+		w1 = 1. / (1. - dE / dE0);
```
lukasmerten, line 188: This weight `w1` account for that part of the energy loss `dE` that was *left over* because of the limited number of photons that have been drawn, due to `maximumSamples`, right?

src/module/SynchrotronRadiation.cpp
```diff
@@ -129,25 +147,58 @@ void SynchrotronRadiation::process(Candidate *candidate) const {
...
+	// while loop before gave total energy which is just a fraction of the required
+	double w1 = 1;
+	if (maximumSamples > 0 && dE > 0)
+		w1 = 1. / (1. - dE / dE0); 
+
+	// loop over sampled photons and attribute weights accordingly
+	for (int i = 0; i < energies.size(); i++) {
+		double Ephoton = energies[i];
+		double f = Ephoton / (E - dE0);
+		double w = w0 * w1 / pow(f, thinning);
```
lukasmerten, line 194: The last part of the weight `1/pow(f, thinning)` then accounts for the fact that only a fraction of the sampled photons is accepted, as in the modules with thinning, right?

src/module/SynchrotronRadiation.cpp
lukasmerten: In that case I would prefer to keep the old value. Just to prevent any possible confusion.

src/module/SynchrotronRadiation.cpp
rafaelab: Exactly. It is just a correction factor to ensure energy conservation based on a sample of all the photons that would have been drawn.

The energy to be lost is dE0. But with the sampling only dE was lost within this step. The closer dE is to dE0, the lower the weights required, since the sample is large enough. For a tiny sample, only a few particles are chosen and dE/dE0 is very small, thus requiring larger weights.

src/module/SynchrotronRadiation.cpp
rafaelab: Yes. This is essentially the usual thinning method used in the other EM modules. It doesn't help a lot in most cases, but combined with the sampling, it increases the performance significantly.

rafaelab:
Thanks, @lukasmerten.
But I'd like to finish running the very long tests for the synchrotron before it is merged, so please wait a couple more weeks. So far all the scenarios I checked look okay, but you never know...
I will also change the default value back to 1e7 eV, although I would be in favour of setting it to 1e9 eV since we rarely go below that in CRPropa anyways.

lukasmerten:
@rafaelab Doing all the tests before merging is good. Just let me know, when I can help with something.

test/testInteraction.cpp
```diff
@@ -568,12 +568,11 @@ TEST(EMPairProduction, secondaries) {
 	// loop over photon backgrounds
 	for (int f = 0; f < fields.size(); f++) {
 		m.setPhotonField(fields[f]);
-		
-		// loop over energies Ep = (1E10 - 1E23) eV
-		for (int i = 0; i < 130; i++) {
+		for (int i = 0; i < 130; i++) { // loop over energies Ep = (1e10 - 1e23) eV
 			double Ep = pow(10, 10.05 + 0.1 * i) * eV;
 			Candidate c(22, Ep);
 			c.setCurrentStep(std::numeric_limits<double>::max());
```
lukasmerten, line 574: Does it make sense to use here and in the other test the same value for the currentStep? This of course should then be the smallest one: 1e3Mpc, which is probably large enough for all simulation setups anyway, isn't it?

rafaelab: For pair production it doesn't really matter which step is used and the tests converge fast.
For inverse Compton scattering and triplet pair production, if the step is very large it takes ages to finilise the test. A suitable value is ~1e4 Mpc (Hubble horizon).

Is your suggestion to change this step to the same value for all EM processes?

lukasmerten: Yes, that was my suggestion.

lukasmerten (approved):

lukasmerten:
When @rafaelab has finished his testing, I think we can merge this pull request, after the conflicts have been solved.

rafaelab:
Ah!  I didn't realise there were conflicts now. There weren't before. I will fix them soon.

lukasmerten:
@rafaelab Could you also update the changelog? Sorry, for not mentioning it earlier.

rafaelab:
> @rafaelab Could you also update the changelog? Sorry, for not mentioning it earlier.

Thanks for the reminder, @lukasmerten.
From my side this can now be merged. 
For the record, in the future my implementation should be compared with #304

PR #284, opened 2020-04-20
Implementation of SingleModeHelicalMagneticField in MagneticField

asavelie:
This contribution adds the possibility of defining a helical magnetic field with a single polarization mode in CRPropa. This includes the more specific circular and linear polarizations, but also the more general elliptical one.


lukasmerten:
Hi @asavelie 
Thanks for your implementation. It looks overall good to me. 
However, I would favour using explicit get- and set-functions to declare the variables. This would allow for some basic tests, checking whether a vector is non-zero, already in the constructor, which probably speeds up the code.

src/magneticField/MagneticField.cpp
```diff
@@ -81,13 +81,35 @@ Vector3d MagneticFieldEvolution::getField(const Vector3d &position,
...
 		}
 		return (unit_r * (unit_r.dot(moment)) * 3 - moment) / pow(r.getR() / radius, 3) * mu0 / (4*M_PI);
 }
 
+Vector3d SingleModeHelicalMagneticField::getField(const Vector3d &position) const {
+		Vector3d r = (position - origin);
+
+		Vector3d e1 = unitVectorOrigin; // first of the unit vectors which span the polarization plane
+		if ((e1.x != 0.0) && (e1.y != 0.0) && (e1.z != 0.0)) {
+			e1 /= e1.getR();
```
TobiasWinchen, line 97: I do not undertand this - do you mean to prevent division by zero? If yes use `if e1.getR() > 0`, otherwise please add comment to explain this line.

include/crpropa/magneticField/PolarizedSingleModeMagneticField.h
```diff
@@ -0,0 +1,78 @@
...
+#include "crpropa/Units.h"
+
+#include <stdexcept>
+
+namespace crpropa {
+
+/**
+
+@class PolarizedSingleModeMagneticField
+@brief General poralized single mode magnetic field (including linear, circular and elliptic polarizations)
```
lukasmerten, line 19: poralized --> polarized

asavelie:
This pull request is obsolete as it has been merged into pull request #326.

PR #286, opened 2020-04-22
introduce proper photon field class

Froehliche-Kernschmelze:
Hi everyone,

as we had discussed previously, the way of how photon fields are implemented in CRPropa needs an overhaul and where reasonable should converge towards the user-concept of magnetic fields. It is also mandatory for variable photon fields to be introduced although this PR constrains itself to the currently available photon fields.

This PR should provide an advanced draft of how such an implementation could look like. The main features are:

1) A user can call a PhotonField object in the steering file, modify it, and pass it to an interaction module as it is currently possible with B-fields. An example of how this looks like is
```
CMB = PhotonField("CMB")
ppp = PhotoPionProduction(CMB)
```

2) Interaction modules infer most photon field data from this PhotonField class. This includes the photon field density in dependence of energy and redshift, the redshift scaling factor and other parameters relevant to the photon field. Shared data files are untouched and still required by all relevant interaction modules. Features coming with this implementation are:
```
In  [1]: IRB = PhotonField("IRB_Kneiske04")

In  [2]: IRB.getFieldName()
Out [2]: "IRB_Kneiske04"

In  [3]: IRB.getHasRedshiftDependence()
Out [3]: True

In  [4]: IRB.getPhotonDensity(1 * eV, 2.1)
Out [4]: 540.77
```

3) The concept of externally defined redshift scaling factors, currently existing as shared data files, is abandoned in favor of an internal scaling factor computation from the information about the photon field itself. This was never a function the user had access to and was inferred by interaction modules to modify the interaction rate with redshift. Now one can type:
```
In  [5]: IRB.getRedshiftScaling(3.14)
Out [5]: 0.114
```
4) All photon field data is inferred from a single point defining the entire photon field. For the time being, in this draft, the photon field data is available from the ```~/Scaling/``` directory in CRPropa in the form of 3 separate files:
* photonEnergy_fieldName.txt - photon energies [J] where the field is defined
* redshift_fieldName.txt - redshifts where the field is defined (if this file doesn't exist the field is initialized with no redshift dependence)
* photonDensity_fieldName.txt comoving photon number density [1/m³]. The length of this file is either equal to photonEnergy (if no redshift dependence) or equal to len(photonEnergy) * len(redshift).
* photonEnergy and redshift do not need to be equidistant
* The three input files are checked for basic correctness upon initialisation of the photon field (correct file lengths, photonEnergies and redshift >=0 & increasing, photonDensity >= 0)

-------------------------------
**How this PR is structured:**
It might look scary at first glance due to the large amount of files modified. However, this is a minimal example of what would have to be modified, thus we do not get around to modifying a couple of files simultaneously. Yet, the type of modifications make quite clear what happened here:

1) PhotonBackground.cpp /.h
The core part: removal of enum environment for photon fields, static name-getter methods and redshift scaling class. Addition of PhotonField class
2) interactionModules .h (8x)
In the header file, the defaulting to the CMB has been disabled. What behaviour is that anyways?
3) interactionModules .cpp (8x)
Replace name-getter and redshift scaling methods by call to PhotonField class
4) testInteraction.cpp
Updated call to modules, declare all photon fields according to PhotonField syntax. Passes all tests.

-------------------------------
**Tests and supporting files**

If you pull this branch you need to generate the 3 photon-field defining data files for each photon field. This can be done in one single call of these [auxfiles.zip](https://github.com/CRPropa/CRPropa3/files/4519070/auxfiles.zip). The first you do here is to run convert_photonFields.py in the main directory of your pulled CRPropa-data clone and have the file outputted all the relevant files which then need to go into ```~/Scaling/``` or your CRPropa shared file section. The other file plot_IRB_parameters.py plots some data of the fields and compares to some aspects of the old functions. Examples include:

The photon densities of the IRB fields:
all IRB models @ z0 0

And the comparison of the internally generated redshift scaling factors to the externally generated factors generated by calc_scaling.py in the CRPropa-data repo.
all_IRB_scalingRatios

And as always, I'm thrilled to know about your thoughts!
Cheers,
Mario

### DIFF
--- a/include/crpropa/PhotonBackground.h
+++ b/include/crpropa/PhotonBackground.h
@@ -1,6 +1,7 @@
 #ifndef CRPROPA_PHOTONBACKGROUND_H
 #define CRPROPA_PHOTONBACKGROUND_H
 
+#include "crpropa/Common.h"
 #include "crpropa/Referenced.h"
 
 #include <vector>
@@ -58,29 +59,106 @@ protected:
  @brief Photon field decorator for tabulated photon fields.
 
  This class reads photon field data from files;
- the data files to be read are inferred from the field name.
- The first file must be a list of photon energies [J], named photonEnergy_fieldName.txt
- The second file must be a list of comoving photon field densities [1/m^3], named photonDensity_fieldName.txt
- Optionally, a third file contains redshifts, named redshift_fieldName.txt
+ The first file must be a list of photon energies [J], named fieldName_photonEnergy.txt
+ The second file must be a list of comoving photon field densities [1/m^3], named fieldName_photonDensity.txt
+ Optionally, a third file contains redshifts, named fieldName_redshift.txt
  */
 class TabularPhotonField: public PhotonField {
 public:
-	TabularPhotonField(const std::string fieldName);
-	// TabularPhotonField(const std::string fieldName, const bool isRedshiftDependent);
+	TabularPhotonField(const std::string fieldName, const bool isRedshiftDependent = true);
 	double getPhotonDensity(double ePhoton, double z = 0.) const;
 	double getRedshiftScaling(double z) const;
 
-protected:
-	void initPhotonEnergy(std::string fieldName);
-	void initPhotonDensity(std::string fieldName);
-	void initRedshift(std::string fieldName);
+	void initPhotonEnergy(std::string filePath);
+	void initPhotonDensity(std::string filePath);
+	void initRedshift(std::string filePath);
 	void initRedshiftScaling();
-	void checkInputData() const;
 
+protected:
 	std::vector<double> photonEnergies;
 	std::vector<double> photonDensity;
 	std::vector<double> redshifts;
 	std::vector<double> redshiftScalings;
+};
+
+class IRB_Kneiske04: public TabularPhotonField {
+public:
+	IRB_Kneiske04() : TabularPhotonField("IRB_Kneiske04", true) {
+		initPhotonEnergy(getDataPath("") + "Scaling/" + this->fieldName + "_photonEnergy.txt");
+		initPhotonDensity(getDataPath("") + "Scaling/" + this->fieldName + "_photonDensity.txt");
+		initRedshift(getDataPath("") + "Scaling/" + this->fieldName + "_redshift.txt");
+		initRedshiftScaling();
+	}
+};
+
+class IRB_Stecker05: public TabularPhotonField {
+public:
+	IRB_Stecker05() : TabularPhotonField("IRB_Stecker05", true) {
+		initPhotonEnergy(getDataPath("") + "Scaling/" + this->fieldName + "_photonEnergy.txt");
+		initPhotonDensity(getDataPath("") + "Scaling/" + this->fieldName + "_photonDensity.txt");
+		initRedshift(getDataPath("") + "Scaling/" + this->fieldName + "_redshift.txt");
+		initRedshiftScaling();
+	}
+};
+
+class IRB_Franceschini08: public TabularPhotonField {
+public:
+	IRB_Franceschini08() : TabularPhotonField("IRB_Franceschini08", true) {
+		initPhotonEnergy(getDataPath("") + "Scaling/" + this->fieldName + "_photonEnergy.txt");
+		initPhotonDensity(getDataPath("") + "Scaling/" + this->fieldName + "_photonDensity.txt");
+		initRedshift(getDataPath("") + "Scaling/" + this->fieldName + "_redshift.txt");
+		initRedshiftScaling();
+	}
+};
+
+class IRB_Finke10: public TabularPhotonField {
+public:
+	IRB_Finke10() : TabularPhotonField("IRB_Finke10", true) {
+		initPhotonEnergy(getDataPath("") + "Scaling/" + this->fieldName + "_photonEnergy.txt");
+		initPhotonDensity(getDataPath("") + "Scaling/" + this->fieldName + "_photonDensity.txt");
+		initRedshift(getDataPath("") + "Scaling/" + this->fieldName + "_redshift.txt");
+		initRedshiftScaling();
+	}
+};
+
+class IRB_Dominguez11: public TabularPhotonField {
+public:
+	IRB_Dominguez11() : TabularPhotonField("IRB_Dominguez11", true) {
+		initPhotonEnergy(getDataPath("") + "Scaling/" + this->fieldName + "_photonEnergy.txt");
+		initPhotonDensity(getDataPath("") + "Scaling/" + this->fieldName + "_photonDensity.txt");
+		initRedshift(getDataPath("") + "Scaling/" + this->fieldName + "_redshift.txt");
+		initRedshiftScaling();
+	}
+};
+
+class IRB_Gilmore12: public TabularPhotonField {
+public:
+	IRB_Gilmore12() : TabularPhotonField("IRB_Gilmore12", true) {
+		initPhotonEnergy(getDataPath("") + "Scaling/" + this->fieldName + "_photonEnergy.txt");
+		initPhotonDensity(getDataPath("") + "Scaling/" + this->fieldName + "_photonDensity.txt");
+		initRedshift(getDataPath("") + "Scaling/" + this->fieldName + "_redshift.txt");
+		initRedshiftScaling();
+	}
+};
+
+class IRB_Stecker16_upper: public TabularPhotonField {
+public:
+	IRB_Stecker16_upper() : TabularPhotonField("IRB_Stecker16_upper", true) {
+		initPhotonEnergy(getDataPath("") + "Scaling/" + this->fieldName + "_photonEnergy.txt");
+		initPhotonDensity(getDataPath("") + "Scaling/" + this->fieldName + "_photonDensity.txt");
+		initRedshift(getDataPath("") + "Scaling/" + this->fieldName + "_redshift.txt");
+		initRedshiftScaling();
+	}
+};
+
+class IRB_Stecker16_lower: public TabularPhotonField {
+public:
+	IRB_Stecker16_lower() : TabularPhotonField("IRB_Stecker16_lower", true) {
+		initPhotonEnergy(getDataPath("") + "Scaling/" + this->fieldName + "_photonEnergy.txt");
+		initPhotonDensity(getDataPath("") + "Scaling/" + this->fieldName + "_photonDensity.txt");
+		initRedshift(getDataPath("") + "Scaling/" + this->fieldName + "_redshift.txt");
+		initRedshiftScaling();
+	}
 };
 
 /**
--- a/include/crpropa/PhotonBackground.h
+++ b/include/crpropa/PhotonBackground.h
@@ -65,16 +65,17 @@ protected:
  */
 class TabularPhotonField: public PhotonField {
 public:
-	TabularPhotonField(const std::string fieldName, const bool isRedshiftDependent = true);
+	TabularPhotonField(const std::string fieldName, const bool isRedshiftDependent);
 	double getPhotonDensity(double ePhoton, double z = 0.) const;
 	double getRedshiftScaling(double z) const;
 
-	void initPhotonEnergy(std::string filePath);
-	void initPhotonDensity(std::string filePath);
-	void initRedshift(std::string filePath);
-	void initRedshiftScaling();
-
 protected:
+	void readPhotonEnergy(std::string filePath);
+	void readPhotonDensity(std::string filePath);
+	void readRedshift(std::string filePath);
+	void initRedshiftScaling();
+	void checkInputData() const;
+
 	std::vector<double> photonEnergies;
 	std::vector<double> photonDensity;
 	std::vector<double> redshifts;
@@ -91,12 +92,7 @@ protected:
  */
 class IRB_Kneiske04: public TabularPhotonField {
 public:
-	IRB_Kneiske04() : TabularPhotonField("IRB_Kneiske04", true) {
-		initPhotonEnergy(getDataPath("") + "Scaling/" + this->fieldName + "_photonEnergy.txt");
-		initPhotonDensity(getDataPath("") + "Scaling/" + this->fieldName + "_photonDensity.txt");
-		initRedshift(getDataPath("") + "Scaling/" + this->fieldName + "_redshift.txt");
-		initRedshiftScaling();
-	}
+	IRB_Kneiske04() : TabularPhotonField("IRB_Kneiske04", true) {}
 };
 
 /**
@@ -109,12 +105,7 @@ public:
  */
 class IRB_Stecker05: public TabularPhotonField {
 public:
-	IRB_Stecker05() : TabularPhotonField("IRB_Stecker05", true) {
-		initPhotonEnergy(getDataPath("") + "Scaling/" + this->fieldName + "_photonEnergy.txt");
-		initPhotonDensity(getDataPath("") + "Scaling/" + this->fieldName + "_photonDensity.txt");
-		initRedshift(getDataPath("") + "Scaling/" + this->fieldName + "_redshift.txt");
-		initRedshiftScaling();
-	}
+	IRB_Stecker05() : TabularPhotonField("IRB_Stecker05", true) {}
 };
 
 /**
@@ -127,12 +118,7 @@ public:
  */
 class IRB_Franceschini08: public TabularPhotonField {
 public:
-	IRB_Franceschini08() : TabularPhotonField("IRB_Franceschini08", true) {
-		initPhotonEnergy(getDataPath("") + "Scaling/" + this->fieldName + "_photonEnergy.txt");
-		initPhotonDensity(getDataPath("") + "Scaling/" + this->fieldName + "_photonDensity.txt");
-		initRedshift(getDataPath("") + "Scaling/" + this->fieldName + "_redshift.txt");
-		initRedshiftScaling();
-	}
+	IRB_Franceschini08() : TabularPhotonField("IRB_Franceschini08", true) {}
 };
 
 /**
@@ -145,12 +131,7 @@ public:
  */
 class IRB_Finke10: public TabularPhotonField {
 public:
-	IRB_Finke10() : TabularPhotonField("IRB_Finke10", true) {
-		initPhotonEnergy(getDataPath("") + "Scaling/" + this->fieldName + "_photonEnergy.txt");
-		initPhotonDensity(getDataPath("") + "Scaling/" + this->fieldName + "_photonDensity.txt");
-		initRedshift(getDataPath("") + "Scaling/" + this->fieldName + "_redshift.txt");
-		initRedshiftScaling();
-	}
+	IRB_Finke10() : TabularPhotonField("IRB_Finke10", true) {}
 };
 
 /**
@@ -163,12 +144,7 @@ public:
  */
 class IRB_Dominguez11: public TabularPhotonField {
 public:
-	IRB_Dominguez11() : TabularPhotonField("IRB_Dominguez11", true) {
-		initPhotonEnergy(getDataPath("") + "Scaling/" + this->fieldName + "_photonEnergy.txt");
-		initPhotonDensity(getDataPath("") + "Scaling/" + this->fieldName + "_photonDensity.txt");
-		initRedshift(getDataPath("") + "Scaling/" + this->fieldName + "_redshift.txt");
-		initRedshiftScaling();
-	}
+	IRB_Dominguez11() : TabularPhotonField("IRB_Dominguez11", true) {}
 };
 
 /**
@@ -181,12 +157,7 @@ public:
  */
 class IRB_Gilmore12: public TabularPhotonField {
 public:
-	IRB_Gilmore12() : TabularPhotonField("IRB_Gilmore12", true) {
-		initPhotonEnergy(getDataPath("") + "Scaling/" + this->fieldName + "_photonEnergy.txt");
-		initPhotonDensity(getDataPath("") + "Scaling/" + this->fieldName + "_photonDensity.txt");
-		initRedshift(getDataPath("") + "Scaling/" + this->fieldName + "_redshift.txt");
-		initRedshiftScaling();
-	}
+	IRB_Gilmore12() : TabularPhotonField("IRB_Gilmore12", true) {}
 };
 
 /**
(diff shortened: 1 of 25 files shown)


Froehliche-Kernschmelze:
Hi there,

I know that this looks a but scary and maybe not appealing due to the tests seeming to fail. Maybe a short wrap-up gets you interested?
* The tests fail because the tab files have been modified. Travis runs the installation on tab files that are in the original data repo and don't suit the new photon field syntax.
* The "big" deal is here that we need to find a consensus on how photon fields are defined in CRPropa. My suggestion: Have 3 files in the Scaling dir with 1) photon energy 2) redshift 3) density

adundovi:
Hi Mario, I'll take a look later afternoon into your PR.

Froehliche-Kernschmelze:
Thank you :)
...I shall also provide the [photonFieldData](https://github.com/CRPropa/CRPropa3/files/4586775/photonFieldData.zip) which go into the ```~/Scaling``` directory. This should give you the entire picture of what happened.

adundovi:
Hi @Froehliche-Kernschmelze,
you did a really good job in generalizing the current CRPropa photon field implementation. And yes, although you described this PR in great detail and modified a lot of files, the main idea is not scary :-)

My first thoughts:
- If we are introducing already a new interface `PhotonField` instead of the enumerator, I would like to use this opportunity to generalize it a step further. I dislike relying on tabulated data if it is not strictly necessary. Why? The software, in general, is opaque for users while putting a bunch of "magic" numbers to it makes it even worse. I know that we cannot provide (semi)analytical models of all photon fields, and I know we have data scripts in another repository that should take partially care of this, but at least we can provide an analytic PhotonField class for CMB, i.e., without tables [1](https://github.com/cosmicrays/hermes/blob/master/include/hermes/photonfields/CMB.h#L43), no? I mean, getters are already here, so one shouldn't think where the photon density is coming from? So, I propose that your class becomes `TabularPhotonField` (or similar) which would become a child of `PhotonField` (abstract class). Another child of it would be `CmbPhotonField`, for example. By making this example, we will stimulate users/developers to add new models in CRPropa as a code, not as tables. A code is transparent, maintainable and modifiable, tables are not. One more technical detail: `PhotonField` should be inherited from `Referenced` so it can work correctly from SWIG/Python.
- `getHasRedshiftDependence` > `hasRedshiftDependence` (the second is easier to read, its purpose is still clear and we use similar wording elsewhere).

You can provide everything as tables for this PR (including CMB), but I would prefer that we already have this additional level of abstractness to support PhotonField models that are not necessary tables.

TobiasWinchen:
I agree with @adundovi - as for the magnetic fields, we probably should aim to have the models directly in the code directly at some point so that data files can be generated on the fly if needed and we get more flexibility for the used fields. 

To be clear, for this PR this means only yo introduce an abstract PhotonField interface and rename your class TabulatedField if you agree with the idea. Non-tabulated CMB can be introduced also later.

Froehliche-Kernschmelze:
Hi again, thanks for your thoughts. I am very happy with how this implementation can be improved and here comes the current status of what this looks like for now.
Brief:
* ```PhotonField``` became abstract
* ```PhotonField``` inherits from ```Referenced```
* added child ```TabularPhotonField``` for IRB models
* added child ```BlackbodyPhotonField``` which if called with T=2.73 K becomes the CMB
* adapted syntax in test files accordingly
Notes:
* I have implemented the black body class as the first analytical photon field in CRPropa (following @adundovi) yet so far we do not (yet!) support any black body other than CMB
* This PR assumes that each tabular photon field is composed of up to 3 files in the Scaling dir (eps, z, dens). Such files need to be provided for all IRB files (they already exist in this unified format but are not included in this PR so far)
* The support of URB_Protheroe96 has been dropped for now. The reason is that photon field data in the CRPropa-data repo does not seem to be complete for this field to have its 3 base files generated. I might have a look into this first (?)

include/crpropa/PhotonBackground.h
```diff
@@ -2,30 +2,50 @@
...
 #include "crpropa/Common.h"
+#include "crpropa/Referenced.h"
 
 #include <string>
 
 namespace crpropa {
-
 /**
- * \addtogroup EnergyLosses
+ * \addtogroup PhotonFields
```
TobiasWinchen, line 11: As you added the new doxygen group, you should also add appropriate documentation in the new system, i.e. one file buildingblocks/photonfields.rst and also adding the new file to building_blocks.rst.

include/crpropa/PhotonBackground.h
```diff
@@ -2,30 +2,50 @@
 #define CRPROPA_PHOTONBACKGROUND_H
 
 #include "crpropa/Common.h"
```
TobiasWinchen, line 4: Why is Common needed in the header?

src/PhotonBackground.cpp
```diff
@@ -75,12 +65,12 @@ void PhotonField::initPhotonEnergy(std::string fieldName) {
...
 }
 
-void PhotonField::initPhotonDensity(std::string fieldName) {
+void TabularPhotonField::initPhotonDensity(std::string fieldName) {
 	std::string path = getDataPath("Scaling/photonDensity_" + fieldName + ".txt");
 	std::ifstream infile(path.c_str());
 
 	if (!infile.good())
-		throw std::runtime_error("PhotonField::initPhotonDensity: could not open file photonDensity_" + fieldName);
+		throw std::runtime_error("TabularPhotonField::initPhotonDensity: could not open file photonDensity_" + fieldName);
```
TobiasWinchen, line 73: printing the path returned bu getDataPath would be more helpful

src/PhotonBackground.cpp
```diff
@@ -51,20 +41,20 @@ double PhotonField::getRedshiftScaling(double z) const {
...
 }
 
-void PhotonField::initPhotonEnergy(std::string fieldName) {
+void TabularPhotonField::initPhotonEnergy(std::string fieldName) {
 	std::string path = getDataPath("Scaling/photonEnergy_" + fieldName + ".txt");
 	std::ifstream infile(path.c_str());
 
 	if (!infile.good())
-		throw std::runtime_error("PhotonField::initPhotonEnergy: could not open file photonEnergy_" + fieldName);
+		throw std::runtime_error("TabularPhotonField::initPhotonEnergy: could not open file photonEnergy_" + fieldName);
```
TobiasWinchen, line 57: printing the path returned bu getDataPath would be more helpful

src/PhotonBackground.cpp
```diff
@@ -122,52 +112,61 @@ void PhotonField::initRedshiftScaling() {
...
 	}
 }
 
+BlackbodyPhotonField::BlackbodyPhotonField(std::string fieldName, double blackbodyTemperature) {
+	this->fieldName = fieldName;
+	this->blackbodyTemperature = blackbodyTemperature;
+}
+
+double BlackbodyPhotonField::getPhotonDensity(double ePhoton) const {
+	return 8 * M_PI * std::pow(ePhoton / (h_planck * c_light), 3) / (std::exp(ePhoton / (k_boltzmann * this->blackbodyTemperature)) - 1);
```
TobiasWinchen, line 167: you should use pow_integer<3>( ... ) instead of std::pow as this (see Common.h) is 10x faster than std:pow for integers. Even a loop is up to 2x faster for integers (if the exponent is not known at compile time).

std::exp( X ) -1 should be replaced by std::expm1(X) due to precision

include/crpropa/module/EMDoublePairProduction.h
```diff
@@ -27,7 +27,7 @@ class EMDoublePairProduction: public Module {
 
 public:
 	EMDoublePairProduction(
-		PhotonField photonField = CMB, //!< target photon background
+		PhotonField photonField, 	   //!< target photon background
```
TobiasWinchen, line 30: As you change from a small type to a class, we need to change all constructors here to use ref_ptr, i.e. 

EMDoublePairProduction(ref_ptr<PhotonField> photonField)

src/module/EMDoublePairProduction.cpp
```diff
@@ -88,7 +88,7 @@ void EMDoublePairProduction::process(Candidate *candidate) const {
 
 	// interaction rate
 	double rate = interpolate(E, tabEnergy, tabRate);
-	rate *= pow(1 + z, 2) * photonFieldScaling(photonField, z);
+	rate *= pow(1 + z, 2) * photonField.getRedshiftScaling(z);
```
TobiasWinchen, line 91: use pow_integer

TobiasWinchen:
Regarding the change of data format, we should probably introduce versioned data downloads here to enable consistent changes and tests and enable users to use old versions, which otherwise would download wrong data. The version can be just a number So please 
1) create a new data tarball
2) Upload it with name data.tar.gz_1 to the download location
3( Change line 323 of the Cmake lists to  https://www.desy.de/~crpropa/data/interaction_data/data.tar.gz_1

Note that it still should be stored as data.tar.gz and extract to /data/..., to avoid problems in other places.

TobiasWinchen:
Ah, and could you also update the Changelog and include also a comment explaining the change in IRB scaling to the users as this might affect the physics results. It is probably best if we update the changelog continously with every PR to ease the work for the release.

include/crpropa/PhotonBackground.h
```diff
@@ -2,30 +2,50 @@
...
  @class PhotonField
- @brief photon field class fully defined with a range of photon energies, redshift and the field's density
+ @brief Abstract base class for photon fields.
  */
-class PhotonField {
+class PhotonField: public Referenced {
 public:
-	PhotonField(std::string fieldName, bool hasRedshiftDependence = true);
-	PhotonField() : PhotonField("CMB", false) {};
+	virtual double getPhotonDensity(double ePhoton, double z = 0.) const {
```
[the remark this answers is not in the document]
Froehliche-Kernschmelze: In this case all constructors in the interaction modules .h files require explicitly if they are supposed to expect a ```TabularPhotonField``` or a ```BlackbodyPhotonField```. Having a constructor for each type of photon field is possible but would also make it more difficult to introduce more types of photon fields in the future. May I not be seeing something simpler here?

Froehliche-Kernschmelze:
> Nice work. Could you resolve the merge conflict (and potentially rebase on master) so that travis will build the PR? I currently get several failing tests due to undefined symbol: _ZNK7crpropa18TabularPhotonField21hasRedshiftDependenceEv'

Thanks a lot for your feedback! This was a bug in how TabularPhotonField has inherited which is fixed now, next to a few more points you highlighted. This should pass the tests now, a rebase will follow as well later

include/crpropa/PhotonBackground.h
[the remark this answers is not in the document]
TobiasWinchen: This of course requires that the interaction classes have only pointers (ref_ptr) to the PhotonField in their constructor and not a copy of the field as now. 
A class can still be accessed via a pointer to a parent, only no object of type A can be constructed if there are only pure virtual methods. 
ref_ptr<A> acts as pointer A* , thus the confusion? Making them pure virtual only prevents that an abstract PhotonField is constructed.  See e.g. the RestrictToRegion module that has a pointer to a Surface which has pure virtual methods.
However, as this is similar to the MagneticField to have the no field defined as default, this can also stay.

TobiasWinchen:
> Thanks a lot for your feedback! This was a bug in how TabularPhotonField has inherited which is fixed now, next to a few more points you highlighted. This should pass the tests now, a rebase will follow as well later

There is still the merge conflict in test/testSimulationExecution.py.in - this prevents travis, the automated build system, to test this PR.

include/crpropa/PhotonBackground.h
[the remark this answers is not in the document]
Froehliche-Kernschmelze: Good point indeed. Current implementation follows your idea and uses ref_ptr to PhotonField in all interaction modules while PhotonField now is an abstract class. Hate to bug you with it but in the current implementation it seems that in the steering file a call like
```
CMB = BlackbodyPhotonField("CMB", 2.73)
PPP = PhotoPionProduction(CMB)
```
returns a type error, explicitly demanding a ref_ptr<PhotonField> object (obviously...):
```
NotImplementedError: Wrong number or type of arguments for overloaded function 'new_PhotoPionProduction'.
  Possible C/C++ prototypes are:
    crpropa::PhotoPionProduction::PhotoPionProduction(crpropa::ref_ptr<rpropa::PhotonField>, ...)
```
But why does this occur in the Python call? The syntax is very similar to how magnetic fields are passed to the propagation modules and there we do not see this behaviour. Do you have an educated guess here?

lukasmerten:
You also have to make some changes in the 2_headers.i file. For all other ref_ptrs there are some declaration like this

`%ignore operator crpropa::MagneticField*;`

and this
`%implicitconv crpropa::ref_ptr<crpropa::MagneticField>;
%template(MagneticFieldRefPtr) crpropa::ref_ptr<crpropa::MagneticField>;
%include "crpropa/magneticField/MagneticField.h"` 

in the case of the MagneticField.

include/crpropa/PhotonBackground.h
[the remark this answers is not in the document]
Froehliche-Kernschmelze: > You also have to make some changes in the 2_headers.i file. For all other ref_ptrs there are some declaration like this
> 
> `%ignore operator crpropa::MagneticField*;`
> 
> and this
> `%implicitconv crpropa::ref_ptr<crpropa::MagneticField>; %template(MagneticFieldRefPtr) crpropa::ref_ptr<crpropa::MagneticField>; %include "crpropa/magneticField/MagneticField.h"`
> 
> in the case of the MagneticField.
That totally did the trick, thanks a huge lot!!!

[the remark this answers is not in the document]
Froehliche-Kernschmelze: solved by @lukasmerten, trick was to update the python 2_headers.i file accordingly

include/crpropa/PhotonBackground.h
```diff
@@ -2,30 +2,50 @@
...
+		return 0.;
+	};
+	virtual double getRedshiftScaling(double z) const {  // returns overall comoving scaling factor (cf. CRPropa3-data/calc_scaling.py)
```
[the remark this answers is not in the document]
Froehliche-Kernschmelze: by making getPhotonDensity pure virtual, PhotonField has become abstract and would therefore suit our requirement for that class to not be instantiated.
Could one therefore leave getRedshiftScaling semi-virtual since it is always worthwhile for this function to return 1. unless needed otherwise?

Froehliche-Kernschmelze:
The very most issues have been resolved now (except for the tarball modification). I have furthermore modified this PR according to:
* Introduce all previously defined photon fields esplicitly, e.g. CMB(), IRB_Kneiske04() as children of BlackbodyPhotonField and TabularPhotonField. The user concept would then be very similar to the current version:
```
In  [1]: CMB = CMB()
In  [2]: PPP = PhotoPionProduction(CMB)  # or directly:
In  [3]: PPP2 = PhotoPionProduction(IRB_Kneiske04())
```
The current version in this PR compiles just fine and enables the user to use these features as intended, yet these declarations do not seem to be working in ```test/testInteraction.cpp``` - Here is what happens:
```
#include "crpropa/PhotonBackground.h"
...
ref_ptr<PhotonField> CMB = new BlackbodyPhotonField("CMB", 2.73);  -> works but is now just a dummy initialisation
ref_ptr<PhotonField> CMB = new CMB();  // not working! Compiler: "CMB does not name a type"
```
What to know and what I tried:
* CMB is a child of BlackbodyPhotonField declared in PhotonBackground.h which is included properly in testInteraction.cpp
* the same issue occurs with the IRB tabular children
* tried typedef
* tried template
* tried forward declaration
Maybe I just did not get my attempts right. Do you have any suggestions here?

Also I hope you like the idea of having the "native" photon fields ready-to-use declared already as we did with the enum syntax in photon fields and the special-case magnetic fields already.

Froehliche-Kernschmelze:
> Regarding the change of data format, we should probably introduce versioned data downloads here to enable consistent changes and tests and enable users to use old versions, which otherwise would download wrong data. The version can be just a number So please
> 
>     1. create a new data tarball
> 
>     2. Upload it with name data.tar.gz_1 to the download location
> 
>     3. Change line 323 of the Cmake lists to  https://www.desy.de/~crpropa/data/interaction_data/data.tar.gz_1
> 
> 
> Note that it still should be stored as data.tar.gz and extract to /data/..., to avoid problems in other places.

Hi @TobiasWinchen,
I am ready to be reviewed again. Very short summary:
* happily addressed all your comments
* exception: I have the data tarball yet I don't know how to upload/authentify to the location at desy
* I cheated in ```test/testInteraction.cpp``` where I initialised photon fields via their non-special-name constructors since I didn't manage to get them visible there. Passes all tests though.

include/crpropa/PhotonBackground.h
```diff
@@ -65,7 +65,7 @@ class PhotonField: public Referenced {
  */
 class TabularPhotonField: public PhotonField {
 public:
-	TabularPhotonField(const std::string fieldName, const bool isRedshiftDependent);
+	TabularPhotonField(const std::string fieldName, const bool isRedshiftDependent = true);
```
lukasmerten, line 68: I am not sure if this is connected to ur discussion of PR #255?!
We decided there to have the non-tabulated redshift evolution as the default because of possible inconsistencies between modules.

Is that inconsitency resolved by this PR or is there any other reason to change the default behavior or am I missing something?

Froehliche-Kernschmelze: well it can be but here this is more involved:

1) ```isRedshiftDependent``` is a property of ```(Tabular)PhotonField``` and helps that class to call appropriate interpolation routines, check input data, calc redshift scalings, etc. which all are properties of the new photon field structures, too
2) Note how all built-in photon fields (aka. all the IRB models) have that property already by being instantiated with ```isRedshiftDependent = true```.
3) #255 in return refers to a separate property ```haveRedshiftDependence``` in ```PhotoPionProduction``` which the user actually is free to choose. The main reason here is, I guess, that the user may choose between the 1D and the more expensive 2D interpolation for runtime reasons (?).
4) In my eyes, that parameter should be abolished and redshift dependence could now be inferred from the photon field at hand. I did not crunch the numbers here but the most severe concern of runtime in the PPP is the threading bottleneck of SOPHIA and less the choice of 1D or 2D interpolations.

TobiasWinchen:
Sorry @Froehliche-Kernschmelze, but I won't have time this week, but probably next week.

Froehliche-Kernschmelze:
> Sorry @Froehliche-Kernschmelze, but I won't have time this week, but probably next week.

Thanks for letting me know. To make it easier for you I have temporarily provided that data in a branch of my repo, having the CMakeLists.txt changed accordingly in line 323. Formal testing still fails somehow... but you could still test it by applying that [data](https://github.com/Froehliche-Kernschmelze/CRPropa3/blob/provideData/data_1.tar.gz) manually.

include/crpropa/PhotonBackground.h
lukasmerten: I mixed up `isRedshiftDepentend` with `haveRedshiftDependence`.
So basically my issue concerning this PR is solved.

In addition:
4) Abolishing that and taking that value from the Photonfield class is in general a good idea. 

My original concern here was, that (at least in the current master) we do not have 2 dimensional tabulated files for the rates for other modules than PPP. So the reason for choosing 1d over 2d might not only be runtime driven but also by wishing for consistency between different interaction losses, e.g. PPP (2 dimensional rates available) and Photodisintegration (only 1 dimensional rates available).

Froehliche-Kernschmelze:
Hey @adundovi - thanks a lot for stepping in here as well. I am yet about to find out if there is a way to have ranged field quantities such as photon energies and redshift generated. Redshift sure enough is easy to realise but for the photon energies this is to be investigated.

Nevertheless:
* I have removed the PowerLawPhotonField so far
* BlackBodyPhotonField could receive a private constructor until later PRs which effectively makes the CMB available only here.
* The most help would be needed now in /testInteraction.cpp - there I can't get the special-case classes visible like CMB, IRB_Kneiske04, etc. However, e.g. TabularPhotonField works. Do you have a clue here? Tried to address this in many ways (see above).

adundovi:
@Froehliche-Kernschmelze I'm sorry for the delay. I've cloned your branch, compiled it and tested.

First we have to upload your new data files. Is the archive you provided above the complete one with everything required for CRPropa to function properly?

What is the issue with `testInteraction.cpp`?
I've run tests and it fails here:
```bash
.../crpropa3_build/test/testInteraction.cpp:478: Failure
Expected: (c.current.getEnergy()) < (100 * EeV), actual: 16.0218 vs 16.0218
[  FAILED  ] PhotoPionProduction.proton (1 ms)
[ RUN      ] PhotoPionProduction.helium
.../crpropa3_build/test/testInteraction.cpp:496: Failure
Expected: (c.current.getEnergy()) < (400 * EeV), actual: 64.0871 vs 64.0871
.../crpropa3_build/test/testInteraction.cpp:498: Failure
Value of: massNumber(id) < 4
  Actual: false
Expected: true
.../crpropa3_build/test/testInteraction.cpp:499: Failure
Value of: c.secondaries.size() > 0
  Actual: false
Expected: true
[  FAILED  ] PhotoPionProduction.helium (1 ms)
[ RUN      ] PhotoPionProduction.secondaries
.../crpropa3_build/test/testInteraction.cpp:531: Failure
Expected: (c.secondaries.size()) > (1), actual: 0 vs 1
[  FAILED  ] PhotoPionProduction.secondaries (0 ms)
```

Froehliche-Kernschmelze:
Curiously, I do not experience any issues at the points you described while running the test. A manual checkup yields that everything works as expected. My best guess here is that it could be an issue with the data files. I suggest doing a minimal example:

* reset all shared files of my branch to the current master branch version
* from the files in data_1 I suggest adding [above](https://github.com/Froehliche-Kernschmelze/CRPropa3/blob/provideData/data_1.tar.gz), only the content of ```/data_1/Scaling/``` matters
* copy the 24 files of that directory into the ```/Scaling/``` directory of your CRPropa installation
* recompile for safety and run the tests - this should show if it was the data files being corrupted

...afterwards, if that is up running, I suggest to focus on the syntactic issue in ```testInteraction.cpp```

adundovi:
Hi @Froehliche-Kernschmelze, can you please merge #300 into your PR and change in `CMakeLists.txt`: **2017-11-09** to **2020-05-25** which is your data file (data_1) that I already uploaded to the CRPropa data download path. I believe that this fixes the last issue and makes continuous integration work again so we can finally merge your PR.

include/crpropa/PhotonBackground.h
```diff
@@ -2,36 +2,214 @@
...
+	std::vector<double> redshiftScalings;
+};
+
+/**
+ @class IRB_Kneiske04
+ @brief Extragalactic background light model from Kneiske et al. 2004
+
+ Source info:
+ DOI:10.1051/0004-6361:20031542,
+ https://www.aanda.org/articles/aa/pdf/2004/03/aa3848.pdf, figure 1 (”Best-fit” model)
```
adundovi, line 91: @Froehliche-Kernschmelze Please change quotation marks here into regular (ASCII) ones, i.e. to "Best-fit". Otherwise, it seems that we cannot generate documentation from this comment (in TravisCI). Then commit.

adundovi:
> Nice work. Could you resolve the merge conflict (and potentially rebase on master) so that travis will build the PR? I currently get several failing tests due to undefined symbol: _ZNK7crpropa18TabularPhotonField21hasRedshiftDependenceEv'

Your review request should be fulfilled now. @Froehliche-Kernschmelze has rebased this PR to the current master. I think only the quotation marks in the source comment are now blocking TravisCI, but we'll fix that too soon.

Issue #267, opened 2020-01-22
Implementation of turbulent magnetic field from Tautz and Dosch 2013 


tfitoussi:
Dear all,

As far as I know CRPropa currently uses the algorithm of [Giacalone & Jokipii 1999](https://ui.adsabs.harvard.edu/abs/1999ApJ...520..204G/abstract) to generate turbulent magnetic field. In 2012, [R. C. Tautz](https://ui.adsabs.harvard.edu/abs/2012JCoPh.231.4537T/abstract)  pointed out that the implementation of a turbulent magnetic field with Giacalone & Jokipii 1999 is problematic. 

"For the test case of isotropic turbulence, the three basic requirements for the turbulence generator are: 

1. all three turbulent magnetic field components have the same magnitude on average;
2. the probability for the random orientation of the sample Fourier modes is isotropic;
3. the divergence of the turbulent magnetic field must be zero. 

As shown recently, this was not possible using previously used approaches because at least one condition was always violated." ([Tautz & Dosch 2013](https://ui.adsabs.harvard.edu/abs/2013PhPl...20b2302T/abstract))

In [Tautz & Dosch 2013](https://ui.adsabs.harvard.edu/abs/2013PhPl...20b2302T/abstract) is presented a modified version of Giacalone & Jokipii 1999 that fulfills all three requirements. Moreover, it should start to have a turbulent behavior with only 8 Fourier mode and should not require more than 16 modes since results are almost indistinguishable with higher numbers of modes.

The low number of modes should allow a computation "on the fly" without using a grid. Moreover, this version does not require Fourier transformations.

I started a long time ago to develop the algorithm in [this branch](https://github.com/tfitoussi/CRPropa3/tree/td13field) (TD13Field in magnetic field folder). I tested the implementation by reproducing the results of section II-C and in particular figure 2 from the article. Here are the [pictures and script to produce it](https://github.com/CRPropa/CRPropa3/files/4098141/TD13.tar.gz).

From this test, it seems that a correct turbulent magnetic field requires at least 64 modes not 8. Moreover, I do not find the same error on the total turbulent magnetic field strength. I guess there is a mistake somewhere in the code but I did not find it.

I have no more time to work on it now but if someone wants to take a look and contribute, he or she is welcome. 

Cheers,

Thomas

lukasmerten:
Hi Thomas

[...]

PR #287, opened 2020-04-30
Introduce new TurbulentField abstract class for turbulent field models

adundovi:
This PR is not yet meant to be merged. It will serve as a tracker for the rewriting turbulent field features.

I introduce a new class called TurbulentField (following the naming scheme of MagneticField). It is an abstract class that inherits MagneticField since it should be a pluggable component of propagation modules. Methods of this class should establish an interface for common operations on turbulent fields - the most obvious example is the method to obtain the correlation length (which is model dependent, but every synthetic turbulent field model has to have it).


TobiasWinchen:
Would it make sense to separate grid construction from field construction? The grid is stored using reference counted pointers anyway, so it is intended do be used by more than one class. It would make sense to not construct the grid inside the Turbulent field but outside -  This makes reusing the same grid for e.g. spatial varying photon fields, dust distributions etc. easier. Also in the future the system can be extended more easily to non-uniform grids.

adundovi:
Hi @TobiasWinchen, thanks for your comment. I'm not even sure myself what to do with it. The current (old) approach is certainly too rigid, i.e., that one particular model of a field is provided by a core function. It is hard to reuse it and to extend it. But your idea could work: to every (grid) turbulent field model one provides a grid constructed from the outside. I'll do that.

TobiasWinchen:
Probably it would be best to completely separate fields and grid as the one is defined on the other, so that the simulation setup could look like this:
`
mygrid = Grid(...)
coherent_field = MyMagneticField(mygrid, B = 1nG, ...)
turbulent_field = MyTurbulentMagneticField(mygrid, B = 1nG, ...)
extra_compnent = MyExtraGridlessComponent( ... )
magnetic_field = composite_magnetic_field(coherent_field turbulent_field, extra_component)
# magnetic_field.addComponent( .... )

photon_field = MyPhotonField(mygrid, ...)
dust_field = MyDustField(mygrid, ...)
`

In the future we could then make this potentially work for arbitrary grids if necessary.

adundovi:
But do we make a copy of the grid inside of classes that modify it or we change the original one? How can we reuse a grid which we modify with `MyTurbulentMagneticField` later in `MyPhotonField`?

TobiasWinchen:
Ok, never mind - not enough coffee yet. Of course the grid is here the data container for the field, so my comments do not make much sense.

adundovi:
> Ok, never mind - not enough coffee yet. Of course the grid is here the data container for the field, so my comments do not make much sense.

:-) OK, so you suggest that I embed this Grid into `GridTurbulence` classes after all? Instead, I can make a copy at the constructor. This would simplify the interface any way.

adundovi:
> Instead, I can make a copy at the constructor. This would simplify the interface any way.
No, this will not work because for large grids, that would unnecessary waste memory. So, either we create the grid inside of these classes, or move its ownership if it is created outside. But the latter option doesn't work well with SWIG and Python logic I'm afraid.

adundovi:
With the newest commit to this PR, I've introduced a new struct `GridProperties` that combines all parameters needed for creation of a new Grid - has the same constructors as the Grid class, but it can be also passed to the Grid constructor. I use it for the GridTurbulence constructor that simplifies the interface as follows:

```python
turbulent_field = MyTurbulentMagneticField(
      GridProperties(origin, size, spacing),
      Brms = 1*muG, ...)
)
```
Maybe also introducing the same concept for the turbulence spectrum and that would be the final interface of new classes in this PR.

adundovi:
This PR is now ready for review.

I will improve and de-duplicate code in turbulence models later (in next PRs) but the functionality, the interfaces and the code structure should not change. The new structure and interfaces should be fruitful for implementation of other turbulent models in the future.

To explain notable changes:
1. `GridProperties` is a new struct that serves to cleanly pass grid parameters around. I've introduced it to reduce the number of arguments for grid turbulence models (otherwise, there is a great chance for mistakes when one needs to enter 8+ doubles to constructors). It doesn't change the interface of the old code, only the `Grid` class acquired a new constructor which accept `GridProperties`.
2. `gridPowerSpectrum` is a new method that calculates a power-spectrum of a turbulent field defined on a grid. It uses FFTW, but from real-space to k-space (the opposite to the grid turbulence models).
3. `GridTurbulence.h` and `GridTurbulence.cpp` are removed, and the code from there is moved to the below-mentioned classes.
4. `GridTurbulence` is a new class that implements a textbook example of the synthetic turbulence defined on a grid.
5. `HelicalGridTurbulence` is a new class of the old function which provides a turbulent model with helicity. I also provide the old interface for it, so the existing code will work, however, I marked it as obsolete. @rafaelab 
6. `PlaneWaveTurbulence` is a new model of turbulence that uses the plane-wave instead of the grid approach to generate turbulence (following Tautz and Dosch 2013, see #267). It is implemented thanks to the effort of @antoniusf and @tfitoussi 
7. `SimpleGridTurbulence` is the original turbulent model that was already present in the code from before (i.e., `initTurbulence`). Now it is converted to the class, but I still provide a "hack" to use it with `initTurbulence`. In some future release we will remove these functions and simplify this class a bit by removing static methods, etc., etc..
8. `TurbulentField.h` provides two new classes, one is `TurbulentField` which all other listed classes inherit and provides the basic interface for models of turbulence. In the same file, you will find `TurbulenceSpectrum` which is a class that provides the energy spectrum for all models except the old ones (`HelicalGridTurbulence` and `SimpleGridTurbulence`). In principle, it can be inherited by new spectra equations, however, one has to provide a correct correlation length formulae for these.

I will introduce more tests in the meantime and fix typos. Nevertheless, I would already appreciate if you can give comments and start the review process. Thanks.

antoniusf:
Awesome, thank you so much!

lukasmerten:
I hope to be able to take closer look tomorrow or on Friday. Please remind me if that's not happening.

include/crpropa/magneticField/turbulentField/PlaneWaveTurbulence.h
```diff
@@ -0,0 +1,110 @@
...
+
+namespace crpropa {
+/**
+ * \addtogroup MagneticFields
+ * @{
+ */
+
+/**
+ @class PlaneWaveTurbulence
+ @brief Interpolation-free turbulent magnetic field based on the GJ99 and TD13
```
lukasmerten, line 16: gj99 and td13 - lowercase - to be consistent

include/crpropa/magneticField/turbulentField/PlaneWaveTurbulence.h
```diff
@@ -0,0 +1,110 @@
...
+which then needs to be interpolated to obtain in-between values -- this method
+only generates the wave modes making up the turbulent magnetic field ahead of
+time. At run time, when the field's value at a particular position is required,
+these plane waves are then evaluated analytically at that position. This
+guarantees that the resulting field is completely free of divergence, reproduces
+the mean field strength accurately, and does not suffer from other
+interpolation-induced problems. The disadvantage is that the number of wave
+modes is drastically smaller when compared with initTurbulence, which might have
+physical ramifications on the particles propagating through the field.
+Additionally, the implementation is somewhat slower.
```
lukasmerten, line 32: Can we specify this "somewhat"? If not I would drop that sentence.

include/crpropa/magneticField/turbulentField/PlaneWaveTurbulence.h
```diff
@@ -0,0 +1,110 @@
...
+
+ ## Using the SIMD optimization
+ In order to mitigate some of the performance impact that is inherent in this
+method of field generation, an optimized version utilizing data-level
+parallelism through SIMD instructions is provided. More specifically, this
+implementation uses the AVX extension to the x86 instruction set. In order to
+use this optimized version, three conditions need to be met:
+
+1. The `USE_SIMD` option needs to be explicitly enabled in CMake. Currently,
+this sets GCC flags that tell the compiler to allow SIMD instructions.
```
lukasmerten, line 42: Make this sentence passive, too.
The `FAST_WAVES` option needs to be enabled...

include/crpropa/magneticField/turbulentField/SimpleGridTurbulence.h
```diff
@@ -0,0 +1,115 @@
...
+ @class SimpleTurbulenceSpectrum
+ @brief Defines the energy spectrum of simple power-law turbulence
+ */
+class SimpleTurbulenceSpectrum : public TurbulenceSpectrum {
+  public:
+	/**
+	 @param Brms         root mean square field strength for generated field
+	 @param lMin	 Minimum physical scale of the turbulence
+	 @param lMax	 Maximum physical scale of the turbulence
+	 @param lBendover	   the bend-over scale
```
lukasmerten, line 27: lBendover is not available in SimpleTurbulenceSpectrum, is it?

include/crpropa/magneticField/turbulentField/SimpleGridTurbulence.h
```diff
@@ -0,0 +1,115 @@
...
+	/**
+	General energy spectrum for synthetic turbulence models
+	*/
+	double energySpectrum(double k) const {
+		return std::pow(k, -getSindex() - 2);
+	}
+
+	/**
+	    @brief       compute the magnetic field coherence length according to
+	   the formula in  Harari et Al JHEP03(2002)045
```
lukasmerten, line 45: Harari et al.

include/crpropa/magneticField/turbulentField/SimpleGridTurbulence.h
```diff
@@ -0,0 +1,115 @@
...
+
+// Compatibility with old functions from GridTurbulence:
+
+/** Analytically calculate the correlation length of the simple model turbulent
+ * field */
+inline double turbulentCorrelationLength(double lMin, double lMax,
+                                         double alpha = -11 / 3.) {
+	KISS_LOG_WARNING
+	    << "turbulentCorrelationLength is deprecated and will be "
+	       "removed in the future. Replace it with an appropriate "
```
lukasmerten, line 87: I would rather write here "more appropriate"

include/crpropa/magneticField/turbulentField/SimpleGridTurbulence.h
```diff
@@ -0,0 +1,115 @@
...
+ @param alpha	Power law index of <B^2(k)> ~ k^alpha (alpha = -11/3 corresponds
+ to a Kolmogorov spectrum)
+ @param Brms	RMS field strength
+ @param seed	Random seed
+ */
+inline void initTurbulence(ref_ptr<Grid3f> grid, double Brms, double lMin,
+                           double lMax, double alpha = -11 / 3., int seed = 0) {
+	KISS_LOG_WARNING
+	    << "initTurbulence is deprecated and will be removed in the future. "
+	       "Replace it with an appropriate turbulent field model instance.";
```
lukasmerten, line 106: It will be replaced with a more appropriate...

include/crpropa/magneticField/turbulentField/TurbulentField.h
```diff
@@ -0,0 +1,119 @@
...
+ */
+
+/**
+ @class TurbulenceSpectrum
+ @brief Defines the energy spectrum of turbulence
+ */
+class TurbulenceSpectrum : public Referenced {
+  private:
+	const double Brms; /**< Brms value of the turbulent field (normalization) */
+	const double sIndex; /**< Spectral index for the inertial range, for example
```
lukasmerten, line 20: Shall we clarify at some point why we are sometime referring to a Kolmogorov spectrum with an index of 5/3 and sometime with 11/3? 

I know the difference come from the additional k^2 when you do the integral for an isotropic turbulence in spherical coordiantes. I am just not sure if we should include a comment somewhere as long as we have these two implementations in parallel.

include/crpropa/magneticField/turbulentField/TurbulentField.h
```diff
@@ -0,0 +1,119 @@
...
+	   @param Brms         root mean square field strength for generated field
+	 @param lMin	 Minimum physical scale of the turbulence
+	 @param lMax	 Maximum physical scale of the turbulence
+	   @param lBendover	   the bend-over scale
+	 @param sindex	 Spectral index of the energy spectrum in the inertial
+	 range
+	 @param qindex	 Spectral index of the energy spectrum in the energy
+	 range
+	                       Usually, you'd want to use s=5./3.
+	                       and q=4 here, which will be comparable
```
lukasmerten, line 48: Do we need this comment as the values are set as the default anyway?

include/crpropa/magneticField/turbulentField/PlaneWaveTurbulence.h
antoniusf: I think that this might be a bit difficult say exactly. We do have some data on this in the paper, but in general this probably depends on a lot of factors – not just the number of wavemodes, but it will of probably also depend on which other modules are running and how much time they take, and we haven't done any testing on module configurations other than just PropagationCK.

include/crpropa/magneticField/turbulentField/PlaneWaveTurbulence.h
```diff
@@ -0,0 +1,110 @@
...
+
+	// data for FAST_WAVES
+	int avx_Nm;
+	int align_offset;
+	std::vector<double> avx_data;
+	// the following are index bases into the avx_data array.
+	// since each subarray has avx_Nm elements, the start offset
+	// of each subarray can be computed by multiplying the two,
+	// and then adding on the alignment offset.
+	// iAxi is a combined array containing the product of A * xi
```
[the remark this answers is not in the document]
antoniusf: Yes, this should be Ak * xi, thanks!

src/magneticField/turbulentField/GridTurbulence.cpp
```diff
@@ -0,0 +1,198 @@
...
+				b.z = Bz[i];
+			}
+		}
+	}
+}
+
+
+} // namespace crpropa
+
+#endif // CRPROPA_HAVE_FFTW3F
```
lukasmerten, line 198: You have separated the old code into logical functions, too? I like it.

include/crpropa/magneticField/turbulentField/PlaneWaveTurbulence.h
antoniusf: I think I did this to avoid doubling the "needs to be enabled" construction, but I can understand that switching to active voice might seem weird or clunky, too.

src/magneticField/turbulentField/PlaneWaveTurbulence.cpp
```diff
@@ -0,0 +1,403 @@
...
+	//  CRPropa.");
+	//}
+#endif
+
+	if (Nm <= 1) {
+		throw std::runtime_error(
+		    "PlaneWaveTurbulence: Nm <= 1. We need at least two wavemodes in "
+		    "order "
+		    "to "
+		    "generate the k distribution properly, and besides -- *what are "
```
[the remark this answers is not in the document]
antoniusf: I think passive voice sounds like a good idea. And I agree, the last sentence is not appropriate in this context. I'm sorry I forgot to take it out.

src/magneticField/turbulentField/PlaneWaveTurbulence.cpp
```diff
@@ -0,0 +1,403 @@
...
+		// see inf's, but maybe we want that, so that things dont just fail
+		// silently...
+
+		// * end of argument reduction
+		// *******
+
+		// ******
+		// * evaluate the cosine using a polynomial approximation
+		// * the coefficients for this were generated using sleefs gencoef.c
+		// * I have no idea what I'm doing, so these coefficients are probably
```
lukasmerten, line 366: This should be rephrased before going into the master branch.

test/testCore.cpp
```diff
@@ -435,15 +435,36 @@ TEST(ScalarGrid, SimpleTest) {
...
+	Grid1f grid(properties);
 
-	ScalarGrid grid(origin, Nx, Ny, Nz, spacing);
+	EXPECT_EQ(spacing, grid.getSpacing());
+
+	// Test index handling: get position of grid point (1, 7, 6)
+	size_t some_index = 1 * Ny * Nz + 7 * Nz + 6;
+	Vector3d some_grid_point = origin + Vector3d(1, 7, 6) * spacing + spacing / 2.;
+
+	grid.get(1, 7, 6) = 12;
```
lukasmerten, line 454: The Grid1f.grid function acts also as a setter/mutator right?

Looks a bit unfamiliar but if it works it's probably fine here.

lukasmerten:
I just went through the files mainly checking for consistent and useful comments and obvious things that I would not understand. So far it looked overall good. See my above comments on smaller things. 

I would like to try out a few things but after that I would be happy to approve it. 

Sorry again for the slow reviewing speed at the moment.

test/testCore.cpp
antoniusf: We could consider renaming the function to .index()? Maybe this would make it clearer that it's not just a getter. I don't think it's that important though either.

include/crpropa/magneticField/turbulentField/PlaneWaveTurbulence.h
lukasmerten: What about: "Furthermore, depending on the exact setting, the implementation is somewhat slower (see Schlegel et al. <link to the paper> for details)."

test/testCore.cpp
lukasmerten: I don't a strong opinion on that. We could follow your suggestion or introduce a new function .set() and remove the setting capabilities from .get(), if that is possible.

include/crpropa/magneticField/turbulentField/PlaneWaveTurbulence.h
antoniusf: That sounds good to me.

adundovi:
@lukasmerten can we continue with this PR? I've been extensively testing the code and the grid part works well. Soon I'll also take a deeper look into the plane-wave part regarding physics, but I think that the PR review can continue. I'll address also your comments above.

lukasmerten:
From my point of view, it looks ok now. Do you want to wait with the merging until you have tested the plane wave approach?

test/testCore.cpp
adundovi: I'll leave this for later PR since it affects the standard interface I think.

src/magneticField/turbulentField/PlaneWaveTurbulence.cpp
adundovi: Yes. Changed.

src/magneticField/turbulentField/PlaneWaveTurbulence.cpp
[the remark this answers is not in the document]
adundovi: Changed (I put the shortest possible version: "Specify at least two wavemodes in ...").

src/magneticField/turbulentField/GridTurbulence.cpp
adundovi: Yes. Thanks! :-)

include/crpropa/magneticField/turbulentField/TurbulentField.h
adundovi: Right. Removed.

include/crpropa/magneticField/turbulentField/TurbulentField.h
adundovi: Right. I've added here the following:
```
/**< Spectral index for the inertial range, for example
                          s=5/3 for Kolmogorov spectrum; in some parts of the code this
                          parameter is referred by alpha which is the total 3D isotropic
                          spectrum with additional k^2 and the minus sign, e.g.,
                          for Kolmogorov: alpha = -(s + 2) */
```

include/crpropa/magneticField/turbulentField/SimpleGridTurbulence.h
adundovi: Changed.

include/crpropa/magneticField/turbulentField/SimpleGridTurbulence.h
adundovi: Changed.

include/crpropa/magneticField/turbulentField/SimpleGridTurbulence.h
adundovi: Fixed.

include/crpropa/magneticField/turbulentField/SimpleGridTurbulence.h
adundovi: True. What I named the simple turbulence spectrum is just the inertial part of turbulence, i.e. a power-law with cut-offs at k_min and k_max.

include/crpropa/magneticField/turbulentField/PlaneWaveTurbulence.h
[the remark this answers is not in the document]
adundovi: Changed.

include/crpropa/magneticField/turbulentField/PlaneWaveTurbulence.h
adundovi: Changed.

include/crpropa/magneticField/turbulentField/PlaneWaveTurbulence.h
adundovi: Changed.

include/crpropa/magneticField/turbulentField/PlaneWaveTurbulence.h
adundovi: I've changed everything to uppercase to be consistent with the rest of the source code.

adundovi:
> From my point of view, it looks ok now. Do you want to wait with the merging until you have tested the plane wave approach?

I'm not sure. I would probably go with the merge now because the new feature technically works and has been extensively tested by the original authors already ( @antoniusf and @tfitoussi ). I would leave more testing and updating docs for later PRs.

include/crpropa/magneticField/turbulentField/TurbulentField.h
lukasmerten: That's a good explanation.

test/testCore.cpp
lukasmerten: Fine for me.

CHANGELOG.md
```diff
@@ -12,16 +16,30 @@
...
   consequently, the new photon fields implementation follows the same logic as
   of the other modules making it is easier to introduce custom ones
+* New class-based interface for turbulent fields introduced
+* New turbulence modules implemented:
+  - GridTurbulence (with the bendover scale) which should in general be used
+    instead of initTurbulence (before, it was implmeneted as
+    initTurbulenceWithBendover);
+  - SimpleGridTurbulence which provides the exact field as initTurbulence;
+  - HelicalGridTurbulence which provides the exact field as initHelicalTurbulence
+  - PlaneWaveTurbulence a new algorthm based on Giacalone & Jokipii, 1999 and
```
lukasmerten, line 26: just a small typo: --> algorithm

CHANGELOG.md
```diff
@@ -12,16 +16,30 @@
...
+    Tautz & Dosch, 2013.
 
 ### Interface change:
 
 * ParticleCollector::getAll() -> ParticleCollector::getContainer()
 * Photon fields are no longer items of the PhotonField enumerator but independent
   classes that shares the same interface, so instead of `CMB` one should use `CMB()`,
```
lukasmerten, line 33: @adundovi can you please fix that typo here in the same run, too: shares -- share, since photon fields are plural

lukasmerten:
When the last two small typos I found are fixed I am happy to approve the PR. Updating and testing in a new separate PR is fine for me, too.

adundovi:
Thanks Lukas. I've fixed the typos.

lukasmerten (approved):

PR #289, opened 2020-05-02
Minimum energy for specific types of particles

rafaelab:
- Allows user-written plugins to access the googletest suite by installing them.

- Provides functionalities to apply a minimum energy break condition for specific types of particles (useful for CR-induced cascades, for example).

The name MinimumEnergyParticle is probably not the best, but I couldn't think of anything better than this... or MinimumEnergySpecificTypesOfParticles.

### DIFF
--- a/include/crpropa/module/Observer.h
+++ b/include/crpropa/module/Observer.h
@@ -205,6 +205,19 @@ public:
 };
 
 /**
+ @class ObserverCustomVeto
+ @brief Custom veto for user-defined particle types.
+ */
+class ObserverCustomVeto: public ObserverFeature {
+private:
+	int vetoParticleId;
+public:
+	ObserverCustomVeto(int pId);
+	DetectionState checkDetection(Candidate *candidate) const;
+	std::string getDescription() const;
+};
+
+/**
  @class ObserverTimeEvolution
  @brief Observes the time evolution of the candidates (phase-space elements)
  This observer is very useful if the time evolution of the particle density is needed. It detects all candidates in regular timeintervals and limits the nextStep of candidates to prevent overshooting of detection intervals.
--- a/include/crpropa/module/Observer.h
+++ b/include/crpropa/module/Observer.h
@@ -205,14 +205,14 @@ public:
 };
 
 /**
- @class ObserverCustomVeto
+ @class ObserverParticleIdVeto
  @brief Custom veto for user-defined particle types.
  */
-class ObserverCustomVeto: public ObserverFeature {
+class ObserverParticleIdVeto: public ObserverFeature {
 private:
 	int vetoParticleId;
 public:
-	ObserverCustomVeto(int pId);
+	ObserverParticleIdVeto(int pId);
 	DetectionState checkDetection(Candidate *candidate) const;
 	std::string getDescription() const;
 };
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -81,8 +81,6 @@ option(ENABLE_TESTING "Build tests and enable test target" ON)
 if(ENABLE_TESTING)
 	include_directories(libs/gtest/include)
 	add_subdirectory(libs/gtest)
-	list(APPEND CRPROPA_EXTRA_LIBRARIES gtest)
-	list(APPEND CRPROPA_EXTRA_INCLUDES libs/gtest/include)
 	if(APPLE)
 		set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -DGTEST_USE_OWN_TR1_TUPLE=1")
 	endif(APPLE)
@@ -484,7 +482,6 @@ install(DIRECTORY ${CMAKE_BINARY_DIR}/include/ DESTINATION include FILES_MATCHIN
 install(DIRECTORY ${CMAKE_BINARY_DIR}/data/ DESTINATION share/crpropa/ PATTERN ".git" EXCLUDE)
 
 install(DIRECTORY libs/kiss/include/ DESTINATION include)
-install(DIRECTORY libs/gtest/include/ DESTINATION include)
 
 # ------------------------------------------------------------------
 # Documentation
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -81,6 +81,8 @@ option(ENABLE_TESTING "Build tests and enable test target" ON)
 if(ENABLE_TESTING)
 	include_directories(libs/gtest/include)
 	add_subdirectory(libs/gtest)
+	list(APPEND CRPROPA_EXTRA_LIBRARIES gtest)
+	list(APPEND CRPROPA_EXTRA_INCLUDES libs/gtest/include)
 	if(APPLE)
 		set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -DGTEST_USE_OWN_TR1_TUPLE=1")
 	endif(APPLE)
@@ -482,6 +484,7 @@ install(DIRECTORY ${CMAKE_BINARY_DIR}/include/ DESTINATION include FILES_MATCHIN
 install(DIRECTORY ${CMAKE_BINARY_DIR}/data/ DESTINATION share/crpropa/ PATTERN ".git" EXCLUDE)
 
 install(DIRECTORY libs/kiss/include/ DESTINATION include)
+install(DIRECTORY libs/gtest/include/ DESTINATION include)
 
 # ------------------------------------------------------------------
 # Documentation
--- a/include/crpropa/module/BreakCondition.h
+++ b/include/crpropa/module/BreakCondition.h
@@ -102,6 +102,28 @@ public:
 };
 
 /**
+ @class MinimumEnergyParticle
+ @brief Deactivates the candidate below a minimum energy for specific particle Ids.
+
+ This modules deactivates the candidate below a given minimum energy for specific particle types.
+ In that case the property ("Deactivated", module::description) is set.
+ Care should be taken when calling this module multiple times (use consistent global energy).
+ */
+class MinimumEnergyParticle: public AbstractCondition {
+	std::vector<double> minEnergies;
+	std::vector<int> particleIds;
+	double minEnergyGlobal;
+public:
+	MinimumEnergyParticle(double minEnergyGlobal = 0);
+	void setMinimumGlobalEnergy(double energy);
+	double getMinimumGlobalEnergy() const;
+	void add(int id, double energy);
+	std::string getDescription() const;
+	void process(Candidate *candidate) const;
+};
+
+
+/**
  @class DetectionLength
  @brief Detects the candidate at a given trajectoryLength
  
--- a/include/crpropa/module/BreakCondition.h
+++ b/include/crpropa/module/BreakCondition.h
@@ -102,21 +102,21 @@ public:
 };
 
 /**
- @class MinimumEnergyParticle
+ @class MinimumEnergyPerParticleId
  @brief Deactivates the candidate below a minimum energy for specific particle Ids.
 
  This modules deactivates the candidate below a given minimum energy for specific particle types.
  In that case the property ("Deactivated", module::description) is set.
- Care should be taken when calling this module multiple times (use consistent global energy).
+ All particles whose minimum energies are not specified follow the more general minEnergyOthers condition.
  */
-class MinimumEnergyParticle: public AbstractCondition {
+class MinimumEnergyPerParticleId: public AbstractCondition {
 	std::vector<double> minEnergies;
 	std::vector<int> particleIds;
-	double minEnergyGlobal;
+	double minEnergyOthers;
 public:
-	MinimumEnergyParticle(double minEnergyGlobal = 0);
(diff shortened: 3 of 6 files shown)


adundovi:
@rafaelab Can you please clarify what for is the first point? Google tests for plugins? I think that should be specify separately in a plugin's CMakeList.txt.

Regarding the second part. I would call it `MinimumEnergyPerParticle` or `MinimumEnergyPerParticleID` (so that is immediately recognized which arguments it takes).
Moreover, I would rename `minEnergyGlobal` to `minEnergyForOthers` since `Global` somehow implies that this criteria obligates all IDs, however, in your implementation those which are being checked in the first part are skipped by `return` so only those non-specified reach the last if: 
```c++
	if (c->current.getEnergy() < minEnergyGlobal)
		reject(c);
```
If that was your intention? If you really want "Global" behaviour, then you can drop the last "if" completely and use the old `MinimumEnergy` condition as an additional module in the ModuleList to achieve the same functionality without duplicating the code.

rafaelab:
> @rafaelab Can you please clarify what for is the first point? Google tests for plugins? I think that should be specify separately in a plugin's CMakeList.txt.

Since google tests are already included in CRPropa, it makes more sense to expose it and make it accessible to external plugins through CRPropa headers/libraries. This way, the developer of a plugin doesn't have to provide the package.

At the moment, Google test generates CRPropa tests but cannot be accessible externally (like KISS, for example).

rafaelab:
> Regarding` the second part. I would call it MinimumEnergyPerParticle or MinimumEnergyPerParticleID (so that is immediately recognized which arguments it takes).

That is indeed a good suggestion. I will change it to the second. The first sounds confusing (if you have a composite system, it could be confused with energy per nucleon).

Moreover, I would rename minEnergyGlobal to minEnergyForOthers since Global somehow implies that this criteria obligates all IDs, however, in your implementation those which are being checked in the first part are skipped by return so only those non-specified reach the last if:

	if (c->current.getEnergy() < minEnergyGlobal)
		reject(c);

> If that was your intention? If you really want "Global" behaviour, then you can drop the last "if" completely and use the old MinimumEnergy condition as an additional module in the ModuleList to achieve the same functionality without duplicating the code.

My idea was that only this module can be used instead of MinimumEnergy, to avoid confusions. Therefore, particles whose Ids and Emins are specified follow the conditions given, and the others follow a more general minimum energy condition. 
Wouldn't the usage of two similar modules be quite confusing (and possibly lead to errors)? What do you think, @adundovi ?

adundovi:
> Since google tests are already included in CRPropa, it makes more sense to expose it and make it accessible to external plugins through CRPropa headers/libraries. This way, the developer of a plugin doesn't have to provide the package.
> 
> At the moment, Google test generates CRPropa tests but cannot be accessible externally (like KISS, for example).

OK. We can go with your suggestion for now because I think that the real issue is somewhere else: should we include a local copy of google-tests within our repository in the first place or should we just ask users to provide them like we ask for FFTW, muParser, SWIG, etc. if they want to test the code ( @TobiasWinchen )? By providing it locally and by copying its headers to the system, we introduce a bit of mess to a user's system which could cause her/him trouble later, especially if she/he would install something else that requires gtest - there will be collisions or at least versions mismatches. With KISS/Sophia/DINT is different as these are not widely used libraries and it is hard to obtain them, hence makes sense to provide them with our code and there is really a small chance that someone would already have it or need it later. This is my only concern.

rafaelab:
> > Since google tests are already included in CRPropa, it makes more sense to expose it and make it accessible to external plugins through CRPropa headers/libraries. This way, the developer of a plugin doesn't have to provide the package.
> > At the moment, Google test generates CRPropa tests but cannot be accessible externally (like KISS, for example).
> 
> OK. We can go with your suggestion for now because I think that the real issue is somewhere else: should we include a local copy of google-tests within our repository in the first place or should we just ask users to provide them like we ask for FFTW, muParser, SWIG, etc. if they want to test the code ( @TobiasWinchen )? By providing it locally and by copying its headers to the system, we introduce a bit of mess to a user's system which could cause her/him trouble later, especially if she/he would install something else that requires gtest - there will be collisions or at least versions mismatches. With KISS/Sophia/DINT is different as these are not widely used libraries and it is hard to obtain them, hence makes sense to provide them with our code and there is really a small chance that someone would already have it or need it later. This is my only concern.

Good point about conflicts.
Maybe we can have an EXPOSE_GOOGLETESTS flag in CMakeList.txt? People like myself who are writing plugins can then enable it by hand. Otherwise, the default is to keep things as they are.

adundovi:
> My idea was that only this module can be used instead of MinimumEnergy, to avoid confusions. Therefore, particles whose Ids and Emins are specified follow the conditions given, and the others follow a more general minimum energy condition.
Wouldn't the usage of two similar modules be quite confusing (and possibly lead to errors)? What do you think, @adundovi ?

If this is the case then there is no alternative to your suggestion. It is OK. You can freely ignore the second part of my comment: if one would combine the old `MinimumEnergy` with your new module, the old condition will be activated for all particles - not only those which are "Others", but that is not what you want as I understand now, and there is no way to make the old `MinimumEnergy` aware which particle IDs should be skipped...

lukasmerten:
> should we include a local copy of google-tests within our repository in the first place or should we just ask users to provide them like we ask for FFTW, muParser, SWIG, etc. if they want to test the code? 

I would vote for *not* including a local copy of google-tests, following @adundovi remarks.

adundovi:
> Good point about conflicts.
> Maybe we can have an EXPOSE_GOOGLETESTS flag in CMakeLists.txt? People like myself who are writing plugins can then enable it by hand. Otherwise, the default is to keep things as they are.

It's a bit _workaroundish_. :-)

I would rather provide a CMake option inside of a plugin's `CMakeLists.txt` where you can specify the full path to the gtest directory, whether you provide it on your own, or use the one from the CRPropa installation. This would keep system paths clean from our copy of gtest while still enable you or anyone else to use the CRPropa copy. Please try if it works (but it should): in `include_directories` of the plugin's `CMakeLists.txt` add `${GTEST_INCLUDES}` and then provide also `set()` for that variable earlier; finally you use it when calling `cmake .. -DGTEST_INCLUDES=/path/to/your/crpropa/gtest/...` This can be updated in the plugin directory in the main repository too, and we explain this in the doc. What do you think about this possibility?

rafaelab:
> I would rather provide a CMake option inside of a plugin's `CMakeLists.txt` where you can specify the full path to the gtest directory, whether you provide it on your own, or use the one from the CRPropa installation. This would keep system paths clean from our copy of gtest while still enable you or anyone else to use the CRPropa copy. 

But if you think that the sole purpose of the plugin is to add functionalities to CRPropa that we don't want to add to the main code, it seems unncessary to have another copy of it in the plugin.

In terms of usability, I personally think it is better to have google tests in CRPropa, so that the user can promptly check if the installation succeeded. This can only be done if google tests is properly working. Having to download and install yet another library separately adds an unnecessary layer of complexity.

Moreover, from the google tests [README](https://github.com/google/googletest/blob/master/googletest/README.md):
> If you want to use gtest in a project which already uses CMake, then a more robust and flexible approach is to build gtest as part of that project directly. This is done by making the GoogleTest source code available to the main build and adding it using CMake's add_subdirectory() command. This has the significant advantage that the same compiler and linker settings are used between gtest and the rest of your project, so issues associated with using incompatible libraries (eg debug/release), etc. are avoided. 

Therefore, there is a reason to have it together with CRPropa. 
Since it is relatively light, I'm for keeping it in the code.

adundovi:
> But if you think that the sole purpose of the plugin is to add functionalities to CRPropa that we don't want to add to the main CRPropa, it seems unncessary to have another copy of gtests in the plugin.

It is not another copy, it is the same copy that we ship with CRPropa, you just specify the path to it which cannot be found automatically in an elegant way.
 
> In terms of usability, I personally think it is better to have google tests in CRPropa, so that the user can promptly check if the installation succeeded. This can only be done if google tests is properly working. Having to download and install yet another library separately, adds an unnecessary layer of complexity.
>  (...)
> Therefore, there is a reason to have it together with CRPropa.
> Since it is relatively light, I'm for keeping it in the code.

I think you are reading the quote incorrectly: they don't say anything whether you ship googletest within your code or provide it separately, it just says if you have CMake, it is easy to use googletest from it wherever the googletest source is located. See what they suggest in the same README just a paragraph later (thanks for linking it ;-) ):

> Making GoogleTest's source code available to the main build can be done a few different ways:
> * Download the GoogleTest source code manually and place it at a known location. This is the least flexible approach and can make it more difficult to use with continuous integration systems, etc.
>  * Embed the GoogleTest source code as a direct copy in the main project's source tree. This is often the simplest approach, but is also the hardest to keep up to date. Some organizations may not permit this method.
>  * Add GoogleTest as a git submodule or equivalent. This may not always be possible or appropriate. Git submodules, for example, have their own set of advantages and drawbacks.
>  * Use CMake to download GoogleTest as part of the build's configure step. This is just a little more complex, but doesn't have the limitations of the other methods.

As you can see, we use the second method and it is fine as long as we keep it inside of our build directory and don't install it system wide. But if we do, we'll create chaos for other codes relying on gtest (or even for our own external plugins) since there could be multiple versions of gtest available on the system. Maybe we should provide googletests as the last option listed suggests which is both automatic (if one uses CMake), which doesn't have other penalties. However, you would still have to use a manual path for plugins (all of these are not the solution for your particular issue). I'm certainly not happy with having many external dependencies for CRPropa (and with it, I agree with you), yet, if we introduce some system-invasive defaults, we can create more troubles for our users and indirectly for us (as it will be harder to support it). Please, reconsider my suggestion of just providing the full-path to CRPropa's gtest directory when running cmake for the plugin: it is not automatic, but neither it is hard.

adundovi:
I'm going to separate the discussion whether should we ship google-test within our code: I think we should, as it is the simplest option. Downloading automatically with cmake is a bit overkill for our case. However, I would really like to avoid installing the google-test headers system-wide when you hit `make install`, as I can already see many problems emerging. In their [FAQ](https://github.com/doo/googletest/blob/master/wiki/FAQ.wiki#Why_is_it_not_recommended_to_install_a_precompiled_copy_of_Google_Test_for_example_into_usrlocal), you can read why this is not recommended. I think I will certainly have issues with this as I develop another code that uses gtest (also included within the code) and I use it alongside CRPropa within the same environment, thus there will be two gtest available already which will differ in compiler flags... i.e., the clash is guaranteed. :-)

rafaelab:
>  I'm certainly not happy with having many external dependencies for CRPropa (and with it, I agree with you), yet, if we introduce some system-invasive defaults, we can create more troubles for our users and indirectly for us (as it will be harder to support it)

I completely agree with you. Then in the next CRPropa call we should discuss how to proceed with gtest. 


> Please, reconsider my suggestion of just providing the full-path to CRPropa's gtest directory when running cmake for the plugin: it is not automatic, but neither it is hard.

I will try your suggestion, @adundovi. Thanks! 

Now I will remove this commit from this pull request so that you can take a look at the other things.

include/crpropa/module/Observer.h
```diff
@@ -204,6 +204,19 @@ class ObserverElectronVeto: public ObserverFeature {
...
 
+/**
+ @class ObserverCustomVeto
+ @brief Custom veto for user-defined particle types.
+ */
+class ObserverCustomVeto: public ObserverFeature {
+private:
+	int vetoParticleId;
+public:
+	ObserverCustomVeto(int pId);
```
adundovi, line 215: Sorry for missing this one before. Can you please rename it to something more specific, to include `ParticleId` so that it is consistent with other modules? For example, `ObserverParticleIdVeto` or something similar...

rafaelab: Good idea. Done with commit [298bf52](https://github.com/CRPropa/CRPropa3/pull/289/commits/298bf5291fc6457a2cb1a10b22a6c7e8a2043d12)

TobiasWinchen:
As Andrej and Lukas pointed out,  installing gtest can create a mess as we need to check for conflicts with existing installations etc. - As  Gtest is not essential for using CRPropa and just a development tool, we should not install it, at least not by default. If people need it for their own plugins, they can either ship it  or have a local installation on the development machines. On Debian based distributions it is probably in the repositories, so `apt install googletest` should work. However, as long as the default is no-installation and I do not have to support potential issues I would not oppose a flag that installs it.

rafaelab:
> As Andrej and Lukas pointed out, installing gtest can create a mess as we need to check for conflicts with existing installations etc. - As Gtest is not essential for using CRPropa and just a development tool, we should not install it, at least not by default. If people need it for their own plugins, they can either ship it or have a local installation on the development machines. On Debian based distributions it is probably in the repositories, so `apt install googletest` should work. However, as long as the default is no-installation and I do not have to support potential issues I would not oppose a flag that installs it.

Thanks for the comment, Tobi.
The easy solution to have it in the plugin without installing gtests is to just guess its path and link against it during the compilation of the plugin. This is essentially what CRPropa does, the difference being that the gtests in another folder that CMake can not always guess correctly. 
It's not the most convenient solution, but works just fine.

PR #294, opened 2020-05-06
Automatic coverage reports through coveralls.io service

adundovi:
I think that coverage reports are rarely (if ever?) used when people (including CRPropa developers) compile the code offline since it is slowing down the compilation and an additional flag in CMake has to be turned on. Thus, I propose with this PR that we introduce automatic coverage which will be executed by TravisCI and pushed to free-for-free-software service Coveralls.io. This will be integrated with the GitHub workflow and advise all of us that we need to cover the code with tests. Consequently, the quality of the code will be higher.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,9 @@ addons:
         - gcc-6
         - g++-6
         - gfortran-6
+        - gcc-9
+        - g++-9
+        - gfortran-9
         - swig
         - libfftw3-dev
         - libcfitsio-dev
@@ -26,6 +29,7 @@ addons:
         - graphviz
         - pandoc
         - python3-pip
+        - lcov
       sources: &sources
         - ubuntu-toolchain-r-test
 cache:
@@ -40,6 +44,13 @@ matrix:
         - COMPILER_NAME=gcc CXX=g++-6 CC=gcc-6 FC=gfortran-6
         - PYTHON_EXECUTABLE=/usr/bin/python3
         - BUILD_DOC=true
+    - os : linux
+      dist: bionic
+      compiler: gcc
+      env:
+        - COMPILER_NAME=gcc CXX=g++-9 CC=gcc-9 FC=gfortran-9
+        - PYTHON_EXECUTABLE=/usr/bin/python3
+        - COVERALLS=true
     - os : linux
       dist: bionic
       compiler: clang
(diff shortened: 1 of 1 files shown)


lukasmerten:
I think this is an good idea. From my non-expert view, the implementation looks also ok. The coverage report is easy to read and can be understood without any previous knowledge.

adundovi:
I did introduce the same feature to my other project [here](https://coveralls.io/github/cosmicrays/hermes) and it works well.

TobiasWinchen:
I fully support integrating with coveralls.io, but have two questions regarding the implementation:
1) You introduce a 6th build in the build matrix, travis IMO supports only five for free projects. Did you check that this works? If not, can this be integrated in one of the other builds? If not, we can probably skip the python 2.7 / clang build as we now have an osX build. 
2) How does coveralls.io handle builds from pull requests? Can we distinguish between test coverage in master and the PRs?

adundovi:
1. I didn't know about the limitation in concurrent builds for FLOSS projects, but it makes sense. Yes, I thought of suggesting to drop Python 2.7 builds, at least on Linux. Why I introduce a new build? I prefer doing coverage with Python 3 and GCC since I don't know if clang works well with lcov or I need to change parts of the tool chain. Moreover, the existing build with Python 3 + GCC is already overloaded with generating docs.

2. I'm not exactly sure - in my previous case it works for PR without any special configuration. But if something has to be additionally configured, we can do it later.

> 
> PULL REQUEST COMMENTS
> 
> Currently, pull request commenting only works with CI’s that provide a pull request number at build time (Travis, CircleCI, TDDium to name a few). If you have a private repo that you would like to receive pull request comments on, you will need to allow pull access for our Coveralls bot: @coveralls
https://docs.coveralls.io/coveralls-notifications

TobiasWinchen:
We should keep one python 2.7 build as 2.7 is still default version in several long-term supported linux distributions, so probably 2.7 gcc-6 makes most sense:

python 2.7, gcc 6 
python 3 gcc 9  (builds test coverage)
python 3 gcc 6  (builds documentation)
python 3 clang
OSx

adundovi:
> We should keep one python 2.7 build as 2.7 is still default version in several long-term supported linux distributions, so probably 2.7 gcc-6 makes most sense:
> 
> python 2.7, gcc 6
> python 3 gcc 9 (builds test coverage)
> python 3 gcc 6 (builds documentation)
> python 3 clang
> OSx

I agree. I've updated in the newest commit the Travis matrix exactly like you suggested.

adundovi:
OK, let's see.

PR #300, opened 2020-06-26
Basic file versioning of the data archive in CMakeLists.txt

adundovi:
This is just a small change in `CMakeLists.txt` to accommodate PR #286: **data.tar.gz** becomes **data-`date +"%Y-%m-%d"`.tar.gz**. I've put also a copy of the current `data.tar.gz` at our DESY [data download path](https://www.desy.de/~crpropa/data/interaction_data/).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -89,7 +89,8 @@ before_script:
   - mkdir build
   - cd build
   - |
-    DATAFILE=$HOME/crpropa_cache/data.tar.gz
+    CRPROPA_DATAFILE_VER=`grep "CRPROPA_DATAFILE_VER " $TRAVIS_BUILD_DIR/CMakeLists.txt | cut -d \" -f2`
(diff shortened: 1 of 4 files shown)


TobiasWinchen:
Could you also update travis to copy all downloaded data files (cp data-*  ...) to the travis cache?

adundovi:
> Could you also update travis to copy all downloaded data files (cp data-* ...) to the travis cache?

Thanks. I missed that part. I did it now, yet not with the wildcard but with a `grep`-ed string from CMakeLists.txt. I think that the wildcard version would require more complicated bash expressions and can be less predictable. Although, by changing the CMakeLists.txt line with that string could break it.

TobiasWinchen:
The wildcard was just a first idea trying to keep it simple - important is that the correct data file ends up in the cache to avoid useless traffic and speed up the build.

PR #302, opened 2020-07-23
New cmake option which enables/disables building docs

adundovi:
To avoid building docs without explicitly stating that in CMake, here a new CMake option is introduced: BUILD_DOC (by default Off). This should reduce building times for users not interested in generating the docs (but having doxygen installed) and also, the same holds for TravisCI builds for which only one worker should build the docs (not all).


TobiasWinchen (approved):

CHANGELOG.md
```diff
@@ -16,7 +16,16 @@
 * Photon field classes replaced the photon field enumerator,
   consequently, the new photon fields implementation follows the same logic as
   of the other modules making it is easier to introduce custom ones
-* New CMake option: `BUILD_DOC` for building Doxygen & Sphinx docs
```
adundovi: this is missed during the rebase.

PR #303, opened 2020-07-29
Em cascades fix

IevgenVovk:
This PR solves the issue of the insufficient EM cascade generation in CRPropa chain, otherwise observed in comparison with other simulations, e.g. done for [IGMF studies](https://arxiv.org/abs/1101.0932).

The issue seems to be mainly related to the step size in the photons and electrons interactions, which was limited only if an interaction did not take place. Also in some places the appropriate scaling of the secondaries energy with redshift was not applied.


rafaelab:
Hi Ievgen,
Thanks for the PR.
It seems that your pull request has significant overlap with an existing one: https://github.com/CRPropa/CRPropa3/pull/281
Apparently the issues you have fixes are already fixed there. Could you please verify that this is the case?

IevgenVovk:
Hello Rafael,

Perhaps it would be better if you review a different version of this PR, which is [here](https://github.com/CRPropa/CRPropa3/pull/304). I had troubles making checks against the `master` branch. Maybe we can move the discussion there?

I will check the PR you've indicated, thanks for the hint.

PR #304, opened 2020-07-29
Em cascades fix 3.1.6

IevgenVovk:
This PR solves the issue of the insufficient EM cascade generation in CRPropa chain, otherwise observed in comparison with other simulations, e.g. [those](https://arxiv.org/abs/1101.0932) done for IGMF studies. Concretely, CRPropa simulation results in a too hard spectrum, peaked at high energies.

The issue seems to be mainly related to the step size in the photons and electrons interactions, which was limited only if an interaction did not take place. Also in some places the appropriate scaling with redshift of the secondaries energies was not applied.

The code in the PR is intentionally based on CRPropa [v3.1.6](https://github.com/CRPropa/CRPropa3/tree/3.1.6), as the latest version in the master branch seems to have issues initializing the `EMPairProduction` module. The proposed changes do not affect the commits made in between `3.1.6` tag and `master`.

Below is an example script to reproduce the results from fig. 5 (lower panel, central plot - 1ES 0229+200, "maximal" case) from [here](https://arxiv.org/abs/1101.0932). It can be used to check the effectiveness of the proposed changes.

```python
#!/usr/bin/env python
# coding=utf-8

import os
import numpy

os.environ['OMP_NUM_THREADS'] = '4'
import crpropa
from crpropa import *

# =================
# === Main code ===
# =================

nParticles = int(1e2)

setCosmologyParameters(0.7, 0.3)
sourceRedshift = 0.14
spectrumStr = "E^-1*exp(-E/(100*TeV))"
eMin = 100e9*eV
eMax = 1e15*eV

Bamplitude = 0e-12*gauss
gridSize = 10
gridSpacing = 1e3 *kpc

jetAngle    = numpy.radians(0.1)
obsAngle    = numpy.radians(0)
obsPosAngle = numpy.radians(0)

output_file_name = "events_max.dat.gz"

# ------------------------------
# *** Setting up the B-field ***
print("Setting up the B-field grid...")
print(f"B amplitude: {Bamplitude/gauss:.1e} G")
print(f"Grid size: {gridSize:d} bins")
print(f"Grid step: {gridSpacing/Mpc:.1e} Mpc")

vgrid = VectorGrid(Vector3d(0), gridSize, gridSpacing)
Bfield = MagneticFieldGrid(vgrid)

print("Done.")
# ----------------------

# print some properties of our field
print(f'<B^2> = {rmsFieldStrength(vgrid) / gauss : .1e} G')   # RMS
print(f'<|B|> = {meanFieldStrength(vgrid) / gauss : .1e} G')  # mean

# -----------------------------
# *** Setting up the source ***
sourceDistance = redshift2ComovingDistance(sourceRedshift)

jetCone_x = -1*numpy.cos(obsAngle)
jetCone_y = -1*numpy.sin(obsAngle)*numpy.cos(obsPosAngle)
jetCone_z = -1*numpy.sin(obsAngle)*numpy.sin(obsPosAngle)

source = Source()
genericSourceComposition = SourceGenericComposition(eMin, eMax, spectrumStr)
genericSourceComposition.add(22,1)

source.add(SourcePosition(sourceDistance))
source.add(SourceRedshift(sourceRedshift))
source.add(genericSourceComposition)
source.add(SourceEmissionCone(Vector3d(jetCone_x, jetCone_y, jetCone_z), jetAngle))
print(source)
# -----------------------------

# add an observer
observer = Observer()
observer.add( ObserverSmallSphere(Vector3d(0), 1*Mpc) )

output = TextOutput(output_file_name)
output.disableAll()
output.enable(Output.CurrentIdColumn)
output.enable(Output.CreatedIdColumn)
output.enable(Output.CurrentEnergyColumn)
output.enable(Output.CurrentPositionColumn)
output.enable(Output.CurrentDirectionColumn)

observer.onDetection(output)

simulation = ModuleList()
simulation.add(PropagationCK(Bfield, 1e-4, 1*kpc, 10*Mpc))
simulation.add(Redshift())
simulation.add(EMPairProduction(CMB, True))
simulation.add(EMPairProduction(IRB_Franceschini08, True))
simulation.add(EMInverseComptonScattering(CMB, True))
simulation.add(EMInverseComptonScattering(IRB_Franceschini08, True))
simulation.add(MinimumEnergy(1e8*eV))
simulation.add(MaximumTrajectoryLength(1*Gpc))
simulation.add(observer)
simulation.setShowProgress(True)

print(simulation)

simulation.run(source, nParticles, True, True)

output.close()
```

### DIFF
--- a/src/module/EMDoublePairProduction.cpp
+++ b/src/module/EMDoublePairProduction.cpp
@@ -63,14 +63,16 @@ void EMDoublePairProduction::performInteraction(Candidate *candidate) const {
 	// Use assumption of Lee 96 arXiv:9604098
 	// Energy is equally shared between one e+e- pair, but take mass of second e+e- pair into account.
 	// This approximation has been shown to be valid within -1.5%.
-	double E = candidate->current.getEnergy();
+	// Scale the particle energy instead of background photons
+	double z = candidate->getRedshift();
+	double E = candidate->current.getEnergy() * (1 + z);
 	double Ee = (E - 2 * mass_electron * c_squared) / 2;
 
 	Random &random = Random::instance();
 	Vector3d pos = random.randomInterpolatedPosition(candidate->previous.getPosition(), candidate->current.getPosition());
 
-	candidate->addSecondary( 11, Ee, pos);
-	candidate->addSecondary(-11, Ee, pos);
+	candidate->addSecondary( 11, Ee / (1 + z), pos);
+	candidate->addSecondary(-11, Ee / (1 + z), pos);
 }
 
 void EMDoublePairProduction::process(Candidate *candidate) const {
--- a/src/module/EMDoublePairProduction.cpp
+++ b/src/module/EMDoublePairProduction.cpp
@@ -97,8 +97,7 @@ void EMDoublePairProduction::process(Candidate *candidate) const {
 	double randDistance = -log(random.rand()) / rate;
 	if (candidate->getCurrentStep() > randDistance)
 		performInteraction(candidate);
-	else
-		candidate->limitNextStep(limit / rate);
+	candidate->limitNextStep(limit / rate);
 }
 
 } // namespace crpropa
--- a/src/module/EMInverseComptonScattering.cpp
+++ b/src/module/EMInverseComptonScattering.cpp
@@ -67,7 +67,7 @@ void EMInverseComptonScattering::initCumulativeRate(std::string filename) {
 	tabE.clear();
 	tabs.clear();
 	tabCDF.clear();
-	
+
 	// skip header
 	while (infile.peek() == '#')
 		infile.ignore(std::numeric_limits < std::streamsize > ::max(), '\n');
@@ -216,8 +216,7 @@ void EMInverseComptonScattering::process(Candidate *candidate) const {
 	double randDistance = -log(random.rand()) / rate;
 	if (candidate->getCurrentStep() > randDistance)
 		performInteraction(candidate);
-	else
-		candidate->limitNextStep(limit / rate);
+	candidate->limitNextStep(limit / rate);
 }
 
 } // namespace crpropa
--- a/src/module/EMPairProduction.cpp
+++ b/src/module/EMPairProduction.cpp
@@ -65,7 +65,7 @@ void EMPairProduction::initCumulativeRate(std::string filename) {
 	tabE.clear();
 	tabs.clear();
 	tabCDF.clear();
-	
+
 	// skip header
 	while (infile.peek() == '#')
 		infile.ignore(std::numeric_limits < std::streamsize > ::max(), '\n');
@@ -219,8 +219,7 @@ void EMPairProduction::process(Candidate *candidate) const {
 	double randDistance = -log(random.rand()) / rate;
 	if (candidate->getCurrentStep() > randDistance)
 		performInteraction(candidate);
-	else
-		candidate->limitNextStep(limit / rate);
+	candidate->limitNextStep(limit / rate);
 }
 
 } // namespace crpropa
--- a/src/module/EMTripletPairProduction.cpp
+++ b/src/module/EMTripletPairProduction.cpp
@@ -149,8 +149,7 @@ void EMTripletPairProduction::process(Candidate *candidate) const {
 	double randDistance = -log(random.rand()) / rate;
 	if (candidate->getCurrentStep() > randDistance)
 		performInteraction(candidate);
-	else
-		candidate->limitNextStep(limit / rate);
+	candidate->limitNextStep(limit / rate);
 }
 
 } // namespace crpropa
--- a/src/module/EMTripletPairProduction.cpp
+++ b/src/module/EMTripletPairProduction.cpp
@@ -67,7 +67,7 @@ void EMTripletPairProduction::initCumulativeRate(std::string filename) {
 	tabE.clear();
 	tabs.clear();
 	tabCDF.clear();
-	
+
 	// skip header
 	while (infile.peek() == '#')
 		infile.ignore(std::numeric_limits < std::streamsize > ::max(), '\n');
@@ -118,8 +118,8 @@ void EMTripletPairProduction::performInteraction(Candidate *candidate) const {
 
 	if (haveElectrons) {
 		Vector3d pos = random.randomInterpolatedPosition(candidate->previous.getPosition(), candidate->current.getPosition());
-		candidate->addSecondary( 11, Epp, pos);
-		candidate->addSecondary(-11, Epp, pos);
+		candidate->addSecondary( 11, Epp / (1 + z), pos);
+		candidate->addSecondary(-11, Epp / (1 + z), pos);
 	}
 
 	// update the primary particle energy; do this after adding the secondaries to correctly set the secondaries parent


IevgenVovk:
Checking this PR against #281 , I see that the logic of applying step limit is somewhat different. 

Taking `EMPairProduction::process(Candidate *candidate)` as an example, in #281 `candidate->limitNextStep(limit / rate)` is not applied if `performInteraction(candidate)` is called. Such requirement caused the insufficient EM cascade generation in my tests, thus in this PR the step limitation is applied also if `performInteraction(candidate)` is invoked.

IevgenVovk:
In terms of redshift corrections this PR and #281 overlap.

rafaelab:
> Checking this PR against #281 , I see that the logic of applying step limit is somewhat different.
> 
> Taking `EMPairProduction::process(Candidate *candidate)` as an example, in #281 `candidate->limitNextStep(limit / rate)` is not applied if `performInteraction(candidate)` is called. Such requirement caused the insufficient EM cascade generation in my tests, thus in this PR the step limitation is applied also if `performInteraction(candidate)` is invoked.

I also found that problem. It was fixed using a very small minimum step size. You can also play with 'limit' to fine-tune what you consider a good step size.
I am not sure that we want to decrease the step size at *every* single step. This might compromise performance.
Sometimes the large step is enough and the next step doesn't need to be decreased.
I need to run some tests to check how this impacts the performance and if the same results can't be achieved with a combination of step size + limit. I will get back to you about this next week.

IevgenVovk:
Thanks for the hints. I've made some quick checks with v3.1.6

- setting the minimal step size to a small value (1e-3 pc) does not seem solve the issue (I've used `PropagationCK(Bfield, 1e-10, 0.001*pc, 10*Mpc)`);
- limiting the step to 1e-2 in all interactions involved indeed helps at a cost of ~6-10x longer calculation, compared to this PR.

So the step size chosen by the code seems to be indeed the culprit - it seems to be too large by default. Based on these steps, it shouldn't be allowed to be larger than `~ limit / rate`.

Can you think of a more elegant solution to limit this step size?

rafaelab:
Thanks for the tests!

Maybe we should indeed change the default value in CRPropa.

I have given it some thought some time ago, and I couldn't find a way to evade this problem. By using a small enough maxStep the problem was gone (also with a 0.1 limit). But this is extremely time-demanding. So I considered having a particle-dependent PropagationCK (different step sizes for different types of particle), but never tested it. This may be a good solution...

I personally have been using a 1% limit for some time (what really matters is limit * maxStep). I have checked it against Elmag and the results match quite well. You might also want to decrease the tolerance of PropagactionCK to, e.g., 1e-3 or 1e-2, to get some performance improvements.

I believe that with the thinning implemented in my pull request, one could make up for the poor performance you are pointing.
You may want to check out my branch (#281 ).

IevgenVovk:
IMHO, it would be right if the simulation gives an accurate result with default settings without much fine-tuning. I.e. if the right step length is between `minStep` and `maxStep` values - it should be taken.

Substantially reducing the `limit` or `maxStep` values is an extra step that is completely unexpected by the user. This is especially true given that without this even reducing the tolerance value down to 1e-20 does not change the results, yielding a false impression of everything's being fine.

Again IMHO, thinning is a somewhat different thing, which should only be necessary if other optimizations fail. A proper selection of the calculation step - combined with thinning - would be the best approach in terms of performance and accuracy.

rafaelab:
Thank you for the suggestion. I will investigate which parameters can be used without sacrificing the physics too much.
But I doubt there is much we can do in terms of default values.

Because we use a single PropagationCK module for everything, from TeV electrons to UHECRs, the step sizes vary immensely.  Some people use CRPropa for UHECRs (we can use steps between 0.1 Mpc and 10 Mpc and results converge), you are using it for TeV photons (both minStep and maxStep should be tiny). There is really no way to set a default value that covers all cases efficiently.

As I commented, we cannot really change maxStep. But we could play with limit. Nevertheless, at higher energies we don't want this to be small, otherwise we will literally need years to run a single simulation. Therefore, the question is, again, what the default values should be, given the different communities we are reaching. For UHE photons the current settings seem to be satisfactory. 
Maybe the best solution is to provide better documentation for the users.

> Again IMHO, thinning is a somewhat different thing, which should only be necessary if other optimizations fail. A proper selection of the calculation step - combined with thinning - would be the best approach in terms of performance and accuracy.

To determine whether an optimisation fails or not, we need to know how long someone is willing to wait for a simulation to run. This changes from person to person. For me, thinning can be used in essentially all cases. However, the thinning parameter does need to be carefully selected.
Furthermore, for a large enough number of events, thinning should really statistically converge to the actual distribution, which completely justifies its use. The real question is then how to define "large enough number of events".

IevgenVovk:
> Because we use a single PropagationCK module for everything, from TeV electrons to UHECRs, the step sizes vary immensely. Some people use CRPropa for UHECRs (we can use steps between 0.1 Mpc and 10 Mpc and results converge), you are using it for TeV photons (both minStep and maxStep should be tiny). There is really no way to set a default value that covers all cases efficiently.

But probably this is exactly the point - why can't we set the default value based on the energy? This is what this PR is about.

IevgenVovk:
Leaving the dynamic step assignment aside for a second, I actually do not understand from the code why the simulation returns wrong results by default:
- a candidate is initialized with `nextStep=0` (`Candidate.cpp`);
- its value does not seem to be modified by the source (`Source.cpp`);
- `PropagationCK` should clip it to [minStep; maxStep] range - i.e. to minStep (which is very small in my tests - have tried even `1*cm`).

..Which yields the wrong cascade prediction.

However limiting `nextStep` from the top (this PR) solves the issue, indicating that the `nextStep` value becomes large somewhere in the code. 

If this is right, the `candidate->getCurrentStep() > randDistance` condition (e.g. in `EMInverseComptonScattering::process()`) is always true, meaning that a particle would interact at every step, which means rarely (step size is large), with no change to adjust it (only done if the above condition is false). Rarefied interactions mean less secondary emission, which explains the too hard cascade spectra seen with v3.1.6 - the secondary part of the cascade never develops.

Should the `nextStep` properly be set to `minStep` value, the simulation should have yielded the right cascade spectrum, albeit in (possibly) longer time (compared to this PR).

However setting it the `nextStep` to some more interaction-motivated value would be a balance in terms of performance and accuracy.

rafaelab:
I believe the issue is more related to the maxStep than to the minStep.
If you decrease maxStep to, e.g., 10 kpc, things will work.

In principle, a proper selection of limit + maxStep solves this issue without the additional performance overhead of removing `limitNextStep` from `else`. 
The problem is that we don't always need to limit the next step. That's why, while I completely agree with you that this issue should be addressed, I'm not sure the solution proposed is the most efficient.

The question then is: which one is faster, limit+maxStep or limitNextStep always?
This has to be checked before any changes are made. 

I also have a clear idea of an in-between solution that might be optimal. I started with it when I first found the issue you reported, but then I setlled for low maxStep+limit.  
I cannot work on it this week, but by mid next week I can share with you the results of my tests.

IevgenVovk:
Hello Rafael, thanks for the suggestions.

I've made a test with different maxStep values. Here are the results:
 - v3.1.6, maxStep = 100 kpc ..: 3:55 minutes (wrong spectrum)
 - v3.1.6, maxStep = 10 kpc ....: 60:11 minutes (almost right spectrum)
 - this PR .................................: 4:15 minutes (right spectrum)

I've kept `limit` at 0.1; lower values result in even longer run times...

So it looks like this PR offers 15x better performance compared to the simple `maxStep` limit.

rafaelab:
Hi Ievgen,

Sorry for the late reply.

I have checked the proposed implementation. I confirm that it does seem to fix the issues you raised with a good performance. I could not see any significant differences with respect to PR #281 in terms of physics or performance for the cases I've tested. 

However, for inverse Compton (and triplet pair production), I believe the implementation from PR #281 is, at least logically, more adequate. In these cases, tests for interactions will be done until the step is ~0. Therefore, it makes sense to have the code within a do/while block to guarantee it's execution at least once, in order to limit the next step. This follows the same idea of, for example, [Photodisintegration.cpp](https://github.com/CRPropa/CRPropa3/blob/master/src/module/PhotoDisintegration.cpp).

For pair production (and double), if the interaction doesn't ocur, the next step is limited -- exactly like in your implementation. However, if the interaction takes place, there is no reason to limit the next step because the candidate is inactivated in `performInteraction`.

Therefore, the problems arises because the first step is not being limited appropriately. The changes from PR #281 supposedly fixes this issue as well. In terms of logic, it is the same as this PR, except that the if statement is for step < randDistance, applying `limitNextStep`. 
Have you test it?

lukasmerten:
Do we have any new discussion on this topic? Otherwise I will assume that the problem was fixed with #281 and close this PR next week.

rafaelab:
Hi @lukasmerten 
There have been so many changes that I don't know if this PR is still relevant.
Since there have been extensive comparisons done by Korochnikin et al., I assume there is no real bug. However, the current proposal could indeed be an improvement.  
@IevgenVovk: are you using this implementation or are you using the released CRPropa version? Could you please let us know if this PR is still worth being looked at?
Thanks!

lukasmerten:
I close this PR for now. If you, @IevgenVovk, think we are still missing some important parts of this topic in the code, please reopen this PR and we continue the discussion.

IevgenVovk:
Apologies for delayed reaction. I've been indeed using CRPropa 3.2 since it was released; am not ready to say yet if the changes in this PR would be still relevant. Perhaps it is fine to close this MR indeed.

PR #309, opened 2020-09-30
fix segfault in DintPropagation due to size check in wrong place

cozzyd:
I know DINT is probably on its way out, but right now it crashes for me. 

This commit fixes it (the size check is in the wrong place here, and trying to access .back().X1 on a zero-size vector triggers a segfault for me, although I guess it can also fail silently).  The second check is probably extraneous but I fixed it anyway.


TobiasWinchen (approved):

TobiasWinchen:
Thx.

PR #320, opened 2020-12-04
Improve comments for the plane wave turbulence code

antoniusf:
Hi, so recently I was going through the implementation for PlaneWaveTurbulence again, and I noticed that some of the comments had become outdated, or had room for improvement otherwise. I just went through and improved some of them a bit, and also deleted some code that was no longer used.

I'm not done with this yet, but I wanted to open a PR to let you know that I'm looking at this, and to give you the opportunity to add your own suggestions. (Or to tell me that I shouldn't be working on this because it's not that important.)

Anyway, feel free to chime in if you want. I'd like to keep this PR small (there might be more things to discuss but I'd prefer keeping this one focused on just the comments), since I'm somewhat busy right now and my energy for this is limited. In particular, it might be worth taking another look at the documentation, but I'd prefer *not* doing that here so that it can get the time and attention it needs.

### DIFF
--- a/src/magneticField/turbulentField/PlaneWaveTurbulence.cpp
+++ b/src/magneticField/turbulentField/PlaneWaveTurbulence.cpp
@@ -281,17 +281,17 @@ Vector3d PlaneWaveTurbulence::getField(const Vector3d &pos) const {
 		// * Computing the cosine
 		// * Part 1: Argument reduction
 		//
-		//  To compute the cosine, we first realize that due to the periodic
-		//  nature of the function, we only need to model its behavior between
-		//  0 and 2*pi to be able compute the function anywhere. In fact,
-		//  by mirroring the function along the x and y axes, even the range
-		//  between 0 and pi/2 is sufficient for this purpose. In this range,
-		//  the cosine can be evaluated with high precision by using a
-		//  polynomial approximation. Thus, to compute the cosine, the input
-		//  value is first reduced so that it lies within this range. Then,
-		//  the polynomial approximation is evaluated. Finally, if necessary,
-		//  the sing of the result is flipped (mirroring the function along
-		//  the x axis).
+		//  To understand the computation of the cosine, first note that the
+		//  cosine is periodic and we thus only need to model its behavior
+		//  between 0 and 2*pi to be able compute the function anywhere. In
+		//  fact, by mirroring the function along the x and y axes, even the
+		//  range between 0 and pi/2 is sufficient for this purpose. In this
+		//  range, the cosine can be efficiently evaluated with high precision
+		//  by using a polynomial approximation. Thus, to compute the cosine,
+		//  the input value is first reduced so that it lies within this range.
+		//  Then, the polynomial approximation is evaluated. Finally, if
+		//  necessary, the sign of the result is flipped (mirroring the function
+		//  along the x axis).
 		//
 		//  The actual computation is slightly more involved. First, argument
 		//  reduction can be simplified drastically by computing cos(pi*x),
@@ -311,6 +311,16 @@ Vector3d PlaneWaveTurbulence::getField(const Vector3d &pos) const {
 		//  0 and 0.5 would automatically have the same accuracy between -0.5
 		//  and 0.
 		//
+		//  (Even if the rounding mode was different, this would at most
+		//  incur an error of 1 ULP (a relative error of about 1e-15) in the
+		//  value of s*s. Since |s| is smaller than 0.5, |s*s| is smaller than
+		//  0.25, which means that the absolute error is somewhere below 1e-15.
+		//  Given that the current polynomial is already quite imprecise (it's
+		//  only accurate to about 1e-7, or 1 ULP of float precision for numbers
+		//  around 1), and that the slope of the approximation shouldn't exceed
+		//  unity by much (since it models the cosine), this is unlikely to make
+		//  a big difference.)
+		//
 		// step 1: compute round(x), and store it in q
 		__m256d q = _mm256_round_pd(
 		    cos_arg, (_MM_FROUND_TO_NEAREST_INT | _MM_FROUND_NO_EXC));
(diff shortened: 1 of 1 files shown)


src/magneticField/turbulentField/PlaneWaveTurbulence.cpp
```diff
@@ -306,15 +283,23 @@ Vector3d PlaneWaveTurbulence::getField(const Vector3d &pos) const {
...
 		// double precision has decayed far enough at that point that the cosine
 		// would be useless anyway.
+		//
+		// explanation: the mantissa of a double-precision
+		// float has 52 bits (excluding the sign bit). if |q| > 2^51, the msb
+		// of the mantissa has a place value of at least 2^50. (the actual
+		// most significant bit is always 1 and isn't stored.) this means that
+		// the lsb of the mantissa has a place value of at least 2^(-1), or 0.5.
+		// for a cos(pi*x), this corresponds to a quarter of a period (pi/2),
+		// so at this point the cosine isn't really a useful cosine anymore.
```
lukasmerten: "...the **implemented** cosine isn't a really usefule cosine anymore."
Otherwise I do not really get the last sentence.

src/magneticField/turbulentField/PlaneWaveTurbulence.cpp
```diff
@@ -100,17 +74,10 @@ PlaneWaveTurbulence::PlaneWaveTurbulence(const TurbulenceSpectrum &spectrum,
 	KISS_LOG_INFO << "PlaneWaveTurbulence: Using SIMD TD13 implementation"
 	              << std::endl;
```
lukasmerten, line 76: What would happen when the code is compiled with SIMD support but run on a CPU that does not support SIMD?
That's of course nothing that should be done but you never know.

My question is, will this lead to some unexpected behavior or just break the code. In the latter case I would be fine with not replacing the check.

src/magneticField/turbulentField/PlaneWaveTurbulence.cpp
```diff
@@ -169,6 +136,16 @@ PlaneWaveTurbulence::PlaneWaveTurbulence(const TurbulenceSpectrum &spectrum,
 
 		Vector3d kappa =
 		    Vector3d(sintheta * cos(phi), sintheta * sin(phi), costheta);
```
lukasmerten, line 138: I think it is ok to use a slightly different implementation compared to the paper, but we should comment on that shortly. I expect that people digging this deep into the code are able deal with that.

src/magneticField/turbulentField/PlaneWaveTurbulence.cpp
antoniusf: Hmm, I've just added a little bit more explanation on this, but I'm not sure how well that works. What I was trying to say is that, fundamentally, a double value that is greater than 2⁵¹ can only be accurate to an absolute error of 0.5. (Does this make sense? I'm kind of struggling with how to say this.) So what this means is that for any implementation of cos(pi*x), if x is larger than 2⁵¹, the implementation could still return the correct value for the given input x – it's just that that x is necessarily rounded to 0.5, which means that the function is only sampled 4 times per cycle, and jumps by +/-1 between consecutive samples.

src/magneticField/turbulentField/PlaneWaveTurbulence.cpp
antoniusf: From my experience (and a brief google search), the program should crash with an error that says something about invalid instructions. I can do some more in-depth research and see if this behavior can be relied upon.

src/magneticField/turbulentField/PlaneWaveTurbulence.cpp
antoniusf: So, I should just make that comment a bit shorter? (Or just remove the explanation for why this discrepancy exists?)

src/magneticField/turbulentField/PlaneWaveTurbulence.cpp
lukasmerten: I think, now I got it. Thanks for explanation.

src/magneticField/turbulentField/PlaneWaveTurbulence.cpp
lukasmerten: I think we can live we such a behavior then. However, additional research is always nice but I wouldn't put as a high priority, when you have other things to do.

src/magneticField/turbulentField/PlaneWaveTurbulence.cpp
lukasmerten: I would probably just delete the last two sentences of that comment block. I like the reference to the GJ99 paper so people can go back and check the difference between the two.

antoniusf: Thanks! Done.

src/magneticField/turbulentField/PlaneWaveTurbulence.cpp
antoniusf: Okay! I've written it down at low priority.

src/magneticField/turbulentField/PlaneWaveTurbulence.cpp
antoniusf: Great! I'm glad I made sense. Do you think that the updated comment is able to convey this as well, or should I maybe put some more thought into that?

antoniusf:
So, I think the comments are in a place that I'm comfortable leaving them in. Feel free to do an additional review, but I haven't changed that much after incorporating Lukas's feedback. Otherwise, from my side, this is ready to be merged.

I'd appreciate if this got merged before the 3.2 release, so that the PlaneWave code gets released in a form that hopefully allows people to understand its internal workings.

This should be the last of my changes to get PlaneWaves release-ready. Thanks for your patience!

src/magneticField/turbulentField/PlaneWaveTurbulence.cpp
```diff
@@ -100,17 +74,10 @@ PlaneWaveTurbulence::PlaneWaveTurbulence(const TurbulenceSpectrum &spectrum,
...
-	// TODO: this is currently uncommented b/c sleef seems to fail to provide
-	// the cpuid function
-	// if (!check_sse()) {
-	//  throw std::runtime_error("TD13Field: This code was compiled with SIMD
-	//  support (SSE1-3), but it is running on a CPU that does not support these
-	//  instructions. Please set USE_SIMD to OFF in CMake and recompile
-	//  CRPropa.");
-	//}
+	// TODO: there used to be a cpuid check here, to see if the cpu running
+	// this code would support SIMD (SEE + AVX). however, the library providing the
```
lukasmerten: [...] (SEE + AVX). **H**owever, [...]

src/magneticField/turbulentField/PlaneWaveTurbulence.cpp
```diff
@@ -169,6 +136,16 @@ PlaneWaveTurbulence::PlaneWaveTurbulence(const TurbulenceSpectrum &spectrum,
...
+		// NOTE: all other variable names match the ones from the TD13 paper.
+		// However, our xi is actually their psi, and their xi isn't used at
+		// all. (Though both can be used for the polarization vector, according
+		// to the paper.) The reason for this discrepancy is that this code
+		// used to be based on the original GJ99 paper, which provided only a
+		// xi vector, and this xi happens to be almost the same as TD13's psi.
+		// Hence, the code kept both the name and the rough shape of the vector,
+		// but is now in disagreement with the TD13 paper. I don't think there's
+		// a good reason to keep it this way, except that I do not want to
+		// be the one who changes it.
```
lukasmerten: Remove the last sentence "I don't think there's a good reason to keep it this way, except that I do not want to be the one who changes it."

src/magneticField/turbulentField/PlaneWaveTurbulence.cpp
```diff
@@ -345,9 +330,11 @@ Vector3d PlaneWaveTurbulence::getField(const Vector3d &pos) const {
 		    invert, _mm256_castsi256_pd(_mm256_set1_epi64x(0x4330000000000001)),
 		    _CMP_EQ_OQ);
 
-		// finally, we need to turn invert and right_invert into masks for the
-		// sign bit on each final double, ie
+		// finally, we need to turn invert into a mask for the
```
lukasmerten: **F**inally, [...]

src/magneticField/turbulentField/PlaneWaveTurbulence.cpp
```diff
@@ -345,9 +330,11 @@ Vector3d PlaneWaveTurbulence::getField(const Vector3d &pos) const {
...
+		// sign bit on each final double, ie:
 		invert = _mm256_and_pd(invert, _mm256_set1_pd(-0.0));
+		// (note that the binary representation of -0.0 is all 0 bits, except
```
lukasmerten, line 381: **N**ote

src/magneticField/turbulentField/PlaneWaveTurbulence.cpp
```diff
@@ -357,8 +344,8 @@ Vector3d PlaneWaveTurbulence::getField(const Vector3d &pos) const {
 		// *******
 
 		// ******
-		// * evaluate the cosine using a polynomial approximation
-		// * the coefficients for this were generated using sleefs gencoef.c
+		// * evaluate the cosine using a polynomial approximation.
```
lukasmerten: **E**valuate

src/magneticField/turbulentField/PlaneWaveTurbulence.cpp
```diff
@@ -357,8 +344,8 @@ Vector3d PlaneWaveTurbulence::getField(const Vector3d &pos) const {
...
+		// * the coefficients for this were generated using sleefs gencoef.c.
```
lukasmerten: **T**he

lukasmerten (changes requested):

antoniusf:
Hi, thanks for the review, I appreciate your time. I'm a bit confused by what github has done here – most of the things in your review are not what the current version of the code looks like. For example, the "however" in the first example is [already fixed](https://github.com/CRPropa/CRPropa3/blob/e75109a20e3ead9bfb4bc9804f895027c0938f5f/src/magneticField/turbulentField/PlaneWaveTurbulence.cpp#L78), and most of the other text samples aren't even in the code anymore, since they've been rewritten completely. I guess github is showing a little "outdated" next to the filename, but I'm not really sure what to do with these? The only one I could fix is "note" -> "Note", which I've done now.

lukasmerten (approved):
When reviewing I had the feeling that it might have been an outdated version.
Anyway, I think everything is fine now and we can proceed to merge this PR.

antoniusf:
Great, thank you.

Issue #321, opened 2020-12-05
TD13/PlaneWaveTurbulence performance


antoniusf:
Hi,

I've just re-run some benchmarks for various magnetic fields, including TD13/PlaneWaveTurbulence, and something came up that I'd forgotten about before the merge. To make things short, there's a compiler flag you can enable that makes the optimized (SIMD) implementation about 25% faster, by my benchmarks.

To go a bit more into detail, there are certain processor instructions that not all processors support, because they are extensions to the core instruction set that have been added over the past few years. The optimized implementation itself uses some of these extensions (most notably, one called [AVX](https://en.wikipedia.org/wiki/Advanced_Vector_Extensions)), which allows it to operate on multiple numbers at the same time (this is also known as SIMD). This extension is not optional (for the optimized version) because I have had to manually tell the compiler which of the precise instructions to use, in order to make it work.

However, there is a second instruction set extension (this one is called [FMA](https://en.wikipedia.org/wiki/FMA_instruction_set)), which is simple enough that the compiler can apply it automatically. All that FMA does is allow the compiler to merge one multiplication and one addition together into a single instruction that is just as fast (in terms of throughput) as either just the multiply or the add. Doing a quick count, this replacement can be done in nine places in the inner TD13 loop, which is a significant portion of the overall amount of instructions in that loop. In other words, this kind of speedup does make sense.

## the data

Benchmarks were obtained using the same method described in [our paper](https://arxiv.org/pdf/1907.09934.pdf) (see page 7 in the preprint). In short, we do a simple test simulation (only PropCK with a few ten thousand particles, a trajectory length limit, and constant step size so all trajectories have the same number of steps), time it using the `time` utility and a best-of-3 system, and then divide the runtime by the total number of steps done to obtain the average amount of time per step. The table below lists these times (all in µs), along with the time needed by the sped-up code relative to the baseline, for different numbers of wavemodes. Note that we would expect the differences to be more pronounced for larger numbers of wavemodes, since for those, the time spent in the interpolation routine increasingly dominates the time spent everywhere else. (Though more involved effects might be at play here, too.)

| modes | t<sub>s</sub>, no FMA [µs] |  t<sub>s</sub>, FMA [µs]  | relative |
| ----- | ------ | ------ | -------- |
|     5 |   0.52 |   0.48 |    93.1% |
|    10 |   0.56 |   0.51 |    90.7% |
|    20 |   0.67 |   0.59 |    87.6% |
[...]

PR #324, opened 2020-12-26
Improve CMake options relating to the PlaneWaveTurbulence optimization

antoniusf:
This changes the CMake option setup relating to the support of certain processor instructions used by the PlaneWaveTurbulence optimized version. There are two concrete improvements:

1. It is now also possible to enable FMA and thus make `PlaneWaveTurbulence.getField` about 1.33 times faster (see #321).
2. There is now an option to tell the compiler to automatically use all extensions supported by the CPU that CRPropa is being built on. This makes life a lot simpler for people who build CRPropa on the same machine that they run their simulations on, since they no longer have to take care of this.

The relevant documentation for PlaneWaveTurbulence has also been updated to reflect these changes.

I have tested that this does what it should on three machines (one with both AVX and FMA, one with AVX only, and one with neither):

1. The optimization is only built in if `FAST_WAVES` is enabled.
2. Setting `SIMD_EXTENSIONS` to `native` will compile the code with all extensions supported by that CPU. When trying to build with `native` on a machine that supports neither AVX nor FMA, and `FAST_WAVES` is enabled, an error is thrown. (Instead of silently not compiling in the optimization.)
3. When building for another machine, it is possible to build with AVX or FMA support on a machine that itself does not support AVX or FMA, by setting `SIMD_EXTENSIONS` to one of the manual selection values.


adundovi (approved):

adundovi:
Great job! Works for me.

antoniusf:
Yay, thanks!

PR #326, opened 2021-01-12
Implementation of the new PolarizedSingleModeMagneticField class

asavelie:
This contribution adds the possibility of defining a magnetic field with a single polarization mode in CRPropa. This includes the more specific circular and linear polarizations, but also the more general elliptical one, as well as a formalism to handle magnetic helicity.


rafaelab:
I believe this pull request is almost ready.
A notebook with an example is the only thing missing.

adundovi:
@asavelie How do we test it quickly? What is the simplest known result one should obtain from this?

src/magneticField/PolarizedSingleModeMagneticField.cpp
```diff
@@ -0,0 +1,124 @@
...
+	sete1(e_1);
+	sete2(e_2);
+}
+
+Vector3d PolarizedSingleModeMagneticField::getField(const Vector3d &pos, const double &sigma) const {
+	Vector3d deltar = pos - r_0;
+
+	Vector3d unitVector_1;
+	if (e_1.getR() == 0)
+		throw std::runtime_error("Vector e_1 cannot be zero.");
```
rafaelab, line 18: For consistency with the rest of CRPropa, exceptions should be handled with kiss:  
KISS_LOG_ERROR, KISS_LOG_WARNING, etc.

include/crpropa/magneticField/PolarizedSingleModeMagneticField.h
```diff
@@ -0,0 +1,67 @@
...
+	@param r_0; // Reference position
+	@param e_1; // First vector spanning the polarization plane 
+	@param e_2; // Second vector spanning the polarization plane
+*/
+	PolarizedSingleModeMagneticField(double B_0, double wavelength, Vector3d r_0, Vector3d e_1, Vector3d e_2);
+
+	Vector3d getField(const Vector3d &pos, const double &sigma) const;
+
+	Vector3d getMaxHelFracField(const Vector3d &pos, const double &fH, const std::string &Bflag) const;
+	Vector3d getOrthogonalMaxHelFracField(const Vector3d &pos, const double &fH, const std::string &Bflag) const;
```
rafaelab, line 45: These names are kind of long (which is fine), but it is hard to guess immediately what 'Frac' means, for example. Isn't there a way to explicitly write these things while keeping the function name relatively short?

src/magneticField/PolarizedSingleModeMagneticField.cpp
```diff
@@ -0,0 +1,124 @@
...
+	wavevector = wavevector/wavevector.getR() * 2 * M_PI / wavelength;
+
+	return B_0 * (unitVector_1 * cos(wavevector.dot(deltar)) + unitVector_2 * sigma * sin(wavevector.dot(deltar)));
+}
+
+Vector3d PolarizedSingleModeMagneticField::getMaxHelFracField(const Vector3d &pos, const double &fH, const std::string &Bflag) const {
+	if (abs(fH) > 1)
+		throw std::runtime_error("The value of sigma has to be in the range [-1;1].");
+
+	if (Bflag == "B0") {
```
rafaelab, line 42: Is B0 the most appropriate name for the flag? Will the user know that it means amplitude?
Perhaps Bflag="amplitude" and "rms" would be more adequate. (I suggest to remove the 'B' from 'Brms' because 'B' is already contained in the name 'Bflag'.

rafaelab (changes requested):

src/magneticField/PolarizedSingleModeMagneticField.cpp
asavelie: Is this really the case? I found throw std::runtime_error error handling in most of the source files. Therefore, I would prefer to keep this way, unless there is a big majority against it.

include/crpropa/magneticField/PolarizedSingleModeMagneticField.h
asavelie: I could extend "Frac" to Fraction... My idea was that since fH is explained in the paper, it would be clear that HelFrac is related to that.

src/magneticField/PolarizedSingleModeMagneticField.cpp
asavelie: I agree - that is a better way to do it, I will take care of that.

asavelie:
> @asavelie How do we test it quickly? What is the simplest known result one should obtain from this?

Well, since there is an analytical form of the magnetic fields, I would say that the simplest test should be to plot a vector plot or a field line plot of the magnetic field for different parameter combinations.

rafaelab:
Hi @asavelie 

This is a much simpler version now!

I have just one (rather important) concern. The function `getField` is called at every step. The shorter it is, the faster the code runs. A lot of unnecessary checks are being done with all the if/else's. 
Maybe there could be one `getField()` for each scenario. For instance:

```
class LinearlyPolarizedMagneticField: public PolarizedMagneticField {

LinearlyPolarizedMagneticField(const double &B_0, const double &wavelength, const double &sigma, const Vector3d &r_0, const Vector3d &e_1, const Vector3d &e_2, std::string flagAmplitudeRms) :
			B_0(B_0), wavelength(wavelength), sigma(sigma), r_0(r_0), e_1(e_1), e_2(e_2), flagAmplitudeRms(flagAmplitudeRms) {
	}

Vector3d getField(const Vector3d &position) const;
```
In this implementation, there would be something like a class `CircularlyPolarizedMagneticField` and a `EllipticallyPolarizedMagneticField`.
In terms of work, this would mean just to separate the current `getField` function in a few others to remove all the if-statements.

rafaelab:
@asavelie 
I forgot to say: could you also update the CHANGELOG file?

src/magneticField/PolarizedSingleModeMagneticField.cpp
asavelie: In the new verison, there are no warnings, just errors. For the latter I used runtime_errors.

include/crpropa/magneticField/PolarizedSingleModeMagneticField.h
asavelie: In the new, completely renewed version, this is not an issue anymore.

src/magneticField/PolarizedSingleModeMagneticField.cpp
asavelie: In the new version I renamed the flags according to your suggestions.

asavelie:
> @asavelie How do we test it quickly? What is the simplest known result one should obtain from this?

I implemented a test in the MagneticField test source file.



> Hi @asavelie
> 
> This is a much simpler version now!
> 
> I have just one (rather important) concern. The function `getField` is called at every step. The shorter it is, the faster the code runs. A lot of unnecessary checks are being done with all the if/else's. Maybe there could be one `getField()` for each scenario. For instance:
> 
> ```
> class LinearlyPolarizedMagneticField: public PolarizedMagneticField {
> 
> LinearlyPolarizedMagneticField(const double &B_0, const double &wavelength, const double &sigma, const Vector3d &r_0, const Vector3d &e_1, const Vector3d &e_2, std::string flagAmplitudeRms) :
> 			B_0(B_0), wavelength(wavelength), sigma(sigma), r_0(r_0), e_1(e_1), e_2(e_2), flagAmplitudeRms(flagAmplitudeRms) {
> 	}
> 
> Vector3d getField(const Vector3d &position) const;
> ```
> 
> In this implementation, there would be something like a class `CircularlyPolarizedMagneticField` and a `EllipticallyPolarizedMagneticField`. In terms of work, this would mean just to separate the current `getField` function in a few others to remove all the if-statements.

Done! Now the if statements are only called once in the beginning.



> @asavelie I forgot to say: could you also update the CHANGELOG file?

Done!

rafaelab:
Hi @asavelie 
Now it is read to be merged. But there are some conflicts. 
You probably should fetch the upstream version of the repository to sync your branch with the main CRPropa repo.
Once that is done, I'll approve and merge the PR immediately.

CMakeLists.txt
```diff
@@ -365,6 +365,72 @@ file(GLOB_RECURSE CRPROPA_INCLUDES RELATIVE ${CMAKE_SOURCE_DIR} include/*.h)
...
+	src/magneticField/ArchimedeanSpiralField.cpp
+	src/advectionField/AdvectionField.cpp
+	src/massDistribution/Nakanishi.cpp
+	src/massDistribution/Massdistribution.cpp
+	src/massDistribution/Ferriere.cpp
+	src/massDistribution/Cordes.cpp
+	src/massDistribution/ConstantDensity.cpp
+
+	${CRPROPA_EXTRA_SOURCES}
+=======
```
rafaelab, line 433: Line 433 is problematic. @asavelie : you probably forgot to remove the "===" after solving the conflicts.

rafaelab:
@asavelie 
I'm having trouble compiling it. You probably synced your branch with the upstream but not CMakeLists.txt.
I believe this issue is fixed by removing the line "src/GridTurbulence.cpp", since the file no longer exists. 
If you are doing that, maybe you can also make sure to use the same indentation for everything (src/module/Redshift.cpp is aligned left; your own PolarizedSingleMode... and TF17 also).
After that I will merge your PR.

Issue #322, opened 2020-12-16
TurbulenceSpectrum.sIndex is initialized to 5, not 5/3


antoniusf:
**Describe the bug**
I would expect the TurbulenceSpectrum constructor to initialize sIndex to 5/3 by default (and this seems to be the [intent of the code](https://github.com/CRPropa/CRPropa3/blob/1aad73b4c2ee6e7005f43dd33695ee6c943b9241/include/crpropa/magneticField/turbulentField/TurbulentField.h#L52)). Instead, it appears to be initialized to 5.

**To Reproduce**
```python
from crpropa import TurbulenceSpectrum
spectrum = TurbulenceSpectrum(1, 1, 10)
print(spectrum.getSindex())
```

**Expected behavior**
The output should be 5/3, ie someting like 1.67. (The actual output is `5.0`.)

**System (please complete the following information):**
 - Operating system: Debian 10
 - Python version: 2.7.16
 - Swig version: 3.0.12
 - Cmake version: 3.13.4

lukasmerten:
I can confirm that issue.

#### System
Operating system: Ubuntu 18
Python version: 3.6.9
SWIG Version: 3.0.12
cmake version: 3.10.2

PR #335, opened 2021-04-16
Fix default argument of TurbulentField

lukasmerten:
This should fix issue #322.

For some reason, which I do not understand, only the numerator of sIndex= 5. / 3. was initialized in the constructor. Adding brackets around that fraction seems to fix the problem.


lukasmerten:
I can confirm now that this issue was likely related to SWIG. The now implemented c++ test works fine also without the update while the python test is only working with the fix introduced here. 

A similar fix might be needed in other classes, however, on first quick look I could not find any.

adundovi (approved):
A bit troublesome issue coming from SWIG, though not surprising! Nevertheless, great to have these tests :-)

PR #336, opened 2021-04-30
Fix KISS Warnings in gas density classes

lukasmerten:
@avvliet found that a lot of KISS warnings are raised, when he tested the massDistribution examples. 

This was related to a bug in the implementation of the check if a warning should be raised. Now the KISS warning is raised when one or more components (HI, HII, H2) of the gas distributions is available but *not* activated. The option to deactivate individual components was implemented to allow for combinations of several gas distribution models. 

Furthermore, some typos were fixed and the formatting of the warnings and the getDescription method updated.

### DIFF
--- a/src/massDistribution/ConstantDensity.cpp
+++ b/src/massDistribution/ConstantDensity.cpp
@@ -26,14 +26,11 @@ double ConstantDensity::getDensity(const Vector3d &position) const {
 	if(isH2)
 		n += H2densitynumber;
 
-	// check if any density is activ and give warning if not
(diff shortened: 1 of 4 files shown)


avvliet (approved):
@lukasmerten These changes look good to me. I've tested the implementation on the examples on the website. There are still a lot of warnings raised in the last part of the 2nd diffusion example, but I guess it makes sense to give a warning there.

lukasmerten:
Yes, the warnings in the second part should show up, as some of the gas density components are deactivated.

PR #337, opened 2021-05-05
Example Notebook Updates

lukasmerten:
Update the examples to reduce the number of deprecation warnings.
This is mainly achieved by switching to the new TurbulentField classes and updating the observer spheres. 
At the moment one deprecation warning remains unsolved: Whenever the turbulent component of the JF12 field is used a deprecation warning is raised. Here we should probably update the generation of the turbulent component; maybe @adundovi has an idea for this.

The last part of galactic_backtracking example was not working. I think it's fixed now, but maybe @TobiasWinchen can confirm that I did not change any relevant part.

At the moment EMTripletPairProduction is really slow (even compared with all other EMInteraction modules). To me it looks like there might be some kind of a memory leak in this module. Therefore, I commented this module out in the secondaries/photons example. When I can track this problem more precisely, I will open an issue.


lukasmerten:
@rafaelab and @cheiter could you please check if the return statement that I added in 451ca46 makes sense?

With this addition EMTripletPairproduction is not leading to a memory overflow for me. 

From the comments in the code and looking at other EM-modules it seems to me that this exit was missing.

rafaelab:
Hi @lukasmerten 

>  With this addition EMTripletPairproduction is not leading to a memory overflow for me.

It does make sense and it explains why the physics seems to work as expected. 
This is also how it is implemented in EMInverseComptonScattering.

lukasmerten:
Thanks for your confirmation, @rafaelab. 

I guess it would be good to merge this PR now (after someone double-checked it) and fix the deprecation warning in the JF12 field separately. Otherwise it might get too confusing.

rafaelab:
@lukasmerten 
I will take a look at the rest of the PR later this afternoon and merge it.

rafaelab (approved):

PR #342, opened 2021-06-08
Fix #264, builtin option can be safely used again with SWIG 4.0.2+

### LINKED ISSUE #264: Reenable swig --builtin
builtin is broken with current swig versions (See https://github.com/swig/swig/issues/1595 and https://github.com/CRPropa/CRPropa3/pull/250). As soon as this is fixed we should:
- Create tests for the issue
- Enable builtin depending on the swig version

For swig 4.0.2 this issue will be apparently fixed: https://github.com/swig/swig/commit/cb5d7398b562e77436e5766fb17a40bfe8c4f973

adundovi:
Here is PR to fix #264, I've tested it with and without SWIG builtin enabled in cmake. Now, all tests pass regarding the vector3 representation, i.e., print(v), and [the issue with keyword arguments](https://github.com/swig/swig/issues/1595). Also, the new test for checking out of range when setting or getting vector elements passes...


TobiasWinchen (changes requested):
I don't see why the JF12 test is needed here; the builtin change looks good to me. Build fails on MacOS though.

adundovi:
OK, I'll remove these tests. I used them just to make sure nothing is changed physically in the process. macOS fails because it lacks the FFTW support - i.e., all turbulence features are disabled in CI. By removing tests this will not be an issue any more.

TobiasWinchen:
The JF12 deprecation warning has nothing to do with the builtin, right? Would be cleaner to have this in a separate PR then.

adundovi:
> The JF12 deprecation warning has nothing to do with the builtin, right? Would be cleaner to have this in a separate PR then.

Yes, that's true - I accidentally mixed those two commits.

PR #353, opened 2021-07-28
Improve documentation; apply code conventions

rafaelab:
This PR aims at making the code more uniform and at improving the documentation. 
- The style conventions described in [CONTRIBUTING.md](https://github.com/CRPropa/CRPropa3/blob/master/CONTRIBUTING.md) are applied.
- Comments were added to header files to be automatically displayed in the documentation.
- Some example notebooks were fixed (e.g. CMB->CMB())
- A new example for electromagnetic cascades with thinning was added

PR #360, opened 2021-09-09
Depecration warnings for DINT/EleCa

rafaelab:
DINT and EleCa haven't been supported for some time, since release 3.1.5, I believe. This PR simply adds deprecation warnings and points the users to the native CRPropa EM* modules.


rafaelab:
This PR mistakenly follows from #353. I will make a new clean one with only the relevant commits. Apologies for the mistake!

Issue #361, opened 2021-09-15
Reflective boundary conditions show undesirable behavior like discontinuities and asymmetry


antoniusf:
**Describe the bug**
While extending @LeanderSchlegel's work on tricubic interpolation to support reflective boundary conditions, we discovered some problems with the current implementation, such as:

 - the interpolation "going backwards" in mirrored sections of the grid, leading to an incorrectly interpolated grid, along with discontinuities,
 - the grid not being centered inside the used boundaries, meaning that points on opposite faces of the original grid being treated differently; some points occur three times in succession, others only once,
 - the reduction function returning out-of-bounds indices, leading to memory access violations and garbled results.

For more details, including the formal analysis which resulted in the discovery of these bugs, please see the [original issue][1] in Leander's repository. We're opening this issue here to formally report the bug and get a fix merged upstream.

**To Reproduce**

```
from crpropa import Grid1f, Vector3d
import numpy as np
import matplotlib.pyplot as plt

grid = Grid1f(Vector3d(0), 3, 1.0)

grid.setValue(0, 0, 0, 1)
grid.setValue(0, 1, 0, 2)
grid.setValue(0, 2, 0, 0.5)

grid.setValue(1, 0, 0, 3)
grid.setValue(1, 1, 0, 4)
grid.setValue(1, 2, 0, 0)

grid.setReflective(True)

# this uses N = 901 to exacerbate the cases where we sample exactly
# on-grid, highlighting the memory access issues
N = 901
sampled_field = np.zeros((N, N))

for (ix, x) in enumerate(np.linspace(0.5, 9.5, N)):
    for (iy, y) in enumerate(np.linspace(0.5, 9.5, N)):
        sampled_field[ix, iy] = grid.interpolate(Vector3d(x, y, 0))

# uncomment vmin and vmax to observe artifacts resulting from oob access
# (note that, naturally, this is not deterministic and might not work reliably)
plt.imshow(sampled_field.T, origin="lower", vmin=0, vmax=2)
plt.colorbar()
plt.show()
```

**Expected behavior**
Definitely not this:

![Figure_1](https://user-images.githubusercontent.com/2171837/133501817-64fae856-6672-4c59-9f37-1b98d7d43ecd.png)

**System (please complete the following information):**
All – this is an algorithmic issue in `Grid.h`.

**Additional context**
See our comments in the [original issue][1].

[...]

PR #363, opened 2021-09-16
Improved interpolation routines and bugfix of reflective boundary

LeanderSchlegel:
Dear all,

with this pull request we want to contribute an additional tricubic- and nearest neighbour interpolation for gridbased data to CRPropa, as reviewed regarding the technical properties and physical effects in (Schlegel et al 2020 ApJ 889 123 / https://arxiv.org/abs/1907.09934).
Also we want to fix the bug in the reflective boundary condition that was described in issue #361 

New interpolations
------------------

First, we introduced a new grid property named "interpolationType" allowing the user to switch between the three now available interpolation routines for scalar and vectorgrids, setting it to "TRILINEAR" , "TRICUBIC", "NEAREST_NEIGHBOUR" ,respectively, while trilinear interpolation remains the default case.

The new nearest neighbour interpolation could simply be implemented by using the existing closestValue function.

The new tricubic interpolation is implemented according to https://www.paulinternet.nl/?page=bicubic and http://graphics.cs.cmu.edu/nsp/course/15-462/Fall04/assts/catmullRom.pdf.

Optimization
----------------

The tricubic interpolation was SIMD optimized on vectorgrids to gain faster execution times. However, there is no option to use it without the SIMD extension, therefore, we introduced a new CMAKE flag `HAVE_SIMD` depending on the `SIMD_EXTENSION` flag given in the installation. Scalargrids work still without SIMD.
In the tricubic interpolation Grid3d are downcasted to float-precision and a corresponding warning is given.

Test coverage
-------------

The unit tests in testCore.cpp were extended to cover next to the trilinear interpolation also the new interpolations and boundary conditions analogue to the existing tests.

Bug fix
--------

The reflective boundary condition had a bug as described in issue #361 
Here, the grid was not centered inside of the tiling cube that was mirrored by the clamping and the fractional distances
used as weights in the trilinear interpolation were calculated with unreduced/unmirrored coordinates. Sometimes indices were out of the grid range. 
We fixed it on the way by shifting the mirroraxis and using the reduced/mirrored coordinates to calculate the fractional distances to the neighbours.

In the following plot all interpolations with both correct boundaries are shown with the example given in #361 

fixed2

Thank you in advance!
Leander, @antoniusf, @JulienDoerner, @reichherzerp, @eicheB


lukasmerten:
After a quick look this makes totally sense to me. I will compile this version tomorrow and report back after running you tests. 

You could already now resolve the conflict in the CMakeLists file.

lukasmerten:
@LeanderSchlegel 
I compiled and tested this version locally. I could not find any strange behavior. Warning messages, when compiling without SIMD extensions are meaningful. Downcasting warning when using Grid3d works, too. 
There is one last request:

- [ ] Please add two short sentences in the CHANGELOG file under new features and bug fixes.

LeanderSchlegel:
@lukasmerten Thank you for your testing! Just added the short changelog :)

lukasmerten (approved):

PR #366, opened 2021-10-01
Custom magnetic fields via Python

### LINKED ISSUE #365: Custom magnetic fields ignored by the propagation modules
**Describe the bug**
A custom magnetic field defined in a Python class is accepted, but apparently not used by the propagation modules `PropagationCK` and `PropagationBP`, resulting in a constant zero field being effectively employed. Though the  problem affects both `PropagationCK` and `PropagationBP`, it seems easier to demonstrate using the latter as it provides access to the field it is using.

Is the bug related to the physics in CRPropa?
No.

**To Reproduce**

`custom_mf.py` script where a simple custom magnetic field is defined:
```
#!/usr/bin/env python
# coding=utf-8

import time
import crpropa
from crpropa import *


class MyUniform(MagneticField):
    def __init__(self, val):
        super().__init__()
        
        self.val = val
        
    def getField(self, position):
        return Vector3d(self.val)


if __name__ == "__main__":
    field = MyUniform(1*gauss)
    
    source = Source()
    source.add(SourcePosition(Vector3d(0, 0, 0)))
    source.add(SourceParticleType(nucleusId(1, 1)))
    source.add(SourceEnergy(1*EeV))
    
    observer = Observer()
    observer.add(ObserverSurface(Sphere(Vector3d(0, 0, 0), 1*Mpc)))
    
    prop = PropagationBP(field, 1e-4, 1*pc, 1*Mpc)
    
    simulation = ModuleList()
    simulation.add(prop)
    
    simulation.add(observer)
    simulation.setShowProgress(True)
    
    tstart = time.time()
    simulation.run(source, 10, True)
    tstop = time.time()
    print(f'Run time: {tstop - tstart:.2f} s')
    
    print(f'Field value (direct): {field.getField(Vector3d(-1, 0, 0))}')
    print(f'Field value (propagation): {prop.getFieldAtPosition(Vector3d(-1, 0, 0), 0)}')
```
Upon execution it prints:
```
python3 custom_mf.py 
crpropa::ModuleList: Number of Threads: 8
Run time: 0.00 s
Field value (direct): Vector(0.0001, 0.0001, 0.0001)
Field value (propagation): Vector(0, 0, 0)
```
The value retrieved from the propagation module is different from the value the field actually has. The run time also indicates the zero field value is actually used. Replacing `PropagationBP` with `PropagationCK` does not change the outcome (the last two lines should be commented out in this case as there's no `getFieldAtPosition()` method in this case).

If the custom field is replaced with a built-in analogue - i.e. L.20 is replaced with `field = UniformMagneticField(Vector3d(1*gauss))`, - the two field value estimates match and run time become noticeable, indicating the propagation works properly:
```
python3 custom_mf.py 
crpropa::ModuleList: Number of Threads: 8
Run time: 2.87 s
Field value (direct): Vector(0.0001, 0.0001, 0.0001)
Field value (propagation): Vector(0.0001, 0.0001, 0.0001)
```

**Expected behavior**
The supplied non-zero custom field should be used by the propagation module.

**System (please complete the following information):**
 - Operating system: Ubuntu Linux 20.04
 - Laptop or computing cluster? Singularity and Docker containers with aforementioned OS running on a laptop
 - Python version: 3.8
 - Swig version: 4.0.1
 - Cmake version: 3.16.3

**Additional context**
The same issue encountered also if the custom field is built as a CRpropa plugin (the swig "-builtin" option needs to be disabled in this case, following #305). For brevity the corresponding code is not attached here, but can provided upon request, if needed. 
The issue was tested with CRPropa v3.1.7.

reichherzerp:
This PR fixes #365. Currently, CRPropa can't handle custom magnetic fields implemented in Python. This PR allows for custom magnetic fields via Python.

**Background**
To correctly implement a Python class that inherits from the wrapped C++ MagneticField class that contains virtual methods, cross-language polymorphism via directors is needed. The child Python class for the custom magnetic field should override the getField base class function. However, as we don't have enabled directors for the MagneticField class, this fails currently.

**Changes**
Inserting the decorator in the `2_headers.i` file solves the issue. For enhanced testability and synchronization between the propagation modules, I introduced getField functions for PropagationCK and DiffsuionSDE, similar to PropagationBP. I also added better handling of possible magnetic field issues during the propagation.

### DIFF
--- a/include/crpropa/module/DiffusionSDE.h
+++ b/include/crpropa/module/DiffusionSDE.h
@@ -79,12 +79,12 @@ public:
 	    double getEpsilon() const;
 	    double getAlpha() const;
 	    double getScale() const;
-		ref_ptr<MagneticField> getField() const;
+		ref_ptr<MagneticField> getMagneticField() const;
 		/** get magnetic field vector at current candidate position
 		 * @param pos   current position of the candidate
 		 * @param z	 current redshift is needed to calculate the magnetic field
 		 * @return	  magnetic field vector at the position pos */
-		Vector3d getFieldAtPosition(Vector3d pos, double z) const;
+		Vector3d getMagneticFieldAtPosition(Vector3d pos, double z) const;
 	    std::string getDescription() const;
 
 };
--- a/include/crpropa/module/DiffusionSDE.h
+++ b/include/crpropa/module/DiffusionSDE.h
@@ -85,7 +85,12 @@ public:
 		 * @param z	 current redshift is needed to calculate the magnetic field
 		 * @return	  magnetic field vector at the position pos */
 		Vector3d getMagneticFieldAtPosition(Vector3d pos, double z) const;
-	    std::string getDescription() const;
+	    ref_ptr<AdvectionField> getAdvectionField() const;
+		/** get advection field vector at current candidate position
+		 * @param pos   current position of the candidate
+		 * @return	  magnetic field vector at the position pos */
+		Vector3d getAdvectionFieldAtPosition(Vector3d pos) const;
+		std::string getDescription() const;
 
 };
 /** @}*/
--- a/include/crpropa/module/DiffusionSDE.h
+++ b/include/crpropa/module/DiffusionSDE.h
@@ -79,6 +79,13 @@ public:
 	    double getEpsilon() const;
 	    double getAlpha() const;
 	    double getScale() const;
+		 /** get functions for the parameters of the class PropagationCK, similar to the set functions */
+		ref_ptr<MagneticField> getField() const;
+		/** get magnetic field vector at current candidate position
+		 * @param pos   current position of the candidate
+		 * @param z	 current redshift is needed to calculate the magnetic field
+		 * @return	  magnetic field vector at the position pos */
+		Vector3d getFieldAtPosition(Vector3d pos, double z) const;
 	    std::string getDescription() const;
 
 };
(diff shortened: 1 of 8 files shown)


TobiasWinchen:
Looks good to me - might be good practice to work the code from the bug report into a test.

reichherzerp:
Thanks @TobiasWinchen - I added tests that govern the bug report and check the new magnetic read functionality in the propagation modules (CK and SDE).

adundovi:
Nice!

adundovi (approved):

lukasmerten:
@reichherzerp nice work. I only have a few things in mind:

- [ ] Can you add the same functionality for the advection field, too? 

Also the getField/getFieldAtPosition methods are not an unambiguous method name for the DiffusionSDE module as this propagation module can have two different fields (magnetic for the diffusion directions and another one for the advection). 

- [ ] Here, we could implement two methods "getMagneticFieldAtPosition" and getAdvectionFieldAtPosition".

If I remember correctly the gas and photon density modules could also be improved with the PR. 

- [ ] To be consistent throughout the code we should probably do it, too.

reichherzerp:
Thanks for the feedback. I also think that your suggestions fit well into the framework of the current PR, which is why I implemented the proposed enhancements. I have then also tested the new cross-language polymorphism features and all tests pass. Is it like you had it in mind?

lukasmerten:
Looka good to me now. I will install this version locally and run the tests also on my system. I think as long as the automated tests are not back only it's good to have at least some people run the tests to not only rely on a single system setting.

PR #376, opened 2021-11-24
Implementation of the new PolarizedSingleModeMagneticField class (new)

asavelie:
This contribution adds the possibility of defining a magnetic field with a single polarization mode in CRPropa. This includes the more specific circular and linear polarizations, but also the more general elliptical one, as well as a formalism to handle magnetic helicity.


rafaelab (approved):

lukasmerten:
I guess we can then close #326?

PR #377, opened 2021-11-25
Update CRPropa download data

lukasmerten:
This PR updates the data download to a new version (data-2021-11-25).

This data is reproducible with the current version of the crpropa-data repository. All files include now information in the header that specifies the git hash of the crpropa-data repository that was used to create them
This made small changes in CRPropa's files readers necessary. 

Furthermore, this PR improves the CRPropa internal calculation of the redshift scaling of photon target fields.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -336,7 +336,7 @@ endif(APPLE)
 # Download data files (interaction data, masses, decay data ...)
 # ----------------------------------------------------------------------------
 OPTION(DOWNLOAD_DATA "Download CRPropa data files" ON)
-set(CRPROPA_DATAFILE_VER "2021-11-25")
+set(CRPROPA_DATAFILE_VER "2021-11-29")
 if(DOWNLOAD_DATA)
   message("-- Downloading data file from crpropa.desy.de ~ 65 MB")
   file(DOWNLOAD
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -336,7 +336,7 @@ endif(APPLE)
 # Download data files (interaction data, masses, decay data ...)
 # ----------------------------------------------------------------------------
 OPTION(DOWNLOAD_DATA "Download CRPropa data files" ON)
-set(CRPROPA_DATAFILE_VER "2021-07-30")
+set(CRPROPA_DATAFILE_VER "2021-11-25")
 if(DOWNLOAD_DATA)
   message("-- Downloading data file from crpropa.desy.de ~ 65 MB")
   file(DOWNLOAD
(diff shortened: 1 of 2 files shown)


rafaelab (changes requested):
@lukasmerten Some tests are not passing. I believe the main issue is that "photon_emission_IRB_Gilmore12.txt" in photodisintegration should be simply "photon_emission_IRB.txt"

There is also an error in testCore with the grid periodicity. @LeanderSchlegel: I think this is related to your interpolation committee. Could you please take a look at it?

LeanderSchlegel:
Hi @rafaelab and @lukasmerten,

Julien and I just tried to check out this branch and testCore passed on our machines, do you get a specific error?

rafaelab:
> Hi @rafaelab and @lukasmerten,
> 
> @JulienDoerner and I just tried to check out this branch and testCore passed on our machines, do you get a specific error?

```
C++ exception with description "Tried to use tricubic Interpolation without SIMD_EXTENSION. SIMD Optimization is neccesary for tricubic interpolation of vector grids.
" thrown in the test body.
```
and
```
[  FAILED  ] 2 tests, listed below:
[  FAILED  ] Grid3f.Periodicity
[  FAILED  ] Grid3f.Reflectivity
```
The test should not depend on SIMD, it should pass regardless.

It's probably just a matter of testing it if SIMD is true.

lukasmerten:
@rafaelab Can you try to run make install first and than run make test again?
For me all tests are passing, with SIMD.

The failing test dor PDI might be related to not updated data files. When an old CRPropa installation on a machine exists, it can happen that CRPropa is loading the old data files during make test. Running "Make install" should fix it, as this will update the existing data files, e.g., in the /share/ directory.

If that's not fixing the problem let me know.

LeanderSchlegel:
We just looked in the code and the error should only be thrown in Grid.h if the flag HAVE_SIMD is not set and tricubic interpolation is used. If HAVE_SIMD is false, the test for tricubic interpolation in testCore.cpp should not be executed, and the only way to get the error message should be if the flag has got changed during this test. We tested it with all available SIMD-options (none, avx, avx+fma, native) but did not get an error message.

rafaelab:
> @rafaelab Can you try to run make install first and than run make test again? For me all tests are passing, with SIMD.

I tested it on a clean installation (no other CRPropa version running or installed). `make install` did not solve it.
Maybe there is an issue with SIMD and OS X? 

Since this seems to be a local problem only I am having, I will approve this PR once @lukasmerten fixes the small bug with the filename.
I will investigate the problem later (test it in another machine). If it persists, I'll report the issue.

lukasmerten:
After checking my installation again, I can confirm the wrong photon_emission file names. I am currently updating the data-repository and upload a new tarball afterwards.

lukasmerten:
The problematic filenaming should be fixed now. 

However, I cannot reproduce the failing test with SIMD-extensions set to "none". So I cannot provide any fix for that.

rafaelab (approved):

PR #378, opened 2021-11-29
Targeting

avvliet:
This pull request adds a targeted emission routine to the sources. The particles are emitted following a von-Mises-Fisher distribution and given a weight so that it is possible to obtain a distribution as if the particles were emitted isotropically. It also adds an example on how to use this targeted emission, including a learning method to obtain optimised parameters for the von-Mises-Fisher parameters.


lukasmerten:
Hi @avvliet 
In general this looks good to me. 
I'll have a closer look as soon as possible, but the following things came already to my mind:

1. Please add a short comment in the CHANGELOG, so we do not forget this feature in the next release notes.
2. Is the calculation of the rotation parameter  (alpha, delta, etc.) time critical? Since the parameters do not change after a source was initialized, there is no need to recalculate them every time the prepareCandidate method is called.

doc/pages/example_notebooks/targeting/Targeting.ipynb
```diff
@@ -0,0 +1,386 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## The targeting algorithm of CRPropa\n",
+    "\n",
+    "Here we will introduce you to the targeting algorithm in CRPropa, which emits particles from their sources using a von-Mises-Fisher distribution instead of an isotropic distribution. After emission from their sources, the particles gain a weight assigned to them so that the resulting distribution at the observer can be reweighted to resemble an isotropically emitted distribution from the sources. This can lead to significantly larger number of hits compared with starting with an isotropic emission."
```
lukasmerten, line 9: should be "...get a weight assigned..."

lukasmerten (changes requested):

avvliet:
Hi @lukasmerten
Thanks for checking and commenting on this pull request. I've implemented all the points that you mentioned, so please have a look again.

lukasmerten (approved):

lukasmerten:
Thanks for the quick improvements. Let's wait for other other comments for one more day. I'll merge it if nothing else comes up.

rafaelab:
Thanks, @avvliet.
It looks good to me. I think it can already be merged if no one else has comments.

Issue #387, opened 2022-01-17
Problematic definition of the spectral index in PlaneWaveTurbulence


LeanderSchlegel:
Dear all,

we noticed a problematic definition respective a bug regarding the spectral index treatment in PlaneWaveTurbulence and opened this issue to ask you if we see it correctly and formally report the issue. If we should change it, it should only be a small modification of the handling of the spectral index in the source code.

**Definition of the issue**
Using SimpleTurbulenceSpectrum just producing a powerlaw, the default value for the spectral index of 5/3. seems not to result in the expected -11/3. spectrum as shown in the following plot, that reweights the observed spectrum with +11/3. Instead the argument of -1/3. yields the -11/3. spectrum.

![spectrum](https://user-images.githubusercontent.com/32835331/149794071-2b50e6b2-94cf-40ea-94ab-f9733ecf9f5b.png)

**Minimal example to reproduce the issue**
Attached you find an example code to reproduce the plot, based on a script by @antoniusf, that generates the turbulent field, samples it onto a grid to perform a FFT gaining a spectrum that gets reweighted with +11/3.

[example.txt](https://github.com/CRPropa/CRPropa3/files/7883007/example.txt)

Thank you!

rafaelab:
Hi @LeanderSchlegel,
Maybe @adundovi can tell a bit more about the conventions he adopted for the definition of the power spectrum.

adundovi:
Hi @LeanderSchlegel, thanks @rafaelab,

yes, it is true that there could be a mistake in calculating the correct index, I have noticed the same in `PlaneWaveTurbulence`, but didn't have time to play around with it to double check it. The internal conversion should convert it from 5/3 to -11/3.

antoniusf:
Hi, just a quick note from me: The problem appears to be in [`TurbulenceSpectrum::energySpectrum`](https://github.com/CRPropa/CRPropa3/blob/f790e32668edc9bee666c0029f5459d2339d84ac/include/crpropa/magneticField/turbulentField/TurbulentField.h#L80), where the energy spectrum is computed as

<img src="https://render.githubusercontent.com/render/math?math=\frac{k^q}{(1 %2b k^2) ^ { (s %2b q)/2 %2b 1}}">

This is in contrast to the formula given by Tautz and Dosch (2013), which as you know PlaveWaves is implemented after, who explicitly state that for their formula, Kolmogorov turbulence is achieved by setting s to 5/3:

[...]

PR #388, opened 2022-02-21
Uniform interface for the turbulence generation

LeanderSchlegel:
Dear all,

with this pullrequest we want to provide a fix to the issue of the non-uniform spectral index handling, as described in #387.
Analyzing the problem, we made two changes:

First, to solve the problem for the PlaneWaveTurbulence method using SimpleTurbulenceSpectrum that was shown in #387, we effectively removed the "+1" in the exponent in the denominator of the formula of the power spectrum, to be accordant to the Tautz and Dosch 2013 paper, as suggested in the discussion of the issue.

Second, we found that for a uniform handling, it would be good to also change the argument that is given to energySpectrum() in the general case of TurbulenceSpectrum. Instead of giving it k^ = k * l_bendover, we put the multiplication with the bendoverscale inside of the function to have the same argument (just k) as in SimpleTurbulenceSpectrum.

We tested the combinations of the turbulence models (SimpleTurbulenceSpectrum and TurbulenceSpectrum) with the available methods (PlaneWaveTurbulence and gridbased GridTurbulence and SimpleGridTurbulence) and found them working as expected, producing the right spectral behaviour, as shown in the plots below, where the spectrum for s=5/3. is shown reweighted with 11/3. On the left the PlaneWaveMethod is shown and on the right the gridbased methods.

fixed

While testing the FastWaves optimization for the PlaneWaveTurbulence method, we found a minor typo preventing it from being built, that we fixed within this PR.

Thank you in advance,
@JulienDoerner , @LeanderSchlegel

### DIFF
--- a/include/crpropa/magneticField/turbulentField/SimpleGridTurbulence.h
+++ b/include/crpropa/magneticField/turbulentField/SimpleGridTurbulence.h
@@ -21,16 +21,16 @@ namespace crpropa {
 class SimpleTurbulenceSpectrum : public TurbulenceSpectrum {
   public:
 	/**
-	 @param Brms         root mean square field strength for generated field
-	 @param lMin	 Minimum physical scale of the turbulence
-	 @param lMax	 Maximum physical scale of the turbulence
-	 @param lBendover	   the bend-over scale
-	 @param sindex	 Spectral index of the energy spectrum in the inertial
+	 @param Brms		root mean square field strength for generated field
+	 @param lMin	 	Minimum physical scale of the turbulence
+	 @param lMax	 	Maximum physical scale of the turbulence
+	 @param lBendover	the bend-over scale is set to 1000 times lMax to ensure to be in the inertial range.
+	 @param sindex	 	Spectral index of the energy spectrum in the inertial
 	 range
 	*/
 	SimpleTurbulenceSpectrum(double Brms, double lMin, double lMax,
 	                         double sIndex = 5. / 3)
-	    : TurbulenceSpectrum(Brms, lMin, lMax, 0, sIndex, 0) {}
+	    : TurbulenceSpectrum(Brms, lMin, lMax, 1000 * lMax, sIndex, 0) {}
 	~SimpleTurbulenceSpectrum() {}
 
 	/**
--- a/include/crpropa/magneticField/turbulentField/SimpleGridTurbulence.h
+++ b/include/crpropa/magneticField/turbulentField/SimpleGridTurbulence.h
@@ -19,18 +19,18 @@ namespace crpropa {
  @brief Defines the energy spectrum of simple power-law turbulence
  */
 class SimpleTurbulenceSpectrum : public TurbulenceSpectrum {
+	const int constScaleBendover = 1000; 	// define the bandover scale as 1000 * lMax to ensure k * lBendover >> 1. The bendover scale is necessary for the implementation of PlaneWaveTurbulence. 
   public:
 	/**
 	 @param Brms		root mean square field strength for generated field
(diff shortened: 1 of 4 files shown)


antoniusf:
Hiya,

just a short update from me: I've been unable to work the past few weeks due to wrist problems, but I will hopefully get around to reviewing this soon.

A quick note of confusion from me: Why is my name undersigned on the PR comment? I've not had any part in this PR up until now.

Cheers,
Antonius

LeanderSchlegel:
Hi,

no problem, I hope it got better! Thanks for reviewing it.
We wanted to credit your analysis tool and suggestion of the missing "+1" in the issue:)

Cheers,
Leander

antoniusf:
I see, thanks for explaining!

reichherzerp (changes requested):
Thank @LeanderSchlegel, @JulienDoerner, @antoniusf  for this important fix. I agree that it is problematic that the default spectral index of the PlaneWaveTurbulence doesn't lead to the Kolomorov-like Turbulence, as opposed to what is written in the comments of the TD13 methods in the source code. 

I tested your code and it works as expected and documented in your description. 

My only change request regards an improvement of the documentation: Please cite and specify in `PlaneWaveTurbulence.h` the formula of the power spectrum from the Tautz and Dosch 2013 paper that is now used.

If someone else can confirm this pull request, we should consider merging it quickly given the problems that the problematic definition in the current master may cause at the moment!

antoniusf:
Hey, just so we're on the same page on this (because in the mess of the past two months I don't think I've ever communicated this publicly) – the current PR doesn't implement the exact fix suggested in #387 (i.e. fixing the spectrum generation in `TurbulenceSpectrum::energySpectrum`), but instead fixes the generated spectrum in PlaneWaveTurbulence, is that correct? If I see this correctly, the resulting spectrum only *mostly* follows the formula from Tautz & Dosch, since it now outputs



which for k >> 1 (or rather k >> l_bendover) is basically the same as TD13's formula



but since they're not exactly equal (and the approximate equality is only valid for large k), I thought it was worth mentioning this.

Also, @LeanderSchlegel, would you mind removing my name from the PR comment? I don't think it's bad that you want to credit me for some tooling I've written a few years ago, but to me it reads as if I've been involved in the PR somehow and I feel like that might be confusing to others too.

I'm sorry for the delays my injury (and general life circumstances) has caused with respect to my response to this issue; but I hope it's still valuable to you.

JulienDoerner:
hey, 

I corrected the difference @antoniusf reported. Now the correction term is 1 + k^2. 
Therefore it is necessary to introduce a bendoverscale in the simpleTurbulenceSpectrum. In this case I set it to 1000 times the maximal turbulence scale to ensure to be in the limit k^2 >> 1. 
I checked all tests mentioned before and there is no change in the expected behavior.

Additionally I updated the documentation as @reichherzerp requested.

adundovi:
Hi @JulienDoerner, thank you for the hard work you put in to resolve this issue I unintentionally caused by merging these two approaches of generating turbulence. I'll checkout the PR now locally and run a few tests myself. However, putting a number, 1000, is a bit confusing - I would either like to avoid it or define a constant equals to 1000 which is named according to some idea from where this 1000 comes from.

JulienDoerner:
Hey @adundovi, 
I introduced a variable for the scaling factor of 1000 and documented the meaning.

antoniusf:
I haven't tested it, but everything looks good to me too. Thank you so much for your work!

antoniusf (approved):
~wait was i still set as reviewer somehow? anyway,~

nevermind it was just an old tab that hadn't updated completely

PR #389, opened 2022-05-16
Fix example 3 d mhd models

eicheB:
I fixed the example notebook "3D MHD models", so that it is appropriate for the Hackstein et al. (2018) EGMF model. For more details have a look in the appended slides. Still the data files need to be updated - I let you know once it's done. 
[CRPropa_Meeting_May2022.pdf](https://github.com/CRPropa/CRPropa3/files/8701503/CRPropa_Meeting_May2022.pdf)


rafaelab:
Looks good to me, @eicheB.
When the files are replaced, let me know and I will approve and merge the PR.

eicheB:
All data files are now updated on the Desy cloud except for the clues_halos.dat which already seems to have the correct (X,Y,Z) ordering. However, the provided halos from that file are still quite far from being perfectly aligned with the mass density distribution which is a bit puzzling..But at least the magnetic field and the mass distribution clearly match the expected distribution as given in Hackstein et al. (2018). I also updated the readme file and added a note here that mentions the applied changes. I kept the old (original) data files under the folder named "v0", if someone still likes to use them due to some reason.

lukasmerten:
@adundovi, can you please comment on keeping the old files. I am not sure if we run into some data storage limits in the DESY cloud at some point.

lukasmerten:
@eicheB: Can we add your comment on the halos in the example, too? Just to be clear, that there might or might not be a problem with the halos but we dco provide these especially without any warranty.

eicheB:
I though it might be good to also keep the original files in the beginning so that people can still easily access these data files if they like to compute the resulting effect of this switched ordering for example. Then after some time (a year or so) one could of course delete the folder; but we could also delete it right away if you don't see any purpose in it or need some more space on the DESY cloud.

eicheB:
With respect to the halo file I also had some email exchange with Stefan (Hackstein), who wrote the following:
" ... I remember Franco showing me plots with a better fit... The halos were identified in the original data (Enzo output with AMR) searching for density peaks, so they should populate the highest densities...
Probably another mix up of coordinates when translating these to crpropa... Have you tried other permutations of the axes? Maybe the y-axis has to be inverted again? "
And yes, I tried all possible permutations and they all look worse, as you can see from the appended file:
[displayHalos.pdf](https://github.com/CRPropa/CRPropa3/files/8740118/displayHalos.pdf)

You may have a look at it to check if it's done correctly. So far I haven't received any further ideas or comments from Stefan after sending him my results as well.

adundovi:
> @adundovi, can you please comment on keeping the old files. I am not sure if we run into some data storage limits in the DESY cloud at some point.

I don't think we have any limit on the DESY cloud, so we can keep the old files and at some moment later (as @eicheB suggests, a year or so), remove them.

lukasmerten:
Alright, than let's keept the old files for now.

Furthermore, I agree that the current implementation of the halo positions looks best by eye and all other permutations seem to be worse. So I suggest to add somehing like: "Note the the discrepancy between the halo positions and the density maxima. Make sure to double check the halo positions prior to any scientific simulation."

eicheB:
Ok, I just added two more sentences in the readme file. In total, the additional note in that file is the following:
"Note: This is an updated version of the data files 
(original data files can be found in the "v0" folder)
with an (X,Y,Z) ordering of the coordinates 
(instead of the (Z,Y,X) ordering of the original data).
Still we noticed a discrepancy between the halo positions (given by clues_halos.dat) 
and the mass density maxima (as given by mass-density_clues.dat). 
Make sure to double check these halo positions prior to any scientific simulation."

lukasmerten (approved):

PR #391, opened 2022-06-07
Log observer time evolution

reichherzerp:
Dear all,

with this pull request, I want to provide the possibility to use a logarithmic scaling of the ObserverTimeEvolution, which is especially useful if you want to resolve different time scales in your simulations. The feature has proven useful in numerous studies (running diffusion coefficients, lightcurves in sources, ...) over the past few years, which is why I'd like to share it. 

The functionality of the linear observer remains unchanged:
`ObserverTimeEvolution(min, dist, numb)`

With constructor overloading, I implemented the log-scaling in analogy to numpy.logspace:
`ObserverTimeEvolution(min, max, numb, True)`

Test to verify the feature are implemented.

Best,
Patrick


lukasmerten (approved):
Looks good to me. If there are no objections within the next days, I will merge it.

PR #394, opened 2022-07-06
Update the data download to the latest crpropa-data repository version

lukasmerten:
This PR will update the downloaded tables to the latest crpropa-data repository version.

This includes new routines to improve the calculation of secondary yields of electromagnetic interactions. It should close the Inverse compton issue.


rafaelab:
Thanks, @lukasmerten 
Just for consistency I would suggest keeping the exact same format ("2022" instead of "22").

lukasmerten:
Done.

rafaelab (approved):

PR #396, opened 2022-07-13
Inheriting weights from parent particles

lukasmerten:
### Issue
Primary candidate weights, e.g. calculated and set by the `SourceTargetedEmission` feature, have not been properly inherited to their secondaries. The secondaries were always created with a default value of weight=1. This could have led to wrong particle spectra etc., when the targeting method was used in combination with non-electromagnetic interactions.[1]

### Solution
Change the behavior of the `addSecondary` method of the `Candidate` class, so that secondary particles will inherit their parents weights on default. 

### Details
 - Fix in the candidate's constructor, which ignored any weight that was given in the constructor.
 - Fixed the setWeight call in addSecondary to set the secondary's weight to w*w_0, where w_0 is the parent particle's weight at creation of the secondary and w is an optional factor for the secondary weight which defaults to w=1.
 - In addition to the `setWeight` method an `updateWeight` method was introduced. Which changes a candidate's weight by multiplication without the need of a getWeight call. --> Used in the acceleration module.
 - Including simple tests of the set/update/getWeight methods and weight inheritance.

### Test
The issue was first noticed by Simone Rossoni, who provided us with some test cases. Additional checks and suggestions for the implementation came from @avvliet.
To illustrate the test here are some before and after plots:
The figure [energy-spectrum-targeted-master.pdf](https://github.com/CRPropa/CRPropa3/files/9103342/energy-spectrum-targeted-int-B.pdf) shows results for a test simulation defined in [targeted-int.txt](https://github.com/CRPropa/CRPropa3/files/9103375/targeted-int.txt), where the weights of the output file have been included. For comparison a run without targeting [2] was done, which lead to the following spectra [energy-spectrum-not-targeted-int-B.pdf](https://github.com/CRPropa/CRPropa3/files/9103405/energy-spectrum-not-targeted-int-B.pdf).
The last plot shows now a run including targeting done with the fixes proposed in this PR [energy-spectrum-targeted-int-B.pdf](https://github.com/CRPropa/CRPropa3/files/9103417/energy-spectrum-targeted-int-B.pdf).

A simpler test can be done based on this jupyter notebook [Target.zip](https://github.com/CRPropa/CRPropa3/files/9103464/Target.zip). Here, a proxy for a not working weight inheritance are weights = 1. After the fix no weights equal to one are found in the output.

[1] In all EM* and the Synchrotronradiation module the inheritance was manually implemented due to the thinning procedures.
[2] Emitting all particles at the source in a cone with the same opening angle as the targeting method, to get similar statistics.

### DIFF
--- a/src/Candidate.cpp
+++ b/src/Candidate.cpp
@@ -7,7 +7,7 @@
 namespace crpropa {
 
 Candidate::Candidate(int id, double E, Vector3d pos, Vector3d dir, double z, double weight) :
-		redshift(z), trajectoryLength(0), weight(1), currentStep(0), nextStep(0), active(true), parent(0) {
+		redshift(z), trajectoryLength(0), weight(weight), currentStep(0), nextStep(0), active(true), parent(0) {
 	ParticleState state(id, E, pos, dir);
 	source = state;
 	created = state;
@@ -82,8 +82,7 @@ void Candidate::setWeight(double w) {
 }
 
 void Candidate::updateWeight(double w) {
-    double w_0 = this -> getWeight();
-    weight = w_0 * w;
+    weight*=w;
 }
 
 void Candidate::setCurrentStep(double lstep) {
@@ -129,12 +128,11 @@ void Candidate::addSecondary(Candidate *c) {
 	secondaries.push_back(c);
 }
 
-void Candidate::addSecondary(int id, double energy, double weight) {
+void Candidate::addSecondary(int id, double energy, double w) {
 	ref_ptr<Candidate> secondary = new Candidate;
 	secondary->setRedshift(redshift);
 	secondary->setTrajectoryLength(trajectoryLength);
-    secondary->setWeight(this -> getWeight());
-    secondary->updateWeight(weight);
+    secondary->setWeight(weight*w);
     secondary->source = source;
 	secondary->previous = previous;
 	secondary->created = previous;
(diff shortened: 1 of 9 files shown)


src/Candidate.cpp
```diff
@@ -124,12 +128,12 @@ void Candidate::addSecondary(Candidate *c) {
...
 }
 
-void Candidate::addSecondary(int id, double energy, double weight) {
+void Candidate::addSecondary(int id, double energy, double w) {
 	ref_ptr<Candidate> secondary = new Candidate;
 	secondary->setRedshift(redshift);
 	secondary->setTrajectoryLength(trajectoryLength);
-	secondary->setWeight(weight);
-	secondary->source = source;
+    secondary->setWeight(weight*w);
```
rafaelab, line 135: To enforce our code style, add white spaces: weight * w.

src/Candidate.cpp
```diff
@@ -81,6 +81,10 @@ void Candidate::setWeight(double w) {
 	weight = w;
 }
 
+void Candidate::updateWeight(double w) {
+    weight*=w;
```
rafaelab, line 85: To keep the code style conventions, "weight *= w"

rafaelab (commented):
Hi @lukasmerten 
It looks all right to me.
I left a couple of comments about code style conventions to keep it all consistent.   Otherwise this can be merged.

PR #407, opened 2022-11-04
Interaction and source tags

JulienDoerner:
With this PR a new Candidate property called `tagOrigin` will be introduced. 

This new feature can be used for two things. 

The first is to distinguish which interaction processes produced the observed candidate. Therefore all interaction processes which can produce a secondary have their own `interactionTag`. This will be given for all produced secondaries and the user is able to customize the tag if the same interaction with different backgrounds, e.g. photonfields, is used. 

The second application is the discrimination of different sources. In case that the sources follows the same spatial distribution or the distribution overlaps it can not be distinguished by the source position anymore. Therefore the new source feature `SourceTag` adds a unique tag to the candidate. 

The `tagOrigin` can be added to the output as a separate column. For the header a comment with the default abbreviations for the interactions is added. 

Also test for all interaction processes and also for the candidate get/set function and the source feature are included.


rafaelab:
Hi @JulienDoerner 
Thanks for this PR. I think we all have been wishing for such a feature for a while.
I haven't checked it yet, but I'm wondering something. My main concern is the possible performance hit that CRPropa takes every time a new candidate property is added. Since now all candidates will have an additional property, this will possibly increase simulation time. Or not? Have you checked this?

JulienDoerner:
Hey @rafaelab 
The performance is the reason why I didn't use the custom property but writing a hard candidate property in line [49](https://github.com/JulienDoerner/CRPropa3/blob/3e1ba07a371518a0b58a5379f99b45ae01649bf8/include/crpropa/Candidate.h#L49). 

For comparing the runtime I took a extreme example with many secondaries. In my test simulation 1e4 iron nuclei with the energy of 1000 EeV are propagated over 100 Mpc. I take the PD and PPP both on the CMB into account and produce all secondaries. 

The result of 50 repetitions can be seen here:
runtime

The difference in runtime is here by about 3%. In other cases with less secondaries the performance impact becomes negligible.

lukasmerten:
I think a three percent simulation time increase is a fair trade off for this feature. 
If you @rafaelab are also fine with that, we can try to merge this soon.

include/crpropa/Candidate.h
```diff
@@ -46,6 +46,7 @@ class Candidate: public Referenced {
 	double trajectoryLength; /**< Comoving distance [m] the candidate has traveled so far */
 	double currentStep; /**< Size of the currently performed step in [m] comoving units */
 	double nextStep; /**< Proposed size of the next propagation step in [m] comoving units */
+	std::string tagOrigin; /**< Name of interaction/source process which creates this particle*/
```
lukasmerten, line 49: I guess it should be past tense hier "[...] created this particle"

include/crpropa/Candidate.h
```diff
@@ -97,6 +99,12 @@ class Candidate: public Referenced {
 	void setNextStep(double step);
 	double getNextStep() const;
 
+	/**
+	 Sets the tagOriginof the particle. Can be used to trace back the interactions
```
lukasmerten, line 103: "[...] tagOrigin of [...]"

include/crpropa/Source.h
```diff
@@ -797,8 +797,27 @@ class SourceGenericComposition: public SourceFeature {
 };
 #endif
 
+/**
+ * @class SourceTag
+ * @brief All particles from this source get a given tag. This can be used to distinguish between differnt sources that follow the same spatial distribution
```
lukasmerten, line 802: "All candidates [...]", "[...] different [...] distribution."

JulienDoerner:
Thanks @lukasmerten for your comments, 
I corrected the typos and renamed the particle to candidate in the new features. But I think this could also be done for the other parts of the Candidate.h file. I would suggest to do this within the PR for the update of the documentation.

src/Candidate.cpp
```diff
@@ -102,6 +102,14 @@ void Candidate::setProperty(const std::string &name, const Variant &value) {
 	properties[name] = value;
 }
 
+void Candidate::setTagOrigin (std::string tagOrigin ) {
```
lukasmerten, line 105: remove whitespace before the closing parenthesis

src/Candidate.cpp
```diff
@@ -128,22 +136,23 @@ void Candidate::addSecondary(Candidate *c) {
 	secondaries.push_back(c);
 }
 
-void Candidate::addSecondary(int id, double energy, double w) {
+void Candidate::addSecondary(int id, double energy, double w, std::string tagOrigin ) {
```
lukasmerten, line 139: remove whitespace before the closing parenthesis

src/Candidate.cpp
```diff
@@ -128,22 +136,23 @@ void Candidate::addSecondary(Candidate *c) {
...
-  secondary->source = source;
+	secondary->setWeight(weight * w);
+	secondary->source = source;
 	secondary->previous = previous;
 	secondary->created = previous;
 	secondary->current = current;
 	secondary->current.setId(id);
 	secondary->current.setEnergy(energy);
 	secondary->parent = this;
+	secondary->setTagOrigin (tagOrigin );
```
lukasmerten, line 151: remove whitespace before the closing parenthesis

src/Candidate.cpp
```diff
@@ -128,22 +136,23 @@ void Candidate::addSecondary(Candidate *c) {
...
 	secondaries.push_back(secondary);
 }
 
-void Candidate::addSecondary(int id, double energy, Vector3d position, double w) {
+void Candidate::addSecondary(int id, double energy, Vector3d position, double w, std::string tagOrigin ) {
```
lukasmerten, line 155: remove whitespace before the closing parenthesis

src/Candidate.cpp
```diff
@@ -157,6 +166,7 @@ void Candidate::addSecondary(int id, double energy, Vector3d position, double w)
 	secondary->current.setPosition(position);
 	secondary->created.setPosition(position);
 	secondary->parent = this;
+	secondary->setTagOrigin (tagOrigin );
```
lukasmerten, line 169: remove whitespace before the closing parenthesis

src/module/TextOutput.cpp
```diff
@@ -123,6 +125,13 @@ void TextOutput::printHeader() const {
 		*out << "# Px/P0x/P1x... Heading (unit vector of momentum)\n";
 	if (fields.test(WeightColumn))
 		*out << "# W             Weights" << " \n";
+	if (fields.test(CandidateTagColumn)) {
+		*out << "# tag           candidate tag can be given by the source feature (user defined tag) or by the following interaction process \n";
```
lukasmerten, line 129: "[...] Candidate [...]"

lukasmerten:
Looks good to me now. 
If I do not hear anything different I will merge this tomorrow or on Friday.

lukasmerten (approved):

lukasmerten:
@JulienDoerner: One last thing. Can you add a note to the CHANGELOG please.

rafaelab:
Thanks for the review, @lukasmerten 
I also finished testing it. I was afraid that there might be some errors because many modules were touched, but this is not the case.
@JulienDoerner: just add this change to the CHANGELO, as Lukas suggested, and then it can be merged.

JulienDoerner:
I added a comment in the change log. 
@lukasmerten can you take a look on the comment and merge afterwards?

PR #408, opened 2022-11-14
Update testing environments

lukasmerten:
This PR updates the automated tests to more recent ubuntu versions. Ubuntu-22 is now included in the test matrix.
The Ubuntu-18 test will only work until 2023-04-01. However, I kept it in the test until then as some people are still using ubuntu-18.

I was not able to update to macos-12. So I removed it again from the mac test matrix. Maybe @rafaelab can at some point help with updating the test.

Also some of the github actions are updated.

Tests should now also be triggered for a PR and not only for a push. --> We get the test results before merging into master.

### DIFF
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -8,21 +8,21 @@ jobs:
       fail-fast: false
       matrix:
         config:
+         - name: "ubuntu-22"
+           os: ubuntu-22.04
+           cxx: "g++-11"
+           cc: "gcc-11"
+           fc: "gfortran-11"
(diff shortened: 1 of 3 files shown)


rafaelab:
Thanks, @lukasmerten. 
I have no comments about this. From my side, it is ready to be merged.
Regarding the OSX tests, for many years the tests have failed for some reason (although the code runs correctly) and I have been able to solve the problem. Right now I can narrow one of the problems down to EMPairProduction and the magnetic lenses.

rafaelab (approved):

PR #419, opened 2023-02-01
Fix Acceleration module

lukasmerten:
The signature of ParticleSplitting's constructor definition differe in the .h and the .cpp files. This is problematic when the constructor is used with positional arguments.

This PR fixes it.

Furthermore: 
 * The Example notebook was updated to include a full example including the particle splitting.
 * The weighting column in the HDFOutput was set from "weight" to "W". This synchronizes the column naming with the TextOutput.

### DIFF
--- a/include/crpropa/module/Acceleration.h
+++ b/include/crpropa/module/Acceleration.h
@@ -152,9 +152,9 @@ class ParticleSplitting : public Module {
 
 	public:
 	/// @params surface               The surface to monitor
-	/// @params crossing_threshold    Number of crossings after which a particle is split
-	/// @params num_splits            Number of particles the candidate is split into
-	/// @params min_weight            Minimum weight to consider. Particles with
+	/// @params crossingThreshold    Number of crossings after which a particle is split
+	/// @params numSplits            Number of particles the candidate is split into
+	/// @params minWeight            Minimum weight to consider. Particles with
 	///                               a lower weight are not split again.
 	/// @params counterid             An unique string to identify the particle
 	///                               property used for counting. Useful if
--- a/include/crpropa/module/HDF5Output.h
+++ b/include/crpropa/module/HDF5Output.h
@@ -82,7 +82,7 @@ class HDF5Output: public Output {
 		double P1x;
 		double P1y;
 		double P1z;
-		double weight;
+		double W;
 		std::string tag;
 		unsigned char propertyBuffer[propertyBufferSize];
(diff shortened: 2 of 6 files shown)


include/crpropa/module/Acceleration.h
```diff
@@ -152,9 +152,9 @@ class ParticleSplitting : public Module {
 
 	public:
 	/// @params surface               The surface to monitor
-	/// @params crossing_threshold    Number of crossings after which a particle is split
-	/// @params num_splits            Number of particles the candidate is split into
-	/// @params min_weight            Minimum weight to consider. Particles with
+	/// @params crossingThreshold    Number of crossings after which a particle is split
```
JulienDoerner, line 155: it should be `@param` without "s"

include/crpropa/module/HDF5Output.h
```diff
@@ -82,7 +82,7 @@ class HDF5Output: public Output {
 		double P1x;
 		double P1y;
 		double P1z;
-		double weight;
+		double W;
```
JulienDoerner, line 85: I would suggest to keep the double naming `weight` as it is easier to read. 
The change should only apply for the naming of the column of the final output file.

src/module/Acceleration.cpp
```diff
@@ -141,8 +141,8 @@ double QuasiLinearTheory::modify(double steplength, Candidate* candidate)
 }
 
 
-ParticleSplitting::ParticleSplitting(Surface *surface, int numSplits,
-		int	crossingThreshold, double minWeight, std::string counterid)
+ParticleSplitting::ParticleSplitting(Surface *surface, int	crossingThreshold, 
+	int numSplits, double minWeight, std::string counterid)
```
JulienDoerner, line 145: Following the [contribution guideline](https://github.com/CRPropa/CRPropa3/blob/master/CONTRIBUTING.md), the number of splits should be called `numberSplits`.

JulienDoerner (changes requested):
The example notebook shows now an error in the second code cell. I think it is only left from the testing. Also the spectrum in the end looks different to the previous version. 
I would suggest rerunning the notebook and update the results.

lukasmerten:
I updated the notebook with a clean run and addressed most of @JulienDoerner comments.

lukasmerten:
Anything else that needs to be done?

JulienDoerner (approved):
From my side the PR looks fine.

PR #427, opened 2023-04-17
Variant improvements

rafaelab:
This PR improves the handling of variants in CRPropa. It updates the previous implementation making the code more efficient. Its main goal is to facilitate the lives of users aiming to write their own plugins. 
This PR also adds some more tests in `testCore.cpp` to check if variants are being treated correctly.

Currently some types like `Vector3d` cannot be treated as a variant. Instead, the user would require three `double` and consequently three operations to handle it.
New types added are: `complex<double>`, `complex<float>`, `Vector3d`, `Vector3f`, `Vector3c` (alias for `Vector3<complex<double>>`, `long double`. A new type for a vector of variants (`vector<Variant>`) was also added.

For some yet-unknown reason these new types have to be called with `variant.toVector3d()`, for instance, whereas operations assigning a variant directly to some types are permitted for the more conventional types like `double`.

### DIFF
--- a/include/crpropa/Variant.h
+++ b/include/crpropa/Variant.h
@@ -4,8 +4,9 @@
 // Licensed under a LGPL-2 or later license                   -
 //-------------------------------------------------------------
 
-#ifndef VARIANT_HH
-#define VARIANT_HH
+#ifndef CRPROPA_VARIANT_H
+#define CRPROPA_VARIANT_H
+
 
 #include <iostream>
 #include <string>
@@ -228,4 +229,4 @@ std::ostream& operator <<(std::ostream& os, const Variant &v);
 
 } // namespace crpropa 
 
-#endif // VARIANT_HH
+#endif // CRPROPA_VARIANT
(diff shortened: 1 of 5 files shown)


lukasmerten:
Hi @rafaelab 
I have tested the implementation and could not find any problems. Although I still do not fully get the advantages of the new implementation, I don't have any reason to hold the PR back.

lukasmerten (approved):

rafaelab:
Thanks, @lukasmerten 

> Hi @rafaelab I have tested the implementation and could not find any problems. Although I still do not fully get the advantages of the new implementation, I don't have any reason to hold the PR back.

There isn't much new. The redesign just allows for a `vector<Variant>` and a few other types that were not available. 
This is particularly useful for plugins, which require `Candidate` properties that don't exist.
The motivation for this was that I needed to use `Vector3` and `vector` as a `Variant`-type to add as a candidate property. This is much more efficient than having, for example, 3 individual `double` variants to represent a `Vector3d`. 
It doesn't break in any way the previous code structure (and now there are also tests for it).

PR #429, opened 2023-05-03
GridProperties: Default Value for Volume Clipping

lukasmerten:
This PR adds a default value for the volume clipping switch to the `GridProperties` module.

Context:
Creating a grid (e.g. `Grid3f`) is possible in different ways. Either grid properties can be defined as separate arguments or a GridProperties instance is used as a single argument for the Grid constructor. 
However, the handling to define the clipVolume value was different depending on the chosen constructor. This behavior is unified by this PR and defaults now to `clipVolume=false` for all Grid constructors. 

Issue:
Using the Grid constructor with the GridProperties argument led to a random initialization of the clipVolume variable.

Problem:
This might have led to an unnoticed effect, e.g. in turbulent magnetic field creation, where the field strength outside the original grid volume would have been set to zero. To see this problem, you can run the following code:
`from crpropa import *`
`randomSeed = 42`
`Brms=8*nG`
`lMin = 60*kpc`
`lMax=800*kpc`
`sIndex=5./3.`
`turbSpectrum = SimpleTurbulenceSpectrum(Brms, lMin, lMax, sIndex)`
`gridprops = GridProperties(Vector3d(0), 256, 30*kpc)`
`BField = SimpleGridTurbulence(turbSpectrum, gridprops, randomSeed)`
`R = Vector3d(10, 1, 5)*Mpc`
`BField.getField(R)`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,13 +3,13 @@
 ### Bug fixes:
 * Re-added ToroidalHaloField and LogarithmicSpiralField models. Note, that the class name was also corrected in spelling: TorroidalHaloField --> ToroidalHaloField
 * Synchronized signature of ParticleSplitting constructor
-* Added default parameter for clipVolume in GridProperties. Could have lead to random clipping of grid values outside of the original grid volume.
 
 ### New features:
-* new candidate property tagOrigin to trace back which source or which interaction created the candidate
-* new interace for massdistributions given on a Grid1f
-* grids can be restricted to the volume without repetition
-* sourceFeature to sample the source position from a given massdistribution
+* New candidate property tagOrigin to trace back which source or which interaction created the candidate
+* New interace for massdistributions given on a Grid1f
+* Grids can be restricted to the volume without repetition
+* SourceFeature to sample the source position from a given massdistribution
+* Added default parameter for clipVolume in GridProperties.
 
 ### Interface changes:
 * Weight column in hdf-Output is now called "W", which is the same as for TextOutput.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,10 +6,9 @@
 
 ### New features:
 * New candidate property tagOrigin to trace back which source or which interaction created the candidate
-* New interace for massdistributions given on a Grid1f
-* Grids can be restricted to the volume without repetition
+* New interface for massdistributions given on a Grid1f
+* Grids can be restricted to the volume without repetition (clipVolume parameter)
 * SourceFeature to sample the source position from a given massdistribution
-* Added default parameter for clipVolume in GridProperties.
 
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -1,14 +1,15 @@
...
 * Re-added ToroidalHaloField and LogarithmicSpiralField models. Note, that the class name was also corrected in spelling: TorroidalHaloField --> ToroidalHaloField
 * Synchronized signature of ParticleSplitting constructor
 
 ### New features:
-* new candidate property tagOrigin to trace back which source or which interaction created the candidate
-* new interace for massdistributions given on a Grid1f
-* grids can be restricted to the volume without repetition
-* sourceFeature to sample the source position from a given massdistribution
+* New candidate property tagOrigin to trace back which source or which interaction created the candidate
+* New interace for massdistributions given on a Grid1f
```
JulienDoerner, line 9: interface

CHANGELOG.md
```diff
@@ -1,14 +1,15 @@
...
+* Grids can be restricted to the volume without repetition
+* SourceFeature to sample the source position from a given massdistribution
+* Added default parameter for clipVolume in GridProperties.
```
JulienDoerner, line 12: I don't think its necessary to mention here the default value, as `clipVolume` is already a new feature.

test/testCore.cpp
```diff
@@ -522,6 +522,12 @@ TEST(Grid1f, SimpleTest) {
 	//nearest neighbour interpolated
 	grid.setInterpolationType(NEAREST_NEIGHBOUR);
 	EXPECT_FLOAT_EQ(7., grid.interpolate(some_grid_point));
+
+	//Test if grid is set to zero outside of volume for clipVolume=true
+	grid.setClipVolume(true);
+	EXPECT_EQ(0, grid.get(0, 0, 12));
+	EXPECT_EQ(0, grid.get(0, 0, 12));
```
JulienDoerner, line 529: Why is it three times the same test?

JulienDoerner (changes requested):

reichherzerp (changes requested):
Hi @lukasmerten, thanks for the PR; I find it important and I have no suggestions for further improvements. It can be merged, once the comments from @JulienDoerner are addressed.

CHANGELOG.md
lukasmerten: I merged the two comments on the clipVolume functionality

test/testCore.cpp
lukasmerten: Fixed, was a copy+paste error.

reichherzerp (approved):

JulienDoerner (approved):

PR #434, opened 2023-07-06
Saldana-Lopez 2021 EBL model

IevgenVovk:
This PR adds the recent [Saldana-Lopez 2021](https://ui.adsabs.harvard.edu/abs/2021MNRAS.507.5144S/abstract) EBL model - along with its lower and upper limits - to CRPropa. The model data files are part of the [related PR](https://github.com/CRPropa/CRPropa3-data/pull/18) to the CRPropa-data repository.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -336,16 +336,18 @@ endif(APPLE)
 # Download data files (interaction data, masses, decay data ...)
 # ----------------------------------------------------------------------------
 OPTION(DOWNLOAD_DATA "Download CRPropa data files" ON)
-set(CRPROPA_DATAFILE_VER "2022-07-06")
+set(CRPROPA_DATAFILE_VER "2023-10-20")
 if(DOWNLOAD_DATA)
-  message("-- Downloading data file from crpropa.desy.de ~ 65 MB")
+  message("-- Downloading data file from crpropa.desy.de ~ 73 MB")
   file(DOWNLOAD
-    https://www.desy.de/~crpropa/data/interaction_data/data-${CRPROPA_DATAFILE_VER}.tar.gz-CHECKSUM
-    ${CMAKE_BINARY_DIR}/data-${CRPROPA_DATAFILE_VER}.tar.gz-CHECKSUM)
+  https://ruhr-uni-bochum.sciebo.de/public.php/webdav/data-${CRPROPA_DATAFILE_VER}.tar.gz-CHECKSUM
+    ${CMAKE_BINARY_DIR}/data-${CRPROPA_DATAFILE_VER}.tar.gz-CHECKSUM 
+    USERPWD "3juW9sntQX2IWBS")
   file(STRINGS ${CMAKE_BINARY_DIR}/data-${CRPROPA_DATAFILE_VER}.tar.gz-CHECKSUM DATA_CHECKSUM LIMIT_COUNT 1 LENGTH_MINIMUM 32 LENGTH_MAXIMUM 32)
   file(DOWNLOAD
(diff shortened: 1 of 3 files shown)


rafaelab:
Thanks for the PR, @IevgenVovk.
I will test it in ~2 weeks. 

For now, I just have one minor comment. 
Since "Saldaña-Lopez" is a composite surname, perhaps the model could be named `IRB_SaldanaLopez21` instead of `IRB_Saldana21`. 
To me it makes no difference at all , so the choice would ultimately depend on the preference of the author of the EBL model (who might just not care about it). 
What I can say is that, as someone with two surnames, I do prefer to be cited with both of them instead of just one.

IevgenVovk:
Thanks for the comment, @rafaelab . `IRB_SaldanaLopez21` was my preference too, but then I've noticed the data files [shared by the authors](https://www.ucm.es/blazars/ebl) are called `*saldana21*`; I've then decided to follow their naming.

rafaelab:
Perfect, @IevgenVovk .
I didn't know that, so I think that settles the issue and we keep the current names :)

rafaelab:
Hi @IevgenVovk 
Everything looks good to me.
Could you just add to `CHANGELOG.md`, under "New features", the information that you added `IRB_Saldana21` EBL model?

IevgenVovk:
Hi @rafaelab , thanks for checking.
Sure, have updated the `CHANGELOG.md`

lukasmerten:
I guess it would be good to include the new model into the default CRPropa models. That means we have to update the default data that is downloaded during the installation process. 

I will run the necessary scripts of CRPropa-data and edit this PR accordingly. Afterwards we can approve and merge.

lukasmerten:
I have updated the data download to a new data package that includes the Saldana 21 data. This is now also reflected in an update to the CRPropa-Data repository (https://github.com/CRPropa/CRPropa3-data/pull/19).

CMakeLists.txt
```diff
@@ -336,16 +336,18 @@ endif(APPLE)
 # Download data files (interaction data, masses, decay data ...)
 # ----------------------------------------------------------------------------
 OPTION(DOWNLOAD_DATA "Download CRPropa data files" ON)
-set(CRPROPA_DATAFILE_VER "2022-07-06")
+set(CRPROPA_DATAFILE_VER "2023-10-20")
 if(DOWNLOAD_DATA)
-  message("-- Downloading data file from crpropa.desy.de ~ 65 MB")
+  message("-- Downloading data file from crpropa.desy.de ~ 73 MB")
```
JulienDoerner, line 341: This download message has to be changed as we do not download the data from the desy server anymore.

JulienDoerner (changes requested):
I checked the installation process. This seems to work for me. 
Also the working test is passing so downloading on the github-action server works too. 

The only thing to change is the download message within the installation.

rafaelab:
Thanks for the contribution, @IevgenVovk.
It works just fine with me.

I have also double checked the CRPropa-data repository modified by @lukasmerten, just to make sure there are no inconsistencies. 

I think we can merge this and just open another PR to fix the minor inconsistency in the information displayed when running CMake, as pointed by @JulienDoerner. I don't suppose Ievgen will have this information (like file size) to implement the change himself.

rafaelab (approved):
@JulienDoerner or @lukasmerten :
Since you are now handling the new data server, could one of you update the CMake information displayed according to the new file size and the new server name?

lukasmerten:
I‘ll update it later today. Thanks for taking a look at it.

JulienDoerner (approved):

Issue #390, opened 2022-05-23
CRPropa3.1.7 installation in macOS 12.3.1


antoniocapanema:
Dear developers,

I have problems installing CRPropa on my system
**System Information:**
 - Operating system: macOS 12.3.1
 - Laptop
 - Python version: 3.9.7
 - Swig version: 4.0.2
 - Cmake version: 3.21.4

I installed version 3.1.7 from git.

I read and followed the installation instructions at https://crpropa.github.io/CRPropa3/ . Before writing this request I also checked previous installation issues at https://github.com/CRPropa/CRPropa3/issues?utf8=%E2%9C%93&q=is%3Aissue+label%3Ainstallation+ . However, I could not find the solution to my problem.

After cmake is complete, upon running the "make" command, I get the following error at error at 74% progress:

[ 74%] Linking CXX shared library libcrpropa.dylib
ld: library not found for -lprofiler
clang: error: linker command failed with exit code 1 (use -v to see invocation)
make[2]: *** [libcrpropa.dylib] Error 1
make[1]: *** [CMakeFiles/crpropa.dir/all] Error 2
make: *** [all] Error 2

Thank you for any assistance you can give.
Best,
Antonio

antoniocapanema:
Hi @lukasmerten,
Thank you so much for your prompt reply. I was able to bypass this issue with the flag -DNO_TCMALLOC=TRUE. However, I ran into another issue a little further ahead during the "make" stage:

---------

[ 88%] Linking CXX shared module _crpropa.so
Undefined symbols for architecture arm64:
  "_PyBytes_AsStringAndSize", referenced from:
      SWIG_AsCharPtrAndSize(_object*, char**, unsigned long*, int*) in crpropa_wrap.cxx.o
  "_PyCMethod_New", referenced from:
      SWIG_PyInstanceMethod_New(_object*, _object*) in crpropa_wrap.cxx.o
  "_PyInstanceMethod_New", referenced from:
      SWIG_PyInstanceMethod_New(_object*, _object*) in crpropa_wrap.cxx.o
  "_PyModule_Create2", referenced from:
      _PyInit__crpropa in crpropa_wrap.cxx.o
  "_PySlice_AdjustIndices", referenced from:
      crpropa_ParticleCollector___getitem____SWIG_1(crpropa::ParticleCollector*, _object*) in crpropa_wrap.cxx.o
  "_PySlice_Unpack", referenced from:
[...]

PR #437, opened 2023-08-08
Update CMakeLists.txt for macos installing

cxwx:
for https://github.com/CRPropa/CRPropa3/issues/390

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -225,6 +225,10 @@ configure_file("${CMAKE_CURRENT_SOURCE_DIR}/src/ModuleList.cpp.in" "${CMAKE_CURR
 list(APPEND CRPROPA_EXTRA_SOURCES "${CMAKE_CURRENT_BINARY_DIR}/src/ModuleList.cpp")
 
 # Google Performance Tools (optional as possible performance tweak for OpenMP)
+if(APPLE)
+  find_package(PkgConfig)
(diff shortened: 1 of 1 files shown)


lukasmerten:
Can someone of the Mac users quickly check this? Looks in general fine to me, but I cannot test it.

rafaelab:
Thanks, @cxwx 
But this doesn't work for me. I get the error message  
```
CMake Error at /opt/homebrew/Cellar/cmake/3.25.2/share/cmake/Modules/FindPkgConfig.cmake:663 (message):
  pkg-config tool not found
Call Stack (most recent call first):
  /opt/homebrew/Cellar/cmake/3.25.2/share/cmake/Modules/FindPkgConfig.cmake:829 (_pkg_check_modules_internal)
  CMakeLists.txt:230 (pkg_check_modules)
```

I also fail to see the underlying logic for *requiring* profiler, as this is not needed to run CRPropa. 
After some quick search in google this problem (#390) seems to be related to Google performance tools, which is also not a requirement for CRPropa. According to [this discussion](https://stackoverflow.com/questions/67034132/gperftools-not-installing-lprofiler-on-mac-after-installing-it-with-brew), this happens when gperf is installed via brew, so it doesn't seem to be a general problem.

lukasmerten:
I close this PR for now. If you, @cxwx, think it is still relevant, please reopen this or a new PR.

PR #457, opened 2023-11-26
Compatibility with newer Python versions

rafaelab:
Several features have been deprecated in newer Python versions, especially after 3.10. Packages like `distutils` and `pkg_resources` some packages are no longer supported. 
This PR adds support for these new Python releases. It also changes the default python finder to CMake's for Python3.12+.
It remove some cmake-related warnings (on newer cmake versions).


lukasmerten:
Hi @rafaelab 
This PR seems to break the MacOS test, which might be related to an old set up of the test runner. Since I have no access to a Mac it woul be good if you either update the macos test or someone, e.g., @sophieaerdker, @LeanderSchlegel or any other mac user can confirm that this PR does compile on more recent macos versions.

lukasmerten:
As discussed:
Please double check if we need matplotlib. Otherwise we can remove the checks for this libraray.

lukasmerten:
I can also confirm that the magnetic lens errors seems to be fixed on:
Debian 12
CMake 3.25.1
gcc 12.2
python 3.9.2

lukasmerten (dismissed):

lukasmerten:
To me it looks like a lot of the problems we are having arise from supporting too old cmake versions.

If we could agree to have the minimum cmake version as 3.14, we could drastically reduce the code of CMakeLists and python.cmake and checkNumpy.py could very likely be removed completely.

Any opinions on that?

JulienDoerner:
I am always using cmake versions much newer than 3.14, so setting this as a requirement would be fine for me.

rafaelab:
I agree, @lukasmerten. Also, if someone has some older CMake version, it is one of the easiest things to update in a system.
Therefore, should we proceed with that? If so, I will implement the changes next week.

lukasmerten:
From my side, you, @rafaelab, have a go for implementing the switch to cmake version >= 14

rafaelab:
This PR should remain on hold a bit longer.
I think I'm on to something (related to the lenses problem) and it is directly related to the way we compile the code and link it to Python. Now I managed to make it work on OSX, but I broke it for Ubuntu. 
I will let you know when it is ready for review.

rafaelab:
Superseded by #466.

Issue #456, opened 2023-11-26
tests for EMPairProduction occasionally fails


rafaelab:
`testInteractions` fails sometimes, for some compiler. The problem is usually in EMPairProduction.
The underlying reason is that it expects 2 secondaries to be created, which is done by `performInteraction`. However, step-size limitation prevents `performInteraction` from being called.

PR #461, opened 2024-01-09
Update Github Actions

JulienDoerner:
Dear all, 

this PR adds two new actions. 

1. The first one is to build the documentation and coverage report. It can be started manual by any CRPropa developer. The documentation and the report can be downloaded afterwards. 
2.  The second action is a test of the example notebooks. This test is based on the `crpropa-testing` test (Ubuntu 20, python3). The notebooks are converted to python scripts and executed afterwards. This test aims to check if interfaces are changed and whether the documentation is updated accordingly. 
At the moment this test excludes the notebooks: `custom-photon-field`, `MHD_models`, `density_grid_sampling`, `lensing_cr` and `lensing_map`. Those notebooks require external data, which are not provided within this repository.

Also in the notebooks the DINT/ELECA part is removed and some ploting is updated to newer matplotlib versions.


.github/workflows/create_documentation.yml
```diff
@@ -0,0 +1,71 @@
...
+        uses: actions/setup-python@v4
+        with:
+          python-version: '3.9'
+          cache: 'pip' # caching pip dependencies
+      - name: Prerequirements
+        run: |
+          sudo apt-get update
+          sudo apt-get install libmuparser-dev libhdf5-serial-dev libomp5 libomp-dev libfftw3-dev libcfitsio-dev lcov doxygen graphviz
+          sudo apt-get install pandoc # do not only use pip to install pandoc, see https://stackoverflow.com/questions/62398231/building-docs-fails-due-to-missing-pandoc
+          pip install -r doc/pages/example_notebooks/requirements.txt # load requrements for notebooks
```
lukasmerten, line 33: requirements

lukasmerten:
Hi @JulienDoerner 
Looks in general good to me and is for sure a good approach to simplify the creation of the documentation and coverage report. 
Also testing the notebooks is a good idea, from my pint of view. However, the ne new test is failing. Is that expected or went something wrong?

JulienDoerner:
hey @lukasmerten, 
The failing test is caused by the notebook `secondaries/photons.ipynb`. In my last execution of the test before the PR this test passed. I guess this is related to the problems described in #456. I guess we could deactivate this notebook until the reason for the failing test has been figured out.

lukasmerten:
If re-running the test fixes the problem, which seems to be the case, we don't need to exclude the test. I will take a closer look at the doc and coverage action in the next days.

lukasmerten:
From my point of view we can merge this, when the typo is fixed. 

I'll approve it now, but let's wait another day or so for additional comments.

lukasmerten (approved):

PR #465, opened 2024-01-30
Implementation of a new EBL model from Finke et al. 2022.

Yannick-Schmidt:
This PR adds the recent model from [Finke et al. 2022](https://arxiv.org/abs/2210.01157) to the existing selection of custom photon fields. It is related to the [corresponding pull request](https://github.com/CRPropa/CRPropa3-data/pull/22) in the CRPropa-data repository.

The implementation includes the ...
- ... general creation of a new class, which simply inherits rom the `TabularPhotonField` base class (in the `PhotonBackground.h` file).
- ... adding of the new model to the interaction test files (`testInteraction.cpp`).


rafaelab:
For the record, I've been waiting for the changes in the CRPropa3-data repo to check and merge both together.

lukasmerten:
Hi @rafaelab, @Yannick-Schmidt, @JulienDoerner 
I updated the download link, after uploading the new tarball to sciebo. Furthermore, I resolved some merge conflicts and updated the changelog. Pleas let Yannick and me know, if there is anything else taht needs to be done for the new EBL modek to be implemeneted.

test/testInteraction.cpp
```diff
@@ -40,6 +40,8 @@ TEST(ElectronPairProduction, allBackgrounds) {
 	epp.setPhotonField(irb);
 	irb = new IRB_Stecker16_lower();
 	epp.setPhotonField(irb);
+    irb = new IRB_Finke22();
```
JulienDoerner, line 43: Whitespace

test/testInteraction.cpp
```diff
@@ -718,6 +724,8 @@ TEST(EMPairProduction, allBackgrounds) {
 	em.setPhotonField(ebl);
 	ebl = new IRB_Stecker16_lower();
 	em.setPhotonField(ebl);
+	ebl = new IRB_Finke22();
```
JulienDoerner, line 727: Why is it called EBL here and IRB else?

JulienDoerner (approved):
I do not see any serious issue here, only some small style comments. 
I will aprove it anyway so that it can be merged afterwards.

test/testInteraction.cpp
lukasmerten: I think that depends on who implemented the test in the first place. Could be cleaned up in the future.

test/testInteraction.cpp
lukasmerten: I am not creating a new commit just to fix the whitepace, but we should probably do this at some point in a consistent way for all files.

Issue #448, opened 2023-11-14
Segmentation fault with lens.transformModelVector


ryohiguchi:
I'm trying the same code as in the documentation page: https://crpropa.github.io/CRPropa3/pages/example_notebooks/galactic_lensing/lensing_maps.v4.html

The example code works until the plot of unlensed map, but transformModelVector command outputs:
zsh: segmentation fault

The information of my current system is as follows:
 - Operating system: IOS 14.1
 - MacBook Pro
 - Python version: 3.7.17
 - Swig version: 3.0.12
 - Cmake version: 3.27.7

I also read the similar issue (https://github.com/CRPropa/CRPropa3/issues/397) and tested changing some versions of python & swig and also CRPropa 3.1.7, but  the results are same.

lukasmerten:
Hi @ryohiguchi 
This is, as you have seen, a known problem. But to my knowledge nobody is currently working on it. In case you you find a fix we would be happy to include it.

ryohiguchi:
It was successful when I change the numpy version==1.18.0
The problem seems to be that they are calling a deprecated numpy function PyArray_GetArrayParamsFromObject
https://numpy.org/devdocs/release/1.19.0-notes.html#deprecation-of-probably-unused-c-api-functions

lukasmerten:
@ryohiguchi can you pleas check if #466 fixed this issue.

potassium-chloride:
I have the same issue.

I checked new version and the problem still exists:

**System Information:**
 - Operating system: Linux Mint 20 Cinnamon
 - Laptop
 - Python version: 3.7.6 (default, Jan  8 2020, 19:59:22), Anaconda
 - NumPy version: 1.21.6
 - Swig version: 4.0.1
 - Cmake version: 3.16.3
 
 Problem disappears on NumPy version==1.18.0 as mentioned @ryohiguchi. Now time I am using conda virtual env with old version of NumPy.
 
I've checked [difference](https://github.com/CRPropa/CRPropa3/pull/466/files) of last pull request and code of python/4_lens.i was a little bit refactored but its logic is absolutely same.
 
[...]

PR #466, opened 2024-02-04
Fix problems with Python, improve installation, and improve tests

rafaelab:
This PR supersedes #457.
It mainly improves the handling of python, but it also contains many other improvements/fixes:
* Use CMake's built-in Python finder and remove CRPropa's Python finder.
* Remove all references to Python2 and completely drops support
* Fix a bug common to several tests when using clang (gtest cannot correctly trace exceptions leading to failures when calling `EXPECT_THROW`)
* Fix a bug in `testInteraction` related to EMPairProduction, in OS X.
* Possible solution the long-standing issue related to the magnetic lenses and python
* Minor improvements in workflow
* Apply some code style convention to *.i files and CMake
* Improve code style in tests

One of the OSX tests (`testPythonExtension.py`) is expected to fail because of a problem applying NumPy functions to `Vector3d` in one of the tests (`testArrayInterface`). This does not happen, for example, in Ubuntu.
Even if this particular test fails, at least it is known and doesn't seem to affect CRPropa in any significant way.


rafaelab:
I'm puzzle by the fact that the crpropa-example-test is passing in my branch and not here.
That is very weird.

JulienDoerner:
@rafaelab: The failing test is related to the notebook `secondaries\photons.ipynb`, but it got two different erros in the two executions. I will try to run it a thrid time.
We had the same problem while integrating these tests (see PR #461). In this case re-runing the test solved the problem, but as it still gives wired errors I would suggest to deactivate this notebook (but not here, I would suggest to open an extra PR). 

Regarding this PR: I will try to have a look unitll end of the week, but I can only test everything on a linux machine. Maybe one of our mac users (@sophieaerdker, @LeanderSchlegel, @reichherzerp) could do some testing.

JulienDoerner:
For the issue with the failing example test see #467.

.github/workflows/testing_OSX.yml
```diff
@@ -29,7 +29,7 @@ jobs:
        run: |
         mkdir build
         cd build
-        cmake .. -DENABLE_PYTHON=True -DPYTHON_EXECUTABLE=${{ matrix.config.py }} -DENABLE_TESTING=On -DENABLE_SWIG_BUILTIN=${{ matrix.config.swig_builtin }} -DSIMD_EXTENSIONS=avx
+        cmake .. -DENABLE_PYTHON=True  -DENABLE_TESTING=On  -DENABLE_SWIG_BUILTIN=${{ matrix.config.swig_builtin }} -DSIMD_EXTENSIONS="none"
```
JulienDoerner, line 32: do we realy want to test the SIMD case "none"? In this case we decrese the coverage of the code.

.github/workflows/test_examples.yml
```diff
@@ -58,7 +58,7 @@ jobs:
           done
       - name: run all python scripts
         env: 
-          PYTHONPATH: "$/home/runner/.local/lib/python3.8/site-packages/"
+          PYTHONPATH: "/home/runner/.local/"
```
JulienDoerner, line 61: this is a change in installation behaviour (see also in CMakeList)

CMakeLists.txt
```diff
@@ -451,14 +444,39 @@ endif(BUILD_DOC)
 # Python
 # ----------------------------------------------------------------------------
 option(ENABLE_PYTHON "Create python library via SWIG" ON)
-find_package(PythonInterp)
-find_package(PythonLibs)
 
-if(ENABLE_PYTHON AND PYTHONLIBS_FOUND)
+find_package(Python 3.0
+  REQUIRED COMPONENTS Interpreter Development
+  OPTIONAL_COMPONENTS NumPy)
```
JulienDoerner, line 450: Here numpy is only an optional component. 

I was not able to complie CRPropa with Python but without numpy. 
And on the other hand we state on our installation documentation, that numpy is required. Therefore I would suggest to set it to the REQUIRED COMPONENTS. 

If we are strictly with the numpy dependence a lot of code checking if numpy is available or not can be skiped, and it gets better readable

CMakeLists.txt
```diff
@@ -520,14 +521,14 @@ if(ENABLE_PYTHON AND PYTHONLIBS_FOUND)
...
   set_target_properties(crpropa-swig PROPERTIES COMPILE_FLAGS "-w")
   set_target_properties(crpropa-swig PROPERTIES OUTPUT_NAME "_crpropa")
-  target_link_libraries(crpropa-swig crpropa ${PYTHON_LIBRARIES})
+  target_link_libraries(crpropa-swig crpropa ${Python_LIBRARIES} ${Python_LIBRARY})
   add_dependencies(crpropa-swig crpropa-swig-wrapper)
 
-  install(DIRECTORY "${CMAKE_SOURCE_DIR}/python/crpropa" DESTINATION "${PYTHON_SITE_PACKAGES}")
-  install(FILES "${CMAKE_CURRENT_BINARY_DIR}/crpropa.py" DESTINATION "${PYTHON_SITE_PACKAGES}/crpropa")
-  install(TARGETS crpropa-swig LIBRARY DESTINATION "${PYTHON_SITE_PACKAGES}/crpropa")
+  install(DIRECTORY "${CMAKE_SOURCE_DIR}/python/crpropa" DESTINATION ${CMAKE_INSTALL_PREFIX})
```
JulienDoerner, line 527: This is a drastical change of installation behaviour. 
In this case the python package is not stored in the same directory as the other packages of an virtual environment. This should be changed to the old behaviour.

python/2_headers.i
```diff
@@ -78,84 +70,70 @@ using namespace crpropa;   // for usage of namespace in header files, necessary
...
-  {
+%extend crpropa::Vector3 {
+  size_t __len__() {
     return 3;
   }
 
-  PyObject* __array__()
-  {
+  PyObject* __array__() {
     npy_intp shape[1];
```
JulienDoerner, line 105: This part is for example not covered by any check if numpy is available or not

python/4_lens.i
```diff
@@ -9,10 +9,10 @@
 %template(DoubleVector) std::vector<double>;
 
 %{
-#include "crpropa/magneticLens/ModelMatrix.h"
-#include "crpropa/magneticLens/Pixelization.h"
-#include "crpropa/magneticLens/MagneticLens.h"
-#include "crpropa/magneticLens/ParticleMapsContainer.h"
+  #include "crpropa/magneticLens/ModelMatrix.h"
+  #include "crpropa/magneticLens/Pixelization.h"
+  #include "crpropa/magneticLens/MagneticLens.h"
```
JulienDoerner, line 14: alphabetic oder of modules

python/4_lens.i
```diff
@@ -83,36 +79,30 @@
...
-                PyObject *galacticLatitudes,
-                PyObject *weights)
-  {
+                        PyObject *energies,
+                        PyObject *galacticLongitudes,
+                        PyObject *galacticLatitudes,
+                        PyObject *weights) {
     //ToDo: Handle strided arrays
 
     //ToDo: Check that input objects are arrays  PyArray_Check
```
JulienDoerner, line 88: Is this still an open ToDo? If yes we should open an issue for that. 
In any case we should not write ToDo statements in the comments.

python/4_lens.i
```diff
@@ -152,114 +137,95 @@
...
+    PyList_SET_ITEM(returnList, 0, (PyObject*) oId);
+    PyList_SET_ITEM(returnList, 1, (PyObject*) oEnergy);
+    PyList_SET_ITEM(returnList, 2, (PyObject*) oLon);
+    PyList_SET_ITEM(returnList, 3, (PyObject*) oLat);
+    
+    return returnList;
   }
 
 };
 #else // with numpy
```
JulienDoerner, line 206: this part can be droped if numpy is required for installation.

test/testInteraction.cpp
```diff
@@ -839,17 +839,17 @@ TEST(EMDoublePairProduction, limitNextStep) {
 
 TEST(EMDoublePairProduction, secondaries) {
 	// Test if secondaries are correctly produced.
-	ref_ptr<PhotonField> CMB_instance = new CMB();
-	ref_ptr<PhotonField> IRB = new IRB_Gilmore12();
-	ref_ptr<PhotonField> URB = new URB_Nitu21();
-	EMDoublePairProduction m(CMB_instance);
+	ref_ptr<PhotonField> cmb = new CMB();
+	ref_ptr<PhotonField> irb = new IRB_Gilmore12();
+	ref_ptr<PhotonField> urb = new URB_Protheroe96();
```
JulienDoerner, line 844: Why is the URB model changed. If we want to test the URB_Protheroe as well I would suggest to add it as a new target.

test/testInteraction.cpp
```diff
@@ -944,23 +943,24 @@ TEST(EMTripletPairProduction, limitNextStep) {
 
 TEST(EMTripletPairProduction, secondaries) {
 	// Test if secondaries are correctly produced.
-	ref_ptr<PhotonField> CMB_instance = new CMB();
-	ref_ptr<PhotonField> IRB = new IRB_Gilmore12();
-	ref_ptr<PhotonField> URB = new URB_Nitu21();
-	EMTripletPairProduction m(CMB_instance);
+	ref_ptr<PhotonField> cmb = new CMB();
+	ref_ptr<PhotonField> irb = new IRB_Gilmore12();
+	ref_ptr<PhotonField> urb = new URB_Protheroe96();
```
JulienDoerner, line 948: see above

test/testTurbulentField.cpp
```diff
@@ -91,25 +91,27 @@ TEST(testVectorFieldGrid, Turbulence_seed) {
...
-			std::runtime_error);
-	// lMin > lMax
-	EXPECT_THROW(initTurbulence(grid, brms, 8.1 * spacing, 8 * spacing),
-			std::runtime_error);
-	// lMax too large
-	EXPECT_THROW(initTurbulence(grid, brms, 2 * spacing, 65 * spacing),
-			std::runtime_error);
-}
+#ifndef CRPROPA_TESTS_SKIP_EXCEPTIONS
+	TEST(testVectorFieldGrid, turbulence_Exceptions) {
```
JulienDoerner, line 95: In the other tests there was no intend after the compiler statements. This should be consistent.

.github/workflows/testing_OSX.yml
rafaelab: Tests were failing with SIMD and it is a simply an optimisation, so I think it is fine to have it this way. Unless the underlying cause for the test to fail has been identified.

python/4_lens.i
rafaelab: I actually don't know whose to-do list this is. (Not mine).
I will just remove it then.

CMakeLists.txt
rafaelab: I will have to check this again then. The reason for the change is that some tests were failing.

test/testInteraction.cpp
rafaelab: This was a copy-paste from before. I didn't pay attention to which background that was. 
I will revert it back to Nitu21.

CMakeLists.txt
rafaelab: At a second glance, after running some tests on my machine, the previous version was also problematic. We should not force installation at `Python_SITE_PACKAGES`. 
That is the reason why it was failing and why my `site-packages` folder was getting polluted against my will.
Do you have a more concrete suggestion to work around this, @JulienDoerner ?

.github/workflows/testing_OSX.yml
JulienDoerner: Okay this sounds like a larger project to trace down this issue. 
For me it is fine, to disable SIMD in the OSX test, but I would suggest to open an issue to document this (and add the label "wont fix").

rafaelab:
Thanks for the review, @JulienDoerner 
I think I've addressed most of the issues you've raised so far and after a long day of attempts I managed to get `crpropa-example-test` to run, to some extent.

I tried to make the installation default to `Python_SITE_PACKAGES` (actually, `Python_SITELIB`) only if the user is not trying to do a local installation using `CMAKE_INSTALL_PREFIX`. This way, we don't affect the previous behaviour too much, and people like myself who prefer to choose the installation paths are satisfied. What do you think?

I've made one further change: I commented out the test that was leading to the problem in the OSX tests. Unless you have any objections, I'd suggest we keep it that way until we figure out how to solve the problem. And it is better not to delete the corresponding lines yet, since they might be useful in the future.

lukasmerten:
It's more or less working for me.

However, I would also prefer an option to specify the CMAKE_INSTALL_PREFIX and Python_INSTALL_PACKAGE_DIR independently of each other. I need to set the CMAKE_INSTALL_PREFIX to something, which is different from the default /usr/local as we do not have write access to that location on our university machines. This leads in the current PR to installing the python module in CMAKE_INSTALL_PREFIX/crpropa which would require manually adapting the python PATH for my virtual envs...

JulienDoerner:
Hey @rafaelab, 

I think adding this variable `Python_INSTALL_PACKAGE_DIR` is a good idea. But for me I could not change (or see) it using ccmake. It only appers in the printed log. 

And also this does not reproduce the old installation behaviour. In this case (given PREFIX) all installation parts are stored on the same level (python package `PREFIX/crpropa`, shared data `PREFIX/share/crpropa`, lib `PREFIX/lib/libcrpropa.so` and include `PREFIX\include\CRPropa.h`). 
I would think the best way would be to allow for the user defined path but default it to `PREFIX/lib/python3.X/crpropa`. In this case the installation procedure would stay the same as in our installation documentation.

rafaelab:
Thanks for the feedback, @JulienDoerner and @lukasmerten.
Now the flag `Python_INSTALL_PACKAGE_DIR` is completely independent from `CMAKE_INSTALL_PREFIX`. 

The remaining issue is what @JulienDoerner said about `$PREFIX/lib/python3.X/crpropa`. This is the internal structure of python packages. But in this case, the user explicitly wants to control all paths, so I fail to see a reason to keep it this way.  Why add so many unnecessary layers when we can just have `$Python_INSTALL_PACKAGE_DIR` directly (at the level of `$PREFIX` or perhaps `$Python_INSTALL_PACKAGE_DIR/python`?

JulienDoerner:
hey @rafaelab, 

I think the change is good, as it is. My point for the `python3.X/site-packages/` was only about the default value for the `$Python_INSTALL_PACKAGE_DIR`. But I agree that installing it directly in this level is the best solution.

lukasmerten:
Thanks for your effort @rafaelab. 
Seems to work for me and I will approve this later.

Just for my records:
You can set the flag for the python installation path with 
($~/build) cmake -DPython_INSTALL_PACKAGE_DIR=<path-to-python-installation>..

Just for curiosity:
Is there a way to add Python_INSTALL_PACKAGE_DIR to the parameters that are listed when I run ccmake?

lukasmerten (approved):

rafaelab:
> Just for my records: You can set the flag for the python installation path with ($~/build) cmake -DPython_INSTALL_PACKAGE_DIR=..
> 
> Just for curiosity: Is there a way to add Python_INSTALL_PACKAGE_DIR to the parameters that are listed when I run ccmake?

No idea, @lukasmerten. I did add a status message with the value of this flag, but I haven't invested too much time in adding it to the parameter. I thought it would show up there automatically.
In any case, I will keep this on my to-do list for next time I make general changes to CRPropa.

JulienDoerner (approved):

python/4_lens.i
```diff
@@ -33,25 +33,22 @@
...
+
+%extend crpropa::MagneticLens {
+  PyObject * transformModelVector_numpyArray(PyObject *input, double rigidity) {
     PyArrayObject *arr = NULL;
     PyArray_Descr *dtype = NULL;
     int ndim = 0;
     npy_intp dims[NPY_MAXDIMS];
-    if (PyArray_GetArrayParamsFromObject(input, NULL, 1, &dtype, &ndim, dims, &arr, NULL) < 0)
-    {
+    if (PyArray_GetArrayParamsFromObject(input, NULL, 1, &dtype, &ndim, dims, &arr, NULL) < 0) {
```
potassium-chloride, line 47: Something goes wrong on this line. See issue #448 . Please, check behavior of this code on NumPy version 1.21.6.

PR #471, opened 2024-02-22
updated doc for new Python interface via cmake

rafaelab:
Minor changes to the documentation to reflect the recent updates.

### DIFF
--- a/doc/pages/Installation.md
+++ b/doc/pages/Installation.md
@@ -12,7 +12,7 @@ git clone https://github.com/CRPropa/CRPropa3.git
 + Fortran Compiler: to compile SOPHIA
 
 Optionally CRPropa can be compiled with the following dependencies to enable certain functionality.
-+ Python, NumPy, and SWIG: to use CRPropa from python (tested for > Python 3.0 and > SWIG 3.0.4)
++ Python, NumPy, and SWIG: to use CRPropa from python (tested for > Python 3.0 and > SWIG 4.0.2)
 + FFTW3: for turbulent magnetic field grids (FFTW3 with single precision is needed)
 + Gadget: magnetic fields for large scale structure data
 + OpenMP: for shared memory parallelization
--- a/doc/pages/Installation.md
+++ b/doc/pages/Installation.md
@@ -181,8 +181,13 @@ cmake -DENABLE_PYTHON=ON ..
   -DPython_INCLUDE_DIRS=<path_to_folder_containing_Python.h>
   -DPython_LIBRARY=<path_to_file>/libpython<version_tag>.so
   ```
-  Note that in systems running OSX, the extension .so should be replaced by .dylib.
-  For further details, see [FindPython.cmake](https://cmake.org/cmake/help/latest/module/FindPython.html#module:FindPython).
+Note that in systems running OSX, the extension .so should be replaced by .dylib. 
+In addition, The path where the CRPropa python module is installed can be specified with the flag:
+```
+-DPython_INSTALL_PACKAGE_DIR=<path_to_folder>
+```
+For further details, see [FindPython.cmake](https://cmake.org/cmake/help/latest/module/FindPython.html#module:FindPython).
+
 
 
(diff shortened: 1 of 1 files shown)


doc/pages/Installation.md
```diff
@@ -10,10 +10,9 @@ git clone https://github.com/CRPropa/CRPropa3.git
 ## Prerequisites
 + C++ Compiler with C++11 support (gcc, clang and icc are known to work)
 + Fortran Compiler: to compile SOPHIA
-+ numpy: for scientific computations
 
 Optionally CRPropa can be compiled with the following dependencies to enable certain functionality.
-+ Python and SWIG: to use CRPropa from python (tested for > Python 2.7 and > SWIG 3.0.4)
++ Python, NumPy, and SWIG: to use CRPropa from python (tested for > Python 3.0 and > SWIG 3.0.4)
```
lukasmerten, line 15: SWIG > 4.0.2 ?

doc/pages/Installation.md
```diff
@@ -178,11 +177,13 @@ cmake -DENABLE_PYTHON=ON ..
...
 + Quite often there are multiple Python versions installed in a system. This is likely the cause of many (if not most) of the installation problems related to Python. To prevent conflicts among them, one can explicitly refer to the Python version to be used. Example:
   ```
-  -DCMAKE_PYTHON_EXECUTABLE=/usr/bin/python
-  -DCMAKE_PYTHON_INCLUDE_DIR=<path_to_folder_containing_Python.h>
-  -DCMAKE_PYTHON_LIBRARY=<path_to_file>/libpython<version_tag>.so
+  -DPython_EXECUTABLE=/usr/bin/python
+  -DPython_INCLUDE_DIRS=<path_to_folder_containing_Python.h>
+  -DPython_LIBRARY=<path_to_file>/libpython<version_tag>.so
   ```
   Note that in systems running OSX, the extension .so should be replaced by .dylib.
```
lukasmerten, line 184: Should we also add something like the following lines?

"The path to the location wehre the crpropa python module will be installed, can be specified with:
`-DPython_INSTALL_PACKAGE_DIR=`"

doc/pages/Installation.md
```diff
@@ -201,41 +202,13 @@ For Fedora/CentOS/RHEL the required packages to build CRPropa:
   ```
 In case of CentOS/RHEL 7, the SWIG version is too old and has to be built from source.
 
-
 ### Mac OS X
```
lukasmerten, line 205: I remember that is was quite tricky to install crpropa on some Mac configuration, so I would suggest to keep the longer explanation for a little longer.

But maybe @reichherzerp, @sophieaerdker, or @LeanderSchlegel can comment on that.

rafaelab: I adjusted the previous version now, @lukasmerten .
I think the reason for that complicated installation had something to do with AppleClang/LLVM and OpenMP, that is why all the flags were required. I think the newer versions of the compilers handle OpenMP much better.

JulienDoerner:
Hey @rafaelab, 
this PR looks good for me. I would have only one minor question. We state "testet for > Python 3.0 and > SWIG 4.0.2", but SWIG itself requires at least Python 3.2. So my question is, which version of Python did we realy tested? The oldest version I ever used was 3.7.

rafaelab:
> Hey @rafaelab, this PR looks good for me. I would have only one minor question. We state "testet for > Python 3.0 and > SWIG 4.0.2", but SWIG itself requires at least Python 3.2. So my question is, which version of Python did we realy tested? The oldest version I ever used was 3.7.

Good point, @JulienDoerner. I said 3.0 just because we don't  support Python 2 anymore.
I didn't actually test versions prior to 3.7.
What is your suggestion? Should we say 3.2, or 3.7?

lukasmerten:
@rafaelab: Just put whatever is the oldest python version that is used in one of the actions. I guess that is python 3.8 in the ubuntu-20 runner.

lukasmerten (approved):

Issue #474, opened 2024-03-01
Improved Photon Field Sampling


ehlertdo:
Hi CRPropa team,
### Background
I was recently trying to use CRPropa with custom, tabulated photon fields relevant for AGN environments. I have processed the fields with CRPropa3-data to get the right format to use with CRPropa and, excluding some minor issues with converting the units correctly, this works generally fine. To check that everything works correctly I also created a tabulated CMB (simple black body) to compare with crp.CMB(). For programmatic reasons my maximum tabulated photon energy (4e10 eV) is very large compared to the typical CMB photon temperature (CRPropa current, epsMax_CMB=0.1 eV).
### Problem
I noticed that in this case there is a problem with the sampling of photon energies when running CRPropa (e.g. for photopion) with fields where the maximum tabulated energy is much larger than any energies where the photon density is non-negligible.
This gives the following error `error: no photon found in sampleEps, please make sure that photon field provides photons for the interaction by adapting the energy range of the tabulated photon field.`. By truncating the tabulated CMB files at lower energies, e.g. 0.1 eV as is the current default, the error can be avoided. However, ideally CRPropa would work correctly no matter the tabulated range of energies.
### Solution
After some digging I noticed that CRPropa appears to use a linear sampling for the photon field energies when processing the interactions. At present, only the maximum interaction probability can be sampled logarithmically (see #368). The relevant line of code is in `PhotopionProduction::sampleEps(...)` where

`double eps = epsMin + random.rand() * (epsMax - epsMin);`

I have verified that by replacing this with a logarithmic sampling, using the inverse CDF of the [reciprocal distribution](https://en.wikipedia.org/wiki/Reciprocal_distribution), 

`double eps = epsMin * pow(epsMax /epsMin, random.rand());`

the sampling works correctly even for the scenario mentioned above. To me it appears that this sampling is more robust and should be generally preferred. Perhaps a flag similar to `PhotoPionProduction::setSampleLog(bool b)` should be available to enable this behavior or the same flag could be used for both purposes.

### Photon Field Density & Sampling
[...]

PR #476, opened 2024-03-07
Implement optional logarithmic sampling of the photon energies

ehlertdo:
Hey,
this is the follow-up to #474 . With this change it is possible to sample photons from tabulated photon fields where the maximum tabulated energy is much larger than the highest energies where the field density is non-zero.
I have attached a minimal CRPropa 1D simulation script that illustrates a scenario where the current version of CRPropa fails to find photons and stops with `error: no photon found in sampleEps, please make sure that photon field provides photons for the interaction by adapting the energy range of the tabulated photon field.`
The files for such a tabulated (CMB-like) photon field are also included and must be placed in the share folder of the local CRPropa installation.
Let me know what you think!
[supplementary_files.tar.gz](https://github.com/CRPropa/CRPropa3/files/14522020/supplementary_files.tar.gz)

### DIFF
--- a/src/module/PhotoPionProduction.cpp
+++ b/src/module/PhotoPionProduction.cpp
@@ -422,7 +422,11 @@ double PhotoPionProduction::sampleEps(bool onProton, double E, double z) const {
 
 	Random &random = Random::instance();
 	for (int i = 0; i < 1000000; i++) {
-		double eps = epsMin + random.rand() * (epsMax - epsMin);
+		double eps;
+		if (sampleLog){
+			eps = epsMin * pow(epsMax / epsMin, random.rand());
+		} else
+			eps = epsMin + random.rand() * (epsMax - epsMin);
 		double pEps = probEps(eps, onProton, Ein, z);
 		if (random.rand() * pEpsMax < pEps)
 			return eps * eV;


JulienDoerner:
I guess an other way of solving the problem could be a change in the maximum photon energy  https://github.com/CRPropa/CRPropa3/blob/1ffe8351ae2f6ddaae847be2ee44243267a776a1/src/PhotonBackground.cpp#L67

I would suggest to add a threshold photon density (maybe eaven 0) to evaluate the maximum energy. In this case the region with $n = 0$ would not be sampled.

ehlertdo:
Good point about the difference in the sampled distribution. I had only checked that sampling from the build-in CMB and my tabulated version yields the same photon distribution after the modification. I will try to think about a solution.
In general, for the rejection sampling one can define an arbitrary proposal function which should ideally be similar to the true distribution of photons. One could consider providing a range of possible proposal distributions to choose from depending on the type of photon field, as opposed to just a flat prior. While this could speed up the sampling it would add a lot of complexity to the code.

ehlertdo:
I implemented a photon sampling with a decreasing-power-law proposal function. Now the sampled distribution for interactions with a (I think) $10^{21}~\mathrm{eV}$ proton agrees with the default implementation in CRPropa. However, because each sample requires the evaluation of the inverse and regular CDF of the proposal function the sampling time increases by a factor of 5-10 (tested for 10k photons). Perhaps there is somem room for improvement here. The upside is that it works even in the scenario with extremely large maximum tabulated photon energies where the normal implementation fails.
hist_sampled_eps

PR #478, opened 2024-03-14
Source sampling

JulienDoerner:
The sampling of the position in the SNR and pulsar distribution is done in cylindrical coordinates. The current approach misses a $r$ term for the radial componet. The differential number of sampled positions should be given by $dN = f(R, z) dV = f(R, z) \ R dR\ dZ \ d\phi$. 

Additionaly the PR reduces the amount computation by skipping the normalisation of the curves. For the MC-rejection sampling only the knowleadge of the overall maximum is needed (once calculated) and not a normalized curve.

### DIFF
--- a/src/Source.cpp
+++ b/src/Source.cpp
@@ -416,7 +416,7 @@ double SourceSNRDistribution::fz(double z) const{
 }
 
 double SourceSNRDistribution::getAlpha() const {
-	return alpha - 1; 
+	return alpha - 1;  // -1 to account for the R-term in the volume element dV = R * dR * dphi * dz
 }
 
 double SourceSNRDistribution::getBeta() const {


src/Source.cpp
```diff
@@ -414,7 +416,7 @@ double SourceSNRDistribution::fz(double z) const{
 }
 
 double SourceSNRDistribution::getAlpha() const {
-	return alpha;
+	return alpha - 1;
```
lukasmerten, line 419: Can you add a comment here, too? Just to make sure that everybody understands why alpha is used differently than what it was set to.

lukasmerten:
Looks generally good to me. I agree that we missed that additional factor from the volume element before.

src/Source.cpp
JulienDoerner: I added the comment

lukasmerten (approved):

lukasmerten:
Thanks, @JulienDoerner. Let's for some more comments and merge otherwise next week.

rafaelab:
Looks good to me, @JulienDoerner .

PR #481, opened 2024-04-09
Add coverage report to documentation

JulienDoerner:
With this PR the coverage report is added to the documentation. 
The Github action is adopted to copy the report into the documentation part and a link to the files is added. 

In case of a manual build of the documentation the coverage report has to be copied by hand to the right location. In the build folder it is done by:

    cp -r coverageReport doc/pages/coverageReport


.github/workflows/create_documentation.yml
```diff
@@ -55,6 +55,7 @@ jobs:
         run: |
           cd build
           make doc
+          cp -r coverageReport doc/pages/coverageReport
```
lukasmerten, line 58: If we copy the coverage report into the documentation: Do we really need a separate upload of the coverage report (line 53 and 66)

JulienDoerner: I guess it can be usefull in the case we only want to check the new coverage report for a PR. But I guess in this case a seperate action would be more helpfull.

lukasmerten:
If we are on the automatic creation of the documentaion, we could also think about automatic deployment to gh-pages.

`- name: deploy documentation #deploys the documentation to the gh-pages branch`
`        uses: JamesIves/github-pages-deploy-action@v4`
`        with:`
`          folder: ./build/doc`

Following this description here: https://github.com/JamesIves/github-pages-deploy-action

In case this seems to be useful, I'll can either provide a new PR or just add to this one.

JulienDoerner:
I will add the deployment in this PR. For the testing phase I will convert it to draft and notify you when it is ready.

JulienDoerner:
I changed the workflows accordingly. 
Now the steps are seperated in three different task 
1. The `create_documentation` builds the docu and uploads the artifact. The coverage report is only included in the documentation and not seperatly stored. 
2. The `create_coverage_report` builds the coverage report and uploads it as an artifact. This can be used to check the changes with new PRs. 
3. The `deploy_new_documentation` workflow builds the new documentation and push it automaticly to the `gh-pages` branch. This works in my fork, but we might have to check the permissions. It is configured to run in all cases, where we push to the master branch. This should be done as the coverage report and the API information are changed with all changes on the master branch.

lukasmerten (approved):

PR #484, opened 2024-04-22
Possible fix for wrong secondary masses of photons and neutrinos

### LINKED ISSUE #483: Small technical bug leading to wrong masses for secondary photons or neutrinos
Dear all,

we noticed a small technical bug in the set of the particle rest-mass for secondaries, in the case of non-nuclei secondaries like photons or neutrinos they inherit the primary particles mass, while for nuclei, electrons or positrons the mass is set as expected. It should probably not affect the physics in the other modules and can be fixed easily, I discussed it with @lukasmerten and @JulienDoerner and would open a PR to fix it.

The bug can be reproduced with the minimal example creating an electron and adding a secondary photon.
`c = Candidate(11)`
`c.addSecondary(22,1*EeV)`
`print (c.secondaries[0].current.getMass())`
`9.10938291e-31`

Instead of the rest-mass 0, the photon inherits the electrons rest-mass. The bug seems to be created when addSecondary() after inheriting the current-state, calls setId() from ParticleState.cpp, where for other than nuclei or electron/positrons, the mass can not be set.

Cheers,
Leander

LeanderSchlegel:
Dear all,

with this PR we try to fix #483. @lukasmerten, @JulienDoerner and me discussed it and introduced a new function particleMass(int id) in ParticleMass.cpp to fix the bug by calling it in setId(int newId) in ParticleState.cpp. It gives all particles other than nuclei or electrons/positrons a zero rest-mass and could possibly be extended for other particles when needed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,9 +6,9 @@
  * Fixed wrong mass inheritance for secondaries other than nuclei or electron/positron
 
 ### New features:
+ * Added new backwards-compatible function particleMass that returns particle mass also for non-nuclei 
 
 ### Interface changes:
- * Added new backwards-compatible function particleMass that returns particle mass also for non-nuclei 
 
 ### Features that are deprecated and will be removed after this release
  * EBL model from Finke et al. 2022
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,10 +3,12 @@
 ### Bug fixes:
  * Fixed sign for exponential decay of magn. field strength with Galactic height in LogarithmicSpiralField 
  * Fixed r term in source distribution for SNR and Pulsar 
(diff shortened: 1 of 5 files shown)


JulienDoerner (changes requested):
The solution looks good to me. 
I have only smaller style comments. 

And maybe we should include a test in the `testCore` part explicitly checking the secondary mass.

lukasmerten (approved):

lukasmerten:
Looks good to me. Let's wait for more comments otherwise I will merge in the next days.

LeanderSchlegel:
Thank you very much for your feedback! 
As suggested by Julien and in discussion with him, I added two small tests in `testCore,` one extending the test of addSecondary to also check their masses and one new test, covering the new function `particleMass(int id)`.

JulienDoerner (approved):

PR #485, opened 2024-04-29
Loading Grids with properties

JulienDoerner:
Dear all, 

this PR introduces the loading of grids with given properties in the header. 
The first implementation only allows property reading in the TXT files. 
An example of the properties in the header looks like this 
```
# GridProperties: Type Grid1f   origin: -4.92744e+20 -4.92744e+20 -1.44641e+19  gridsize: 512 512 16    spacing: 1.92855e+18 1.92855e+18 1.92855e+18    reflective: 0       clipVolume: 1   interpolation: TRILINEAR
```
all units are in SI base units (in this case m). The loading itself has the conversion variable as before. 
Loading a grid can now be done as 
``` 
grid = loadGrid3f("myGridFile.txt", 1 * gauss)
```

I also introduce the possibility to store the properties in the `dumpGridToTxt` function and additional tests for this grid reading/dumping. 

To have an example, I adjusted the notebook with the grid-based glactic gas densities to use the grid property reading. The files on sciebo (additional resources) are already updated (additional line with the properties in the right format).


include/crpropa/Grid.h
```diff
@@ -247,6 +263,21 @@ class Grid: public Referenced {
...
+
+	std::string getInterpolationTypeName() {
+		if (ipolType == TRILINEAR)
+			return "TRILINEAR";
+		if (ipolType == TRICUBIC)
+			return "TRICUBIC";
+		if (ipolType == NEAREST_NEIGHBOUR)
+			return "NEAREST_NEIGHBOUR";
+
+		return "NOT_UNDERSTOOD";
```
lukasmerten, line 278: Could you add a more meaningful error message. Something like "ipolType not well defined: should be either TRILINEAR, TRICUBIC, or NEAREST_NEIGHBOUR".

lukasmerten (changes requested):
Looks overall good to me. Just a few minor comments

include/crpropa/Grid.h
JulienDoerner: This is not meant to be an error message. It is the conversion from the data-type to a string for saving. 
To be consistent with the reading it has to be without spaces. 

The alternative would be raising a runtime error at this point. In this case the saving would not run at all.

lukasmerten (approved):

PR #486, opened 2024-05-02
Add names to omp critical statements

JanNiklasB:
Dear All,

with this pull request I want to add names to various `#pragma omp critical` statements.
If such a statement has no name, it blocks every other critical statement with no name, which may cause an increase in runtime.
For example, a shell output should only block other shell outputs and not file outputs.
Furthermore, if not accessed through a python script but directly through c++, critical statements from the users files may also block critical statements in CRPropa.

### DIFF
--- a/include/crpropa/Referenced.h
+++ b/include/crpropa/Referenced.h
@@ -47,7 +47,7 @@ public:
 #elif defined(__GNUC__)
 		newRef = __sync_add_and_fetch(&_referenceCount, 1);
 #else
-		#pragma omp critical
+		#pragma omp critical(newRef)
 		{newRef = _referenceCount++;}
 #endif
 		return newRef;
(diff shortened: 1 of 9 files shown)


lukasmerten:
Hi @JanNiklasB 
Thanks for contributing to CRPropa. I did not see any problems with your suggested improvement. 
If you apply the two small updates suggested above to the code, I will approve it and merge afterwards.

JanNiklasB:
Hey,
I applied the two small updates, now the typo is fixed and the name for the HDFOutput is more generalized.
Sorry that this took so long, I am new to github and did not saw or could not find the "two small updates" you wrote about
and forgot about it untill now.

lukasmerten:
Hi @JanNiklasB 
That was not your fault. I missed to save my comments properly. They were showing up for me in the conversation but not for anyone else. So do not worry: Obviously you can make mistakes even after several years with GitHub.
So now everything looks fine from my sight. If I do not hear any objections, I will merge this in the next days.

PR #492, opened 2024-07-03
Fixes for example test and PlaneWaveTurbulence

JulienDoerner:
Dear all, 

this PR includes two small fixes: 

1. the example notebook test does not run with the new numpy 2.0.0 version. Therefore I fixed the version in the requirement file to 1.26.4
2. The PlaneWaveTurbulence with the `FastWave` option uses the function `std::align` for which the header `<memory>` is not included so far.

Best, 
Julien


lukasmerten (approved):

lukasmerten:
Looks like an obvious fix to me. If nobody objects, I will merge this tomorrow.

PR #502, opened 2024-08-28
Memory management in Variants

### LINKED ISSUE #493: Memory management in BreakingCondition
For context, I ran several simulations inside a loop, varying mass composition to see the energy spectrum at the Earth. I noticed that the RAM consumption was exceedingly high after some time. Further inspection revealed that RAM usage was steadily increasing during my 3D simulations, and the only way to free up that memory was by closing the program. 

After debugging, I think the problem is related to the breaking conditions I used, which are `MinimimEnergy` and `MaximumTrajectoryLength`. I came up with a minimum reproducible example by modifying this [1D example](https://crpropa.github.io/CRPropa3/pages/example_notebooks/sim1D/sim1D.html#Simulation-Setup) on the documentation.

**To Reproduce**
```
from crpropa import *

# simulation setup
sim = ModuleList()
sim.add( SimplePropagation(1*Mpc, 100*Mpc) )
sim.add( Redshift() )
sim.add( MaximumTrajectoryLength( 2000 * Mpc ) ) # Important part

# observer and output
obs = Observer()
obs.add( Observer1D() )
sim.add( obs )

# source
emissionDirection = Vector3d(-1, 0, 0) # Change to (1, 0, 0) to trigger MaximumTrajectoryLength
source = Source()
source.add( SourceUniform1D(1 * Mpc, 1000 * Mpc) )
source.add( SourceDirection( emissionDirection ) )
source.add( SourceRedshift1D() )

# power law spectrum with charge dependent maximum energy Z*100 EeV
# elements: H, He, N, Fe with equal abundances at constant energy per nucleon
composition = SourceComposition(1 * EeV, 10 * EeV, -3)
composition.add(1,  1,  1)  # H
composition.add(4,  2,  1)  # He-4
composition.add(14, 7,  1)  # N-14
composition.add(56, 26, 1)  # Fe-56
source.add( composition )

# run simulation
sim.setShowProgress(True)
sim.run(source, 100_000_000, True)
```

In this example, I noticed two things. First, the `Observer` object correctly manages inactive particles, since running this simulation as it is didn't increase my RAM usage significantly, as shown in the following `htop` command output: 
image

But if I trigger the breaking condition by changing the source emission direction to `Vector3d(1, 0, 0)`, the RAM usage increases and there is no obvious way for me to decrease it:
image

**Expected behavior**
Based on what I showed earlier, I expected the breaking conditions to behave similarly to the observer object. That is, freeing up 
particles once they meet the condition, thus making them inactive. 

**Additional context**
This is a problem for me because, at some point, I am using `ParticleCollector` to rerun simulations from saved files, but after some time the outputs of these extra simulations are like this:
image

Which gave me the hint that this was a memory-related problem.

JulienDoerner:
fixes #493: 

The `abstractConditions` added a property to the candidate on the rejection. Up to now it only adds the key `Rejected`. I added the module name as the default Value for the flag. 

Also the variant (in which this key is stored), does not have the right destructor. If it is deleted, the type of the variant was not passed to the `clear` function. Therefore no deletion of the variant was possible. This leaded to the bug in the memory consumption.

### DIFF
--- a/include/crpropa/Module.h
+++ b/include/crpropa/Module.h
@@ -59,7 +59,17 @@ public:
 	void setMakeAcceptedInactive(bool makeInactive);
 	void setRejectFlag(std::string key, std::string value);
 	void setAcceptFlag(std::string key, std::string value);
+
+	void printRejectFlag();
+	void printAcceptFlag();
 };
+
+class DeactivateModule: public AbstractCondition {
+	public: 
+		void process(Candidate *cand) const { reject(cand); }
+};
+
+
 } // namespace crpropa
 
 #endif /* CRPROPA_MODULE_H */
--- a/include/crpropa/Module.h
+++ b/include/crpropa/Module.h
@@ -64,7 +64,11 @@ public:
 	void printAcceptFlag();
 };
 
-class DeactivateModule: public AbstractCondition {
+/**
+ @class Deactivation
+ @brief Direct deactivation of the candidate. Can be used for debuging.
+*/
+class Deactivation: public AbstractCondition {
 	public: 
 		void process(Candidate *cand) const { reject(cand); }
 };
--- a/include/crpropa/Module.h
+++ b/include/crpropa/Module.h
@@ -60,8 +60,11 @@ public:
 	void setRejectFlag(std::string key, std::string value);
 	void setAcceptFlag(std::string key, std::string value);
 
-	void printRejectFlag();
-	void printAcceptFlag();
+	// return the reject flag (key & value), delimiter is the "&".
+	std::string getRejectFlag();
+
+	// return the accept flag (key & value), delimiter is the "&"
+	std::string getAcceptFlag();
 };
 
 /**
--- a/include/crpropa/module/Observer.h
+++ b/include/crpropa/module/Observer.h
@@ -281,6 +281,17 @@ public:
 	DetectionState checkDetection(Candidate *candidate) const;
 	std::string getDescription() const;
 };
+
+class ObserverTrajectoryLength: public ObserverFeature {
+private: 
+	double maxLength; 
+public: 
+	ObserverTrajectoryLength(double l);
+
+	DetectionState checkDetection(Candidate *candidate) const;
+};
+
+
 /** @} */
 
 }
--- a/include/crpropa/module/Observer.h
+++ b/include/crpropa/module/Observer.h
@@ -282,16 +282,6 @@ public:
 	std::string getDescription() const;
 };
 
-class ObserverTrajectoryLength: public ObserverFeature {
-private: 
-	double maxLength; 
-public: 
-	ObserverTrajectoryLength(double l);
-
-	DetectionState checkDetection(Candidate *candidate) const;
-};
-
-
 /** @} */
 
 }
--- a/src/Module.cpp
+++ b/src/Module.cpp
@@ -76,12 +76,14 @@ void AbstractCondition::setAcceptFlag(std::string key, std::string value) {
 	acceptFlagValue = value;
 }
 
-void AbstractCondition::printRejectFlag() {
-	std::cout << "rejectKey: " << rejectFlagKey << " with flag: " << rejectFlagValue << "\n";
+std::string AbstractCondition::getRejectFlag() {
+	std::string out = rejectFlagKey + "&" + rejectFlagValue; 
+	return out;
 }
 
-void AbstractCondition::printAcceptFlag() {
-	std::cout << "acceptKey: " << acceptFlagKey << " with flag: " << acceptFlagValue << "\n";
+std::string AbstractCondition::getAcceptFlag() {
+	std::string out = acceptFlagKey + "&" + acceptFlagValue;
+	return out;
 }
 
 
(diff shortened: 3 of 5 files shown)


include/crpropa/Module.h
```diff
@@ -59,7 +59,17 @@ class AbstractCondition: public Module {
 	void setMakeAcceptedInactive(bool makeInactive);
 	void setRejectFlag(std::string key, std::string value);
 	void setAcceptFlag(std::string key, std::string value);
+
+	void printRejectFlag();
```
lukasmerten, line 63: Do we need the print methods or are they just for debugging?

include/crpropa/Module.h
```diff
@@ -59,7 +59,17 @@ class AbstractCondition: public Module {
...
+	void printAcceptFlag();
+};
+
+class DeactivateModule: public AbstractCondition {
```
lukasmerten, line 67: Is it used somewhere or just a general module to deactivate the candidate? 

In this case I would simply call it `Deactivation` as we use nouns for other modules, too, e.g. `MaximumTrajectoryLength`, etc.

Please also add short comment for documentation.

include/crpropa/module/Observer.h
```diff
@@ -281,6 +281,17 @@ class ObserverTimeEvolution: public ObserverFeature {
 	DetectionState checkDetection(Candidate *candidate) const;
 	std::string getDescription() const;
 };
+
```
lukasmerten, line 284: What's the difference compared to the `MaximumTrajectoryLength` module?

src/Module.cpp
```diff
@@ -19,8 +19,7 @@ void Module::setDescription(const std::string &d) {
 
 AbstractCondition::AbstractCondition() :
 		makeRejectedInactive(true), makeAcceptedInactive(false), rejectFlagKey(
-				"Rejected") {
-
+				"Rejected"), rejectFlagValue( typeid(*this).name() ) {
```
lukasmerten, line 22: This would allow to check which module deactivated the candidate, right?

lukasmerten (changes requested):
I have not yet checked if it fixes the actual bug, but in principle it looks good to me. When I have verified the fix and my comments are addressed, I am happy to merge.

include/crpropa/Module.h
JulienDoerner: This is intended for the debugging. Up to now there is no getter function for the key and value of the reject/accept flags. 
One could also think about a real get-function, but this would return a pair of strings, where users can easily be confused by the ordering of the key / value.

include/crpropa/Module.h
JulienDoerner: This is a general module, also intended for debugging. This makes it easier to reproduce the original error as they are faster in deactivating the candidate. 

I will rename it.

include/crpropa/module/Observer.h
JulienDoerner: The first difference is that is Observer based. In this case it is easier to combine different break conditions, having them in the same observer. In principle all functionality could be received with the `MaximumTrajectoryLength`. 

In the practical solution there are two small differences: 
1. The Observer implementation does not have the accept/reject flags, which might reduce the output, if all properties are enabled. 
2. The Observer implementation does not have the overhead with the `observerPositions`. In case without them, the size of the list is checked in every propagation step.

src/Module.cpp
JulienDoerner: Yes, this is the idea. This function was already implemented earlier but the value was not set correctly. I think the module name as default is a good starting option, without going through all modules and set it by hand.

include/crpropa/module/Observer.h
JulienDoerner: I removed this feature as it does not bring a new functionality. Same can be done with `ObserverTimeEvolution` with a single time step.

include/crpropa/Module.h
JulienDoerner: I have changed it to a get function which returns key & value.

JulienDoerner:
The error also occurred for setting a property to a candidate, but only if the property is of type STRING. 

    for i in range(10_000_000): 
      c = Candidate() 
      c.setProperty("k", "str")

will give the same issue as discussed before. With the explicit delete this issue is solved. 

The only remaining point is related to the `AssocVector`. 

Python gives the warning
` swig/python detected a memory leak of type 'Loki::AssocVector< std::string,crpropa::Variant > *', no destructor found.`
when running the following example: 

     c = Candidate()
     c.properties 
     c.isActive() 
     
One solution could be to change the `AssocVector` to `std::map` which should have the same functionality, or does someone no a difference why we should keep the `AssocVector`?

rafaelab:
Hi @JulienDoerner 
In the last exchange you said there is a problem with `AssocVector`.
Is that fixed? Do you think the PR is good to go with this issue unaddressed?

JulienDoerner:
Hey @rafaelab, 

the `AssocVector` gives a warning `swig/python detected a memory leak of type 'Loki::AssocVector< std::string,crpropa::Variant > *', no destructor found.`
At the moment I have not found any real problem, therefore I would suggest leaving it as it is for the moment. 

The underlying problem from #493 is solved with the fixes here.

rafaelab:
Fine with me. But apparently there are a few comments by @lukasmerten that were not addressed (or at least, they were not marked as resolved, such that I cannot  approve the PR).

JulienDoerner:
Hey @rafaelab, 
There was only one question about the usage left. I marked it as resolved, and you should be able to approve it now.

Issue #510, opened 2024-12-10
Wrong time scaling in ObserverTimeEvolution


JanNiklasB:
Dear All,

the non logarithmic time scaling of the `ObserverTimeEvoultion` module in the function `ObserverTimeEvolution::addTimeRange`
in the current master of CRPropa adds the time like `min + index * (max - min) / numb`.
However, since the maximum index is `numb-1`, the function only adds times in the range of 
`(min, min + (numb-1) * (max - min) / numb)`. 

The expected behavior would be a range of time like `(min, max)`,
this can be achieved by calculating the maximum time like `min + index * (max - min) / (numb - 1)`

lukasmerten:
Hi @JanNiklasB 
If have time to provide a fix, I will be happy to to check and merge the PR. Otherwise, thanks for reporting the issue.

JulienDoerner:
I close this as #511 has solved the issue.

PR #511, opened 2024-12-19
Reworking of the ObserverTimeEvolution

JanNiklasB:
Dear All,

in this pull request, I want to introduce a reworked `ObserverTimeEvolution`.
I replaced the `addTimeRange` function which filled the `detList` array with a function
that takes the required index as an argument and returns the time that should be observed.
For that, I also changed other parts of the `ObserverTimeEvolution` class to
correctly utilize that new `getTime` function.
I also kept the option of using a `detList` by utilizing a new constructor.
Furthermore, it is possible to use a user-defined `getTime` function by utilizing
the `setUserDefinedGetTime` function.

Additionally, I fixed the issue mentioned in #510.
For that, it was necessary to change some tests, since those handled
the maximum observed time like `max = numb * dist` when it should
be handled like `max = min + numb * dist`.


doc/pages/example_notebooks/Diffusion/MomentumDiffusion.ipynb
JanNiklasB, line 1: In this file I changed the `Run_MomentumDiffusion` function in cell 12:
`maxD = N_obs * deltaD` -> `maxD = (N_obs + 1) * deltaD`
since the minimum is set to `D_min = deltaD`

test/testBreakCondition.cpp
```diff
@@ -261,7 +262,7 @@ TEST(ObserverFeature, TimeEvolution) {
 
   // detection two
   c.setCurrentStep(0.1);
-  c.setTrajectoryLength(10.05);
+  c.setTrajectoryLength(15.05);
```
JanNiklasB, line 265: Here I changed the trajectory length of the candidate to a length larger then the last
detection threshold so a detection is expected.
10.05 is not enough, since 10 is not a detected length (see comment on line 239)

include/crpropa/module/Observer.h
```diff
@@ -241,12 +241,21 @@ class ObserverParticleIdVeto: public ObserverFeature {
...
 class ObserverTimeEvolution: public ObserverFeature {
 private:
-	std::vector<double> detList;
+	int numb;
+	bool islog = false, useCustomGetTime = false;
+	double min, max;
 public:
+	// List containing all used times, is constructed in constructor or by user manually
+	// (leave empty if you want to rather use functions)
+	std::vector<double> detList;
```
JanNiklasB, line 250: detList is here defined public since it is necessary that the user has full control over it

include/crpropa/module/Observer.h
```diff
@@ -241,12 +241,21 @@ class ObserverParticleIdVeto: public ObserverFeature {
...
+
+	// Pointer to user defined version of getTime (should return time for corresponding index)
+	double (*customGetTime)(int index, double min, double max, int numb);
```
JanNiklasB, line 253: the same goes for the customGetTime pointer

JanNiklasB:
After discussing pull request with @JulienDoerner, I added setter and getter functions as well as some older functions that I previously removed to preserve compatibility.

- The ObserverTimeEvolution now creates the protected vector detList when calling any functions that require detList like addTime or addTimeRange.
- Futhermore, function to edit detList more freely like clear and setTimes were added.
- Since it is only necessary to use custom functions inside of c++, the override of the getTime can now be done directly by overriding the virtual function getTime.
- I added tests to testBreakCondition where the behavior of the arrays is tested in more detail.

For a more detailed description of the single changes, see the comments.

include/crpropa/module/Observer.h
```diff
@@ -240,52 +240,119 @@ class ObserverParticleIdVeto: public ObserverFeature {
...
-	 @param userDefinedFunction	function which will be called instead of getTime, should return time for corresponding index
-	 	min time, max time, and number of times
+	/** Function
+	 Generates the detList if it is empty when for example the 
+	 ObserverTimeEvolution(const std::vector<double> &detList) constructor
+	 was not used.
+	 Use this function to create a detList with can then be modified.
+	 When detList is not empty its entries are used instead of a runtime calculation of the times.	 
+	 */
+	void constructDetListIfEmpty();
```
JanNiklasB, line 299: this function enables the user to manually construct detList

src/module/Observer.cpp
```diff
@@ -246,7 +246,7 @@ std::string ObserverParticleIdVeto::getDescription() const {
 ObserverTimeEvolution::ObserverTimeEvolution() {}
 
 ObserverTimeEvolution::ObserverTimeEvolution(double min, double dist, double numb) {
-	this->max = min + numb * dist;
+	this->max = min + (numb - 1) * dist;
```
JanNiklasB, line 249: fixes the original bug, where the last time to observe was too large because of this

test/testBreakCondition.cpp
```diff
@@ -336,10 +336,12 @@ TEST(ObserverFeature, TimeEvolutionArray) {
   EXPECT_TRUE(times == obs.getTimes());
 
   // test addTimeRange for logarithmic ranges
-  times = {1, 1000};
+  times.clear();
+	for (int i=0; i<4; i++)
+		times.push_back(pow(1000., i / 3.));
```
JanNiklasB, line 341: pow(1000, 2/3) does not exactly return 100 which caused a wrong negative in line 345

test/testBreakCondition.cpp
```diff
@@ -274,37 +277,89 @@ TEST(ObserverFeature, TimeEvolutionLog) {
...
   // detection two
   c.setCurrentStep(0.1);
   c.setTrajectoryLength(10.05);
   obs.process(&c);
-  EXPECT_TRUE(c.isActive());
+  EXPECT_FALSE(c.isActive());
   EXPECT_TRUE(c.hasProperty("Detected"));
 }
 
+TEST(ObserverFeature, TimeEvolutionArray) {
```
JanNiklasB, line 314: This test checks if detList is created and used correctly when using the modifying functions.

src/module/Observer.cpp
```diff
@@ -299,29 +305,82 @@ DetectionState ObserverTimeEvolution::checkDetection(Candidate *c) const {
...
+	constructDetListIfEmpty();
+	detList.push_back(time);
+	numb += 1;  // increase number of entries by one
 }
 
 void ObserverTimeEvolution::addTimeRange(double min, double max, double numb, bool log) {
 	for (size_t i = 0; i < numb; i++) {
-		if (log == true) {
+		if (log) {
+			if ( min == 0 ){
```
JulienDoerner, line 336: Is it better to prove for `min <= 0`?

src/module/Observer.cpp
```diff
@@ -299,29 +305,82 @@ DetectionState ObserverTimeEvolution::checkDetection(Candidate *c) const {
...
+	this->min = detList.front();
+	this->max = detList.back();
+	this->doDetListConstruction = false;
+}
+
+double ObserverTimeEvolution::getTime(size_t index) const {
+	if (!detList.empty()) {
+		return detList.at(index);
+	} else if (islog) {
+		if ( min == 0 ){
```
JulienDoerner, line 361: Is it necessary to prove this in every `getTime` call, which is called in every propagation step? I guess a proof only in the setter functions would be enough.

test/testBreakCondition.cpp
```diff
@@ -273,37 +277,91 @@ TEST(ObserverFeature, TimeEvolutionLog) {
   obs.setFlag("Detected", "Detected");
   // usage of a log scaling for the observer
   bool log = true;
-  obs.add(new ObserverTimeEvolution(5, 5, 2, log));
+  obs.add(new ObserverTimeEvolution(5, 10.05, 2, log));
```
JulienDoerner, line 280: Does this test really reflect the log-scaling? If it only has two points it would be the same in linear or log

test/testBreakCondition.cpp
JulienDoerner: maybe you can check for `EXPECT_NEAR` with some small errorband.

JulienDoerner (changes requested):
Hey @JanNiklasB, 

the overall changes look good for me and it improves the usability for non-uniform observations. 
I left some comments about code style (see above). 

And can you please add your changes to the `CHANGELOG.md`.

test/testBreakCondition.cpp
JanNiklasB: right, it should at least test 3 points

src/module/Observer.cpp
JanNiklasB: right, if it is set right in the setter functions the check here would be unnecessary

JulienDoerner (approved):
This looks good for me now and can be merged.

PR #512, opened 2025-01-31
Implemented ReflectiveShell as optional boundary

ehlertdo:
Hi,
I needed a reflective boundary in my 3D simulations but in the form of a sphere instead of the currently available ReflectiveBox. Maybe this feature could be of more general interest since astrophysical environments are often more spherical than they are boxy.
Note that the shell reflects in both directions, i.e. outside particles are kept out and inside particles are kept in.
Best,
Domenik

PS: I copied the distance() and normal() functions from the Sphere class in Geometry.cpp but probably there's a more elegant way to do this with inheritance.

cr_tracks
cr_tracks


JulienDoerner (changes requested):
Hey @ehlertdo, 

Thanks for sharing your implementation. In general, the tool looks good for me. 

I have tested it and found some deviation when using large steps. I guess this is caused by the position where you calculate the normal. This is given at the current position which is not necessarily a position on the sphere. Would you have an idea how to fix this, or can this be neglected in regular cases?
image

Beside this I would ask you to add your changes to the Changes.md and maybe you can provide a simple test case as a notebook and as unit test.

ehlertdo:
Hi Julien,
I did indeed approximate the true normal to the sphere at the point of intersection by the normal at the final point of the particle. I have now added a few lines to calculate the intersection point before constructing the normal which, in my tests reduces the error at larger step sizes significantly. I also added a unit test for the ReflectiveShell and updated the changelog and doc page.

include/crpropa/module/Boundary.h
```diff
@@ -37,6 +37,32 @@ class PeriodicBox: public Module {
...
+private:
+	Vector3d center;
+	double radius;
+
+public:
+	/** Constructor
+	 @param cen	vector corresponding to the center of the sphere
+	 @param r	value corresponding to the radius of the shell
+	 */
+	ReflectiveShell(Vector3d cen, double r);
```
JulienDoerner, line 57: Code Style: use full words for variable names: `center` instead of `cen`

src/module/Boundary.cpp
```diff
@@ -40,6 +40,58 @@ std::string PeriodicBox::getDescription() const {
 	return s.str();
 }
 
+
+ReflectiveShell::ReflectiveShell(Vector3d cen, double r) :
```
JulienDoerner, line 44: Code Style

src/module/Boundary.cpp
```diff
@@ -40,6 +40,58 @@ std::string PeriodicBox::getDescription() const {
...
+void ReflectiveShell::process(Candidate *c) const {
+	double currentDistance = distance(c->current.getPosition());
+	double previousDistance = distance(c->previous.getPosition());
+	// check if cosmic ray crossed boundary in last step
+	if (currentDistance * previousDistance < 0){
+		Vector3d currentDirection = c->current.getDirection();
+		Vector3d previousPosition = c->previous.getPosition();
+		// get point where trajectory intersects the shell boundary
+		double p_half = previousPosition.dot(currentDirection) / currentDirection.getR2();
+		double q = (previousPosition.getR2() - radius*radius) / currentDirection.getR2();
```
JulienDoerner, line 67: Code Style: spaces around operators. Use `a * b` instead of `a*b`. 
This also happens below. I mark it only once here.

src/module/Boundary.cpp
```diff
@@ -40,6 +40,58 @@ std::string PeriodicBox::getDescription() const {
...
+	center = cen;
+}
+void ReflectiveShell::setRadius(double r) {
+	radius = r;
+}
+
+std::string ReflectiveShell::getDescription() const {
+	std::stringstream ss;
+	ss << "Reflective sphere: " << std::endl
+			<< "   Center: " << center << std::endl
```
JulienDoerner, line 89: Is it useful to define some default unit for the printing like kpc or Mpc?

test/testBreakCondition.cpp
```diff
@@ -345,6 +345,31 @@ TEST(PeriodicBox, low) {
...
+	c.setCurrentStep(20);
+	c.previous.setPosition(Vector3d(80, 20, 30));
+	c.previous.setDirection(Vector3d(10, -1, -1));
+	c.current.setPosition(Vector3d(99.803, 18.0197, 28.0197
+));
+	c.current.setDirection(Vector3d(10, -1, -1));
+
+	shell.process(&c);
+
+	EXPECT_NEAR(89.9965, c.current.getPosition().x, 1e-4);
```
JulienDoerner, line 364: Please leave a comment in the code on how you get to those expected values (or even calculate them with the trigonometric functions here).

JulienDoerner (changes requested):

PR #516, opened 2025-03-17
MacOS install issue

lukasmerten:
I recently had a problem installing the lastest master version of CRPropa on my M1Max, this seemed to be coming from a bad include statement in ModuleList.cpp.in

`#include <bits/stdc++.h>`

This seems to be bad practice according to [this](https://stackoverflow.com/questions/31816095/why-should-i-not-include-bits-stdc-h) stack overflow post. 
I replace the general import with explicits imports of the necessary libraries, which fixed the issue for me.

In addition I tried to fix issue #515 but was not yet successful with it.

So any input on the issue with the not working MacOS runner is appreciated.

### DIFF
--- a/.github/workflows/testing_OSX.yml
+++ b/.github/workflows/testing_OSX.yml
@@ -20,7 +20,7 @@ jobs:
      - name: Preinstall 
        run: | 
          brew install hdf5 fftw cfitsio muparser libomp swig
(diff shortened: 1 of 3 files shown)


rafaelab:
Hi @lukasmerten,
This runs smoothly on my side.
I'll merge it

Issue #517, opened 2025-04-08
Encountered error in ObserverTimeEvolution


JanNiklasB:
Hi,

I tried to construct ObserverTimeEvolution with a linear scaling with a minimum at 0, and encountered the errormessage:
"minimum can not be <= 0 if isLogarithmicScaling=true"
even though I set isLogarithmicScaling=false.

The Error lies in the setMinimum function in line 358, where I forgot to add isLogarithmicScaling as a condition for the error message.
Furthermore, the constructor tries to first call setMinimum and then calls setIsLogarithmicScaling,
which would result in never doing the check, since isLogarithmicScaling has the default false.

I'm preparing the pull request.

With regards,
Jan-Niklas Bohnensack

PR #518, opened 2025-04-08
ObserverTimeEvolution minimum errormessage bug

JanNiklasB:
Hi,

with this pull request I fix the bug in Issue #517.
I added a condition in the if condition to check if isLogarithmicScaling is true or false, which was missing.
Furthermore I set the above variable before calling setMinimum in the constructors.


src/module/Observer.cpp
```diff
@@ -355,7 +355,7 @@ void ObserverTimeEvolution::setTimes(const std::vector<double> &detList){
 }
 
 void ObserverTimeEvolution::setMinimum(double min){
-	if ( min <= 0 ){
+	if ( (min <= 0) && isLogarithmicScaling){
 		std::cout << "minimum can not be <= 0 if isLogarithmicScaling=true" << std::endl;
 		throw new std::invalid_argument("minimum can not be <= 0 if isLogarithmicScaling=true");
```
JulienDoerner, line 360: You are using the `invalid_argument` error type which can be used in C++ but can not be converted to a python error with swig. Please use `runtime_error` instead.

JulienDoerner (changes requested):

src/module/Observer.cpp
JanNiklasB: Done.

Issue #523, opened 2025-06-12
Extern libs (kiss etc) not found when crpropa not top level


JanNiklasB:
Hi,

when installing crpropa over add_subdirectory or FetchContent it is not on top level anymore.
This leads to the problem that all relative paths of external libraries are incorrect.

This should be fixed by adding ${CMAKE_CURRENT_SOURCE_DIR}/ before each relative path.

lukasmerten:
@JanNiklasB was this fixed with your recent PR #537 or is it still an issue?

JanNiklasB:
No, it wasn't, but I have a [branch](https://github.com/JanNiklasB/CRPropa3/tree/RelativePathFix) prepared for a PR, I just didn't found the time to make one yet

JanNiklasB:
It basically would require all relative paths to be replaced with correct absolute paths over the `CMAKE_CURRENT_SOURCE_DIR` and `CMAKE_CURRENT_BINARY_DIR` where the CURRENT is very important.
Furthermore, it is necessary to give the main projects all the additional includes for the subprojects which will require a hidden cache variable.

I will try to create a PR later today

JanNiklasB:
Resolved with #540

PR #524, opened 2025-06-12
Adjusted relative paths of CMakeLists.txt

JanNiklasB:
Hi,

as mentioned in issue #523 when including CRPropa over `add_subdirectory` in other projects, the relative paths are not pointing to the correct folders.

To better support CRPropas behavior when included elsewhere, I changed all relative paths to the correct absolute paths utilizing `CMAKE_CURRENT_SOURCE_DIR` and `CMAKE_CURRENT_BINARY_DIR` respectively.

I also added a new cache variable for `CRPROPA_EXTRA_INCLUDES` so the include paths can be used on the above levels (I do not know how to hide this from the user if possible).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -4,7 +4,7 @@ project(CRPropa Fortran C CXX)
 set(CRPROPA_RELEASE_VERSION 3.2.1+)  # Update for new release
 
 set(CMAKE_CXX_STANDARD 11)
-set(CMAKE_MODULE_PATH "${CMAKE_SOURCE_DIR}/cmake" ${CMAKE_MODULE_PATH})
+set(CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/cmake" ${CMAKE_MODULE_PATH})
 
 set(CRPROPA_EXTRA_SOURCES)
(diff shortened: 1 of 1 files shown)


JanNiklasB:
Here is a short example where the problem might occur when using the  [CRPropa:master](https://github.com/CRPropa/CRPropa3/tree/master) instead of [JanNiklasB:master](https://github.com/JanNiklasB/CRPropa3/tree/master):

https://gist.github.com/JanNiklasB/ca187681f446d323bdc773c425d2e57f

The issue is that when CRPropa is compiled as subproject, the main project can not find the includes of the crpropa subproject (like for example KISS), since those are first added to the usual include folder in the install step which is never done during FetchContent.
The solution is to add all `CRPROPA_EXTRA_INCLUDES` as hidden cache variable (with `mark_as_advanced`)  so the main project can include those as well.

JanNiklasB:
I merged the newest master and took the opportunity to make some other small changes.

JanNiklasB:
I am closing my PR for now.
The reason is that in my inexperience with Github I made a PR from the master of my fork which I would typically like to avoid.
Furthermore, I added many other unrelated changes in the believe that one PR for many things is better then many, but I do not believe that anymore and would rather make a PR for each change separately to better discuss them.

PR #525, opened 2025-06-27
Explicit time in CRPropa

JulienDoerner:
Dear all, 

with this PR we introduce an explicit time within CRPropa. 
It is written as a new candidate property, which will be increased within the propagation step. Here we are still using $v=c$, but this might be adjusted in the future. 

With the explicit time from the Candidate we also add the corresponding outputs and units to the code. 

As a first example we implement the time-dependence of the advection field. Here the interface for the Advection base class is changed, and it takes now also the time (besides the space dependence that was included before). @sophieaerdker added one example for the documentation to illustrate this feature. 

On behalf of 
@LeanderSchlegel, @sophieaerdker, @JulienDoerner


src/advectionField/TimeDependentAdvectionFields.cpp
```diff
@@ -0,0 +1,188 @@
...
+	double r = position.getR();
+	double A = rho0; 
+
+	double R = pow(E0 / A, 1./5.) * pow(time, 2. / 5.); 
+    double vs = 2. / 5. * pow(E0 / A, 1. / 5.) * pow(time, -3. / 5.); 
+
+	double a = (-3 * pow(r,3) * (1 + pow(r/R, 8)) * 1./pow(cosh((r - R)/l_sh),2)) / (l_sh * (5 + (3 * pow(r,8))/pow(R,8)))
+     + (9 * pow(r,2) * (1 + pow(r/R, 8)) * (1 - tanh((r - R)/l_sh))) / (5 + (3 * pow(r,8))/pow(R,8))
+     - (72 *pow(r,10) * (1 + pow(r/R,8)) * (1 - tanh((r - R)/l_sh)))/(pow(5 + (3 * pow(r,8))/pow(R,8),2) * pow(R,8)) 
+     + (24 * pow(r,10) * (1 - tanh((r - R)/l_sh)))/((5 + (3 * pow(r,8))/pow(R,8)) * pow(R,8));
```
rafaelab, line 140: Use CRPropa's `pow_integer<n>(x)` for x^n.
This is faster than `pow`.

src/advectionField/TimeDependentAdvectionFields.cpp
```diff
@@ -0,0 +1,188 @@
...
+    double dudr = 0.5 * vs * pow(r, -2) * 1./ R * a;
+
+	return dudr;
+
+}
+
+
+void SedovTaylorBlastWave::setShockWidth(double w) {
+	l_sh = w;
+	return;
```
rafaelab, line 151: In principle, there is no need for this `return`, especially considering that the function is a setter.

rafaelab: In principle, there is no need for this `return`, especially considering that the function is a setter.

JulienDoerner:
hey @rafaelab, 

I have applied your code style comments. 

- > Now we have two quantities that handle time, redshift and time itself. Many of our functions are of the form getWhatever(position, redshift). Having only the advection field of the form getWhatever(position, time) makes things inconsistent. Especially because users might (rightly) think that the magnetic field behaves in the same way. I have no clear idea at the moment on how to handle this.
- Yes this is a critical point. I think the difference is that the advection fields do not have any kind of redshift dependence. In the future also the interface for the magnetic field, gas distribution, and photon field have to be adapted to add the time variable. 

- > Similarly, I see a lot of redundancy with the trajectory length colum. But of course, when v≠c, they are both needed and not redundant.
- Of couse this will be of special relevance for the v<c case. But also now, if we add a time dependent injection there is a difference between the local CR time (in the trajectory length) starting at the injection and the global system time (in the time column) which has to be defined by the user. 

- > It should be made more clear in the documentation how T=0 is defined.
- This choice is something the user can decide on. Therefore, there is a difference to the time defined by the redshift as z=0 is properly defined. We have to add the definition at every module on its own. 

- > I suppose the performance impact of having one more candidate property is small, right?
- I have looked for runtime differences but could not find any. I have not tested the used memory but typically the candidates are small compared to the loaded grids used. And a percentage effect here should not be important. 


A more general time class might be a solution, but I would assume this would be much more complicated to use. 
In this the user would have to define the relation between the local system time $T$ and the definition of $z=0$.

rafaelab:
Thanks for the review, @JulienDoerner 
From my side, it looks ok. 
However, some tests are not running properly. (If it's just the OSX test, it's ok, but in this case ubuntu20 seems to be stuck).

lukasmerten:
I just wanted to add that a time feature would be really useful for the ensemble averaged approach. In our current version SDE Solver even I get confused when I have to convert time to length step etc.  

However, I agree that having potentially two different measures of time (time and redshift) is not ideal. For the future, we could think about a general switch for a CRPropa simulation. Something like:
"time measure" = time or "time measure" = redshift, which would determine how the additional argument in magnetic fields or advection field, etc. has to be interpreted. 

Probably something for discussion in a future CRPropa meeting.

lukasmerten:
From my point of view the failing tests should not block the merging, as we know what is causing them.

PR #527, opened 2025-07-23
Numpy2 implementation and updates on testing

### LINKED ISSUE #515: MacOS runner not working
The MacOS test workflow is not working because of a numpy version mismatch.

`ERROR: Could not find a version that satisfies the requirement numpy==1.26 (from versions: 1.3.0, 1.4.1, 1.5.0, 1.5.1, 1.6.0, 1.6.1, 1.6.2, 1.7.0, 1.7.1, 1.7.2, 1.8.0, 1.8.1, 1.8.2, 1.9.0, 1.9.1, 1.9.2, 1.9.3, 1.10.0.post2, 1.10.1, 1.10.2, 1.10.4, 1.11.0, 1.11.1, 1.11.2, 1.11.3, 1.12.0, 1.12.1, 1.13.0, 1.13.1, 1.13.3, 1.14.0, 1.14.1, 1.14.2, 1.14.3, 1.14.4, 1.14.5, 1.14.6, 1.15.0, 1.15.1, 1.15.2, 1.15.3, 1.15.4, 1.16.0, 1.16.1, 1.16.2, 1.16.3, 1.16.4, 1.16.5, 1.16.6, 1.17.0, 1.17.1, 1.17.2, 1.17.3, 1.17.4, 1.17.5, 1.18.0, 1.18.1, 1.18.2, 1.18.3, 1.18.4, 1.18.5, 1.19.0, 1.19.1, 1.19.2, 1.19.3, 1.19.4, 1.19.5, 1.20.0, 1.20.1, 1.20.2, 1.20.3, 1.21.0, 1.21.1, 1.22.0, 1.22.1, 1.22.2, 1.22.3, 1.22.4, 1.23.0, 1.23.1, 1.23.2, 1.23.3, 1.23.4, 1.23.5, 1.24.0, 1.24.1, 1.24.2, 1.24.3, 1.24.4, 1.25.0, 1.25.1, 1.25.2, 1.26.2, 1.26.3, 1.26.4, 2.0.0, 2.0.1, 2.0.2, 2.1.0rc1, 2.1.0, 2.1.1, 2.1.2, 2.1.3, 2.2.0rc1, 2.2.0, 2.2.1, 2.2.2, 2.2.3)
ERROR: No matching distribution found for numpy==1.26
Error: Process completed with exit code 1.`

### LINKED ISSUE #526: Ubuntu-20 workflows needs to be replaced
Our workflows that are still based on Ubuntu-20 need to be updated to more recent versions of the OS. See the following for currently available systems (https://github.com/actions/runner-images).

JulienDoerner:
In this PR the `numpy` interface is adapted to account for the changes introduced with the change to version 2. 

The Vector3 class can now be used as a numpy array. 
The second issue with numpy >= 2.0 is the compatibility with the Lens interface. This can not easily be adapted. Therefore, I set the default installation behavior to off and add a warning when compiling CRPropa using the lenses with numpy 2. 

In this PR I also address the issues in the testing. 
The Ubuntu20 test (#526) is removed and replaced by a test for Ubuntu24. This test is now run on an x86 and an arm architecture. 
The MacOS test (#515) has been adapted to the recommended installation by setting all needed flags/pathes directly. This might become a problem in the future when the MacOS runner is updated with newer versions, but this is only a small issue. 
In the MacOS test there was also an issue with the EMPairProduction, where sometimes the sampling of the center of mass energy was sampled within the first bin and therefore it used non-valid elements in the tabulated values.


include/crpropa/Vector3.h
```diff
@@ -6,6 +6,11 @@
 #include <vector>
 #include <limits>
 #include <algorithm>
+#include <Python.h>
```
JanNiklasB, line 10: Including python header with the cmake option `ENABLE_PYTHON=off` would result in an include error, since it is neither ensured that python would exist for the user nor do we include the python includes anywhere

include/crpropa/Vector3.h
```diff
@@ -410,6 +415,84 @@ class Vector3 {
 		data[2] = f;
 		return *this;
 	}
+
```
JanNiklasB, line 426: The above includes and the following new lines (except `getDescription`) need to be guarded by a definition when python is not found

JanNiklasB (changes requested):
Hey,

you include python headers inside `Vector3.h` without any protection when python headers could not be found (see comments), those would fail to compile if the headers do not exist or are not included when `ENABLE_PYTHON=OFF`.

Other then that it looks good.

include/crpropa/Vector3.h
JulienDoerner: I added the CRPROPA_HAVE_PYTHON flag to check before.

lukasmerten (approved):

JanNiklasB (approved):

PR #535, opened 2026-01-26
position dependent photon field for EM interactions (optimised)

GDMarco:
### General description

This is a fully working version of the CRPropa code with some implementation to allow spatially dependent photon field to be initialised and employed for electromagnetic interactions. A particular example is the introduction of the galactic interstellar field (the density and interaction rate tables are at this repo: [ISRF tables](https://saco.csic.es/s/2FD4AespQ4TQd57).) 
In principle, this implementation can be used for any spatial dependent photon field properly build by the user. 
To allow the introduction of such spatially variable photon fields in CRPropa, the classe and modules that have been modified are`PhotonBackground.*`, `EMPairProduction.*`, `EMDoublePairProduction.*`, `EMInverseComptonScattering.*` and `EMTripletPairProduction.*`. Moreover, the class `InteractionRates.*` has been implemented to handle the process rates. 
An example on how to use the code is contained in the notebooks. 

### Technical details

The `.cc` files of `PhotonBackground.*`,`EMPairProduction.*`, `EMDoublePairProduction.*`, `EMInverseComptonScattering.*` and `EMTripletPairProduction.*` employ some tools of the `#include <filesystem>` library, only available since `C++17`. The proper library is specified in the `CMakeLists.txt` files, as well as in the header of the `.cc` files.  

The code uses the [nanoflann](https://github.com/jlblancoc/nanoflann) functionalities to look for the closest photon field point in your custom grid and evaluate the density and/or the rate correctly. In this new version, nanoflann is implemented as a header-only library among the ones in the `/libs/` directory.


src/module/EMDoublePairProduction.cpp
```diff
@@ -1,25 +1,57 @@
 #include "crpropa/module/EMDoublePairProduction.h"
 #include "crpropa/Units.h"
 #include "crpropa/Random.h"
+#include "crpropa/PhotonBackground.h"
+#include "crpropa/InteractionRates.h"
+#include "crpropa/Geometry.h"
```
JanNiklasB: `PhotonBackground.h`, `InteractionRates.h` and `Geometry.h` are already included in the corresponding `EMDoublePairProduction.h` header. Please remove them here for style reasons.

include/crpropa/module/EMDoublePairProduction.h
```diff
@@ -23,53 +25,66 @@ namespace crpropa {
...
-	void initRate(std::string filename);
-	void process(Candidate *candidate) const;
-	void performInteraction(Candidate *candidate) const;
+  /** Constructor
+   @param photonField		target photon field
+   @param haveElectrons	if true, add secondary electrons as candidates
+   @param thinning		weighted sampling of secondaries (0: all particles are tracked; 1: maximum thinning)
+   @param limit			step size limit as fraction of mean free path
+   @param surface     suface to enclose the grid nodes to be loaded
+   @param interactionRates(hidden)  object to store and access to the interaction rates of the process
```
JanNiklasB: The interactionRates should not appear as parameter since the user is not able to actually specify them that way. Instead mention them in the description of the constructor like for example:
```
/** Constructor
* You can set interactionRates via initRate(std:.string filename)
* ....
* /
```

include/crpropa/module/EMDoublePairProduction.h
```diff
@@ -23,53 +25,66 @@ namespace crpropa {
...
+  /** Apply thinning with a given thinning factor
+   * @param thinning factor of thinning (0: no thinning, 1: maximum thinning)
+   */
+  void setThinning(double thinning);
+  
+  /** Apply a surface that confine the position dependent photon field region.
+   * @param surface closed surface to confine the grid to be  uploaded
+   */
+  void setSurface(ref_ptr<Surface> surface);
+  bool hasSurface() const;
```
JanNiklasB: It might be better to return the Surface pointer instead, that way the user can do the check and still use the pointer afterwards.

include/crpropa/module/EMDoublePairProduction.h
```diff
@@ -23,53 +25,66 @@ namespace crpropa {
...
+  
+  /** set a custom interaction tag to trace back this interaction
+   * @param tag string that will be added to the candidate and output
+   */
+  void setInteractionTag(std::string tag);
+  std::string getInteractionTag() const;
+  
+  void initRate(std::string filename, InteractionRatesHomogeneous* intRatesHom);
+  void initRatePositionDependentPhotonField(std::string filepath, InteractionRatesPositionDependent* intRatesPosDep);
```
JanNiklasB: A description of `initRate` and `initRatePositionDependentPhotonField` is necessary, it is not clear for the average user how to use those functions.

Furthermore, we might need to add a `initRate(std::string)` overload so the old call behavior is preserverd.

src/module/EMDoublePairProduction.cpp
```diff
@@ -1,25 +1,57 @@
...
-	initRate(getDataPath("EMDoublePairProduction/rate_" + fname + ".txt"));
+  
+  this->photonField = photonField;
+  std::string fname = photonField->getFieldName();
+  setDescription("EMDoublePairProduction: " + fname);
+  
+  if (!this->photonField->hasPositionDependence()) {
+    
+    this->interactionRates = new InteractionRatesHomogeneous("interactionRatesHomogeneous", false);
+    InteractionRatesHomogeneous* intRatesHom = static_cast<InteractionRatesHomogeneous*>(this->interactionRates.get());
```
JanNiklasB: With the requested change of `initRate` in the header file this pointer and the one in line 51 are not required. Furthermore initRate would be used in both cases so the if condition would is then only required inside the initRate function.

include/crpropa/module/EMDoublePairProduction.h
JanNiklasB: `initRate` and `initRatePositionDependentPhotonField` should accept a ref_ptr of a generalized InteractionRates module which contains a property that tells us if it is position dependent or not like the photon fields, otherwise it would be difficult for the user to create their own interaction rates.
With that is would probably be better to do a check in just one `initRate` function instead of creating another one.

src/module/EMDoublePairProduction.cpp
```diff
@@ -34,37 +66,142 @@ void EMDoublePairProduction::setThinning(double thinning) {
...
-void EMDoublePairProduction::initRate(std::string filename) {
+void EMDoublePairProduction::setSurface(ref_ptr<Surface> surface) {
+    this->surface = surface;
+}
+
+bool EMDoublePairProduction::hasSurface() const {
+    return this->surface != nullptr;
+}
+
+void EMDoublePairProduction::initRate(std::string filename, InteractionRatesHomogeneous* intRatesHom) {
```
JanNiklasB: As mentioned in the header, the old code call behaviour must be preserved, so the required InteractionRates must at least be optional

src/module/EMDoublePairProduction.cpp
```diff
@@ -34,37 +66,142 @@ void EMDoublePairProduction::setThinning(double thinning) {
...
 	std::ifstream infile(filename.c_str());
 
+  std::vector<double> tabEnergy;
+  std::vector<double> tabRate;
```
JanNiklasB: those vectors might need to be global in case no interaction rates are used

include/crpropa/module/EMDoublePairProduction.h
```diff
@@ -23,53 +25,66 @@ namespace crpropa {
...
 private:
-	ref_ptr<PhotonField> photonField;
-	bool haveElectrons;
-	double limit;
-	double thinning;
-	std::string interactionTag = "EMDP";
-
-	// tabulated interaction rate 1/lambda(E)
-	std::vector<double> tabEnergy;  //!< electron energy in [J]
-	std::vector<double> tabRate;  //!< interaction rate in [1/m]
```
JanNiklasB, line 37: Those vectors will still be needed in the process function if we keep the behavior without interactionRates.

src/module/EMDoublePairProduction.cpp
```diff
@@ -34,37 +66,142 @@ void EMDoublePairProduction::setThinning(double thinning) {
...
+    
 	if (!infile.good())
 		throw std::runtime_error("EMDoublePairProduction: could not open file " + filename);
 
-	// clear previously loaded interaction rates
-	tabEnergy.clear();
-	tabRate.clear();
-
```
JanNiklasB, line 46: Since we still need `tabEnergy`and `tabRate` as class members we also still need to clear here

src/module/EMDoublePairProduction.cpp
```diff
@@ -34,37 +66,142 @@ void EMDoublePairProduction::setThinning(double thinning) {
...
+      
+    }
+    infile.ignore(std::numeric_limits < std::streamsize > ::max(), '\n');
+    
+  }
+	
+  infile.close();
+    
+  intRatesHom->setTabulatedEnergy(tabEnergy);
+  intRatesHom->setTabulatedRate(tabRate);
```
JanNiklasB: Should only be done in case `intRatesHom` are not `NULL`

src/module/EMDoublePairProduction.cpp
```diff
@@ -34,37 +66,142 @@ void EMDoublePairProduction::setThinning(double thinning) {
...
+    
+}
+
+  if (tabRate.empty())
+    throw std::runtime_error("Rate's table empty! Check if the surface is properly set.");
+    
+  intRatesPosDep->setTabulatedRate(tabRate);
+  intRatesPosDep->setPhotonDict(photonDict);
+    
+}
```
JanNiklasB, line 87: If the function is only done in case `intRatesPosDep` is position dependent there should be no conflict with the internal `tabRate` and `tabEnergy` as long as we rename them here (they are not needed in case of interaction rates anyway

src/module/EMDoublePairProduction.cpp
```diff
@@ -34,37 +66,142 @@ void EMDoublePairProduction::setThinning(double thinning) {
...
 
-void EMDoublePairProduction::performInteraction(Candidate *candidate) const {
-	// the photon is lost after the interaction
-	candidate->setActive(false);
+void EMDoublePairProduction::initRatePositionDependentPhotonField(std::string filepath, InteractionRatesPositionDependent* intRatesPosDep) {
 
-	if (not haveElectrons)
-		return;
+  std::vector<std::vector<double>> tabRate;
```
JanNiklasB: Use different name for `tabRate` vector since we will need to keep the member version (tabRate as local variable works but it is confusing)

src/module/EMDoublePairProduction.cpp
JanNiklasB, line 87: As mentioned above it would be best to only have one `initRate` function (or one with multiple overloads) which checks for the position dependence of the interactions rates if any

src/module/EMDoublePairProduction.cpp
```diff
@@ -96,14 +240,16 @@ void EMDoublePairProduction::process(Candidate *candidate) const {
...
 	// interaction rate
-	double rate = interpolate(E, tabEnergy, tabRate);
-	rate *= pow_integer<2>(1 + z) * photonField->getRedshiftScaling(z);
+  double rate = this->interactionRates->getProcessRate(E, position);
+	
+    
+  if (rate < 0)
+    return;
+    
+  rate *= pow_integer<2>(1 + z) * photonField->getRedshiftScaling(z);
```
JanNiklasB: Since the interaction rates are not a requirement the deleted lines are still needed and must be used in case no interaction rate is defined.

include/crpropa/module/EMDoublePairProduction.h
```diff
@@ -23,53 +25,66 @@ namespace crpropa {
...
+  std::string getInteractionTag() const;
+  
+  void initRate(std::string filename, InteractionRatesHomogeneous* intRatesHom);
+  void initRatePositionDependentPhotonField(std::string filepath, InteractionRatesPositionDependent* intRatesPosDep);
+  
+  void process(Candidate *candidate) const;
+  void performInteraction(Candidate *candidate) const;
+  
+protected:
+  std::string splitFilename(const std::string str);
```
JanNiklasB: This should either be a Common.h function or defined locally inside the .cpp file that needs it

include/crpropa/module/EMInverseComptonScattering.h
JanNiklasB: The comments on EMDoublePairProduction.h are here true aswell, we cannot just drop the old call behavior like this and we should keep the accepted modules as general as possible (see initRates)

src/InteractionRates.cpp
```diff
@@ -0,0 +1,215 @@
+#include "crpropa/InteractionRates.h"
+#include "crpropa/Referenced.h"
+#include "crpropa/Vector3.h"
+#include "crpropa/Units.h"
+#include "crpropa/Random.h"
```
JanNiklasB, line 3: Try not to double includes between this file and InteractionRates.h

include/crpropa/InteractionRates.h
JanNiklasB, line 1: Every public function needs to have a short and detailed description (see other classes). It is important to be in that style so we can automatically update the documentation over doxygen

include/crpropa/PhotonBackground.h
```diff
@@ -18,45 +21,60 @@ namespace crpropa {
...
+  
+  bool hasRedshiftDependence() const {
+    return this->isRedshiftDependent;
+  }
+  
+  bool hasPositionDependence() const {
+    return this->isPositionDependent;
+  }
+  
+  bool hasSurface() const {
```
JanNiklasB: similar to the modules I would return the surface pointer instead of checking if it exists, the user has then more possibilities

include/crpropa/PhotonBackground.h
```diff
@@ -66,16 +84,16 @@ class PhotonField: public Referenced {
  This class reads photon field data from files;
  The first file must be a list of photon energies [J], named fieldName_photonEnergy.txt
  The second file must be a list of comoving photon field densities [1/m^3], named fieldName_photonDensity.txt
- Optionally, a third file contains redshifts, named fieldName_redshift.txt
+ Optionally, a third file contains redshifts, named fieldName_redshift.txt.
  */
 class TabularPhotonField: public PhotonField {
 public:
-	TabularPhotonField(const std::string fieldName, const bool isRedshiftDependent = true);
+	TabularPhotonField(const std::string fieldName, const bool isRedshiftDependent = true, const bool isPositionDependent = true);
```
JanNiklasB: I do not see why we would need `isPositionDependent` as argument here, this class is never position dependent and there is no requirement for this argument from the base class.

src/PhotonBackground.cpp
```diff
@@ -1,52 +1,84 @@
...
-	if (this->isRedshiftDependent)
-		readRedshift(getDataPath("") + "Scaling/" + this->fieldName + "_redshift.txt");
-
-	checkInputData();
-
-	if (this->isRedshiftDependent)
-		initRedshiftScaling();
+  this->isPositionDependent = isPositionDependent;
+
+  if (this->isPositionDependent) {
```
JanNiklasB: instead of throwing an error when the user sets `isPositionDependent=true` remove the argument from the constructor and just set it manually to false

src/PhotonBackground.cpp
```diff
@@ -60,127 +92,382 @@ double TabularPhotonField::getRedshiftScaling(double z) const {
...
+    int nLoadedFiles = 0;
+    
+    if (this->isRedshiftDependent) {
+        
+        throw std::runtime_error("Photon Field " + fieldName + " is redshift dependent! It is not the correct class. \n");
+        
+    } else if (!this->isPositionDependent) { 
+        
+        throw std::runtime_error("Photon Field " + fieldName + " is not position dependent! It is not the correct class. \n");
+
```
JanNiklasB: Like above, if you do not want to have a RedshiftDependence then don't give the user the option to set it

include/crpropa/PhotonBackground.h
```diff
@@ -90,6 +108,40 @@ class TabularPhotonField: public PhotonField {
...
+ This class reads photon field data from files in the appropriate directory;
+ The first files must be lists of photon energies [J], named fieldName_photonEnergy.txt and contained in the subdirectory /photonEnegy/;
+ The second files must be lists of comoving photon field densities [1/m^3], named fieldName_photonDensity.txt and contained in the subdirectory /photonDensity/;
+ The generated files through the CRPropa procedure (https://crpropa.github.io/CRPropa3/pages/example_notebooks/custom_photonfield/custom-photon-field.html) have a different ordering: the energy bins from the larger to the lower.
+ No redshift dependence is available.
+ The surface is defined to include the nodes of the grid contained within.
+ */
+class TabularSpatialPhotonField: public PhotonField {
+public:
+    TabularSpatialPhotonField(const std::string fieldName, const bool isRedshiftDependent = true, const bool isPositionDependent = true, ref_ptr<Surface> surface = nullptr);
```
JanNiklasB: If the field is not dependent on Redshift or Position, then it would be better to manually set them to false instead of giving the user the option to set them here. Also, since the options are set to true by default and the constructor throws an error when that happens, the user probably is forced to set them to false in every case.

CMakeLists.txt
```diff
@@ -3,7 +3,16 @@ cmake_minimum_required(VERSION 3.14)
 project(CRPropa Fortran C CXX)
 set(CRPROPA_RELEASE_VERSION 3.2.1+)  # Update for new release
 
-set(CMAKE_CXX_STANDARD 11)
+# Detect platform and set C++ standard
+if(APPLE)
+    message(STATUS "Building on macOS, using C++11")
+    set(CMAKE_CXX_STANDARD 11)
+#    set(CMAKE_CXX_STANDARD_REQUIRED ON)
```
JanNiklasB: Is there something on macOS that prevents us to set this CXX standard to required? If not it would be good to uncomment this I think.

CMakeLists.txt
```diff
@@ -538,6 +550,10 @@ if(ENABLE_PYTHON AND Python_FOUND)
 
 endif(ENABLE_PYTHON AND Python_FOUND)
 
+target_include_directories(crpropa
```
JanNiklasB: Please add nanoflann to crpropa over `CRPROPA_EXTRA_INCLUDE` and `CRPROPA_EXTRA_SOURCES` (see other libraries).

JanNiklasB (changes requested):
Hi,

thanks for your contribution!
I have done a first review of your pull request and noticed the following things.
- Please try to not include lines in your pull request that are only altered in white spaces without having any purpose, you often only change tabs to spaces or add new empty lines, this makes it difficult to spot actual changes and leads to difficult to read code.
- Reduce the amount of includes you use in your files, any not needed include or already in the header existing includes should not be in the .cpp file again. This is only a style thing since otherwise the code seems a little cluttered.
- For the 4 changed module files (EM*.h and EM*.cpp) I only commented in EMDoublePairProduction.* but my comments hold true for all module files:
  -> The old behavior of function and classes regarding function calls can not be changed, it needs to be backwards compatible ( We need to at least notify the community first when we remove or change crucial features) 
  -> Everything needs to be as modular as possible, so if you use a class like the interaction rates in a function you need to only require the base class so the user can create their own versions and hand them over. Those base classes then need everything to distinguish them inside those functions
- You require the function `splitFilename` in many of your classes but this function does not depend on any class member, so it would be better to add this function to Common.h as a standalone
- Every function which is not obvious to use or where it is not obvious what it does needs to have a description
- You need to add your changes to the change log `CHANGELOG.md`

Many comments would repeat themself so I usually mention something only 1 or 2 times.

With regards,
Jan-Niklas

include/crpropa/module/EMDoublePairProduction.h
GDMarco: Thanks for the comment! I put a brief description of the `init*Rate` functions, although the user do not need to handle those directly. 

Could you please elaborate further the second part of this comment?

JanNiklasB: Originally `initRate` was callable with the signiture `initRate(std::string)`, this still needs to be possible, otherwise existing simulations will fail.

GDMarco: thanks for the comment! 

Please, note the different behaviour of the two `init*Rate` functions for the two different cases, i.e. homogenous _vs_ spatially dependent, since also the organisation of the data, how they are read etc. 

Moreover, the `initRate` function preserves the _old_ structure of the equivalent function, with the only difference due to the fact that the tables are loaded to a specific `InteractionRates` object.

Note also that the users do not need to handle anything regarding the `InteractionRates` class. This latter is internally built and managed in the module initialisation and during the simulation run, just according to the flag in `hasSpatialDependent()` of each background class.

src/module/EMDoublePairProduction.cpp
GDMarco: thanks! As previously commented, this definition does not affect the common usage of CRPropa. It is handled internally.

Moreover, in this new version the `InteractionRates` class is used both in the homogenous and inhomogeneous radiation cases. Thus, it is unavoidable to go through this class. As I reported in some of the developer meetings, it does not affect the computational time for the homogenous case, compared to the current CRPropa version.

src/module/EMDoublePairProduction.cpp
GDMarco: Thanks, see the the comment above please.

include/crpropa/module/EMDoublePairProduction.h
GDMarco: Please, follow from the comment above.

src/module/EMDoublePairProduction.cpp
GDMarco: Lastly, a last brief comment concerning the design of the code. I think the `InteractionRates` class introduces a _symmetric_ treatment of the electromagnetic interaction rates.

src/module/EMDoublePairProduction.cpp
GDMarco: Same as above.

src/module/EMDoublePairProduction.cpp
GDMarco: Same as above. `intRatesHom` has to be set to handle the interactions with homogenous photon backgrounds.

src/module/EMDoublePairProduction.cpp
GDMarco: As I write above, there would not be any conflict within this design. `std::vector<std::vector<double>> tabRate;` is only used in this function to then set the proper table inside the dedicated `InteractionRates` object.

include/crpropa/module/EMDoublePairProduction.h
JanNiklasB: > Moreover, the initRate function preserves the old structure of the equivalent function, with the only difference due to the fact that the tables are loaded to a specific InteractionRates object.

Without a default value for `InteractionRatesHomogenous*` it does not preserve the old behavior. I get that it is not possible to call initRate the new way without this parameter and that it is not really possible to give it a default value, but because of that we should give the option to use the old way, at least for now.

> Note also that the users do not need to handle anything regarding the InteractionRates class. This latter is internally built and managed in the module initialisation and during the simulation run, just according to the flag in hasSpatialDependent() of each background class.

`initRate` is a public function, if the user should not call them directly we should maybe consider making it `protected`. However, we do not know if any user has called initRate() directly in their projects, for example when writing their own initialization.

src/module/EMDoublePairProduction.cpp
JanNiklasB: Sorry, what does "symmetric treatment" mean?

src/module/EMDoublePairProduction.cpp
JanNiklasB: No, if initRatesHom is set as nullpointer then there would be an error, so you always need to check if the pointer is valid first

src/module/EMDoublePairProduction.cpp
GDMarco: Please, also note that the advantage of having a complete and general `InteractionRates` class for both homogeneous and inhomogeneous fields regards the computation of the `double rate` in the `process` function. Such calculation is not performed strictly inside the `process` function itself but in the `InteractionRates` class.
The same advantage is spotted in the `performInteraction` function to get the proper vectors used to compute the secondaries energy, see e.g. `EMInverseComptonScattering`.  

This is why the designed was changed to introduce a general `InteractionRates` class. While the simulation is running, the `process` and  `performInteraction` functions are called several times. Any 'if' sentence or any possible complications inside them would critically slower the simulation.

src/module/EMDoublePairProduction.cpp
GDMarco: Following the previous comment: I think it is better to have a slightly more complex initialisation (and it is obviously slower in the spatial dependent case) than overloading the simulation while it is running, hence the `process` and `performInteraction` functions. In addition the rates and cumulative distribution vectors are just computed in another object, namely the specific `InteractionRates` one. 

The re-introduction of a case without the `InteractionRates` class, as in the previous design would mean a different treatment of the interaction in the `process` and `performInteraction` functions, according to the property of the photon field.

src/module/EMDoublePairProduction.cpp
JanNiklasB: I do not doubt the advantage of your code over the old version, I mainly talk about coding style when I talk about design. In CRPropa we have a style guide which needs to be followed and we need to preserve any old behavior (or warn users long enough before a change). My other suggestions are regarding the modularity.

Just to clarify potential misunderstandings.

src/module/EMDoublePairProduction.cpp
GDMarco: Again the interaction rate is always internally defined. 

In the yet implemented case of the interstellar radiation field (ISRF) the redshift dependence is carefully settled to `false`, thus there are no physical inconsistencies due to the redshift.

src/module/EMDoublePairProduction.cpp
JanNiklasB: I do not see why a overload would affect the performance relevantly.
What do you mean "overloading the simulation while it is running"? Overloading a function only means to provide a different way of calling it while having the same name -> see for context [CPP Overloading](https://en.cppreference.com/w/cpp/language/operators.html)

include/crpropa/InteractionRates.h
GDMarco: By adding the description to `init*Rate`, although they are not _public_ (in the sense they are handled internally, see above), I think they are everywhere commented as - or at least similar to the current version.

Please, refer to the missing points otherwise. Thanks!

JanNiklasB: sorry, I of course meant public as in a public member function. All public member functions are exposed to other child classes and in that sense the user, so they require a description. If your function should not be accessible by other classes then you should not set them to public (you can set them for example to protected or private).

The argument "handled internally" does not really hold when considering that this is open source code and developed by people who might not know what your code should do, so by providing a description you help the other CRPropa Developer to easier utilize your code in the future. So even if you define a function as private member, you should still provide at least a short description.

CMakeLists.txt
GDMarco: I add it to the `CRPROPA_EXTRA_INCLUDE`, as the other libraries (e.g. `kiss`). Why also to `CRPROPA_EXTRA_SOURCES`?

JanNiklasB: You are right, for nanoflann you only have Includes, so only to `CRPROPA_EXTRA_INCLUDE`

src/module/EMDoublePairProduction.cpp
GDMarco: Same as before.

GDMarco:
Thanks for the comments! I fixed and commented all the issues you raised. I will need to implement it merging the currently accepted PRs.

doc/pages/example_notebooks/photon_propagation/galactic_propagation_gammarays_ISRF.ipynb
```diff
@@ -0,0 +1,268 @@
...
+    "z = 0.\n",
+    "\n",
+    "# source distance from the Earth, placed at (-8.5, 0, 0) kpc in the JF12 galactic magnetic field model. \n",
+    "D = np.sqrt((x - (-8.5))**2. + y * y + z * z)\n",
+    "\n",
+    "# aperture angle of the source (rad).\n",
+    "Acon = 0.1\n",
+    "\n",
+    "# radius of the spherical observer placed at the Earth position (kpc).\n",
+    "Orad = np.tan(Scon) * D \n",
```
JanNiklasB, line 40: I think `Scon` should be `Acon` since `Scon` is nowhere defined

doc/pages/example_notebooks/photon_propagation/galactic_propagation_gammarays_ISRF.ipynb
```diff
@@ -0,0 +1,268 @@
...
+    "outputName = 'galactic_propagation_3Disrf.txt'\n",
+    "outputName_inj = 'galactic_propagation_3Disrf_inj.txt'\n",
+    "\n",
+    "# photon backgrounds:\n",
+    "# homogeneuos, \n",
+    "cmb = CMB()\n",
+    "ebl = IRB_Gilmore12()\n",
+    "\n",
+    "# position dependent. (Sphere to load only the ISRF nodes close to observer and the source.)\n",
+    "isrf = ISRF_Freudenreich98(Sphere(Vector3d(x, y, z) * kpc, D * kpc + 1 * kpc))"
```
JanNiklasB, line 78: This is currently failing, but only because of the currently missing photon files so it is fine (fyi @Grindegreen )

CMakeLists.txt
JanNiklasB, line 573: please add `install(DIRECTORY libs/nanoflann/include/ DESTINATION include)` so the nanoflann header files are included correctly (I missed that last time, sorry)

JanNiklasB (changes requested):
Hey,

sorry it took me a little bit longer.
I found a very small typo in one of the example files and encountered a missing install statement in the `CMakeLists.txt` which I missed last time (sorry)
Other then that I think everything looks good!

GDMarco:
Hi Jan-Niklas, 

I was reviewing it and I realise that your version that is pending for merging approval would need something like: 

`%include "crpropa/Geometry.h"
%implicitconv crpropa::ref_ptr<crpropa::Surface>;
%template(SurfaceRefPtr) crpropa::ref_ptr<crpropa::Surface>;`

in the `2_headers.i` file. It is needed to define `Surface` objects. 

Please, let me know if you spot also this issue.

Best, 
Gaetano

JanNiklasB:
Oh, I see what you mean, that would indeed be missing.

I had that issue for every `ref_ptr` occurence in one of my experimental builds where I replaced the `Referenced` base class with a `shared_ptr`.

I personally work mainly with pybind11 so I think I do not fully understand how SWIG works, but it seems SWIG runs into problems when handing a raw pointer to a `ref_ptr` which then is automatically converted to a `ref_ptr` on c++ side. This can cause problems regarding the lifetime of objects on python side and on c++ side and might result in seg faults. However, doing a `%implicitconv` solves this problem since swig now knows to call `addReference` and `removeReference` whenever a `ref_ptr` object goes out of scope.

Grindegreen:
Hi everyone,

I have run a few tests with `CustomPhotonField`, including generating a custom photon field and the corresponding spatial grid. Some of the results are shown below. From my side, the modules appear to work as expected.

The only issues I encountered were the ones @GDMarco mentioned above regarding `Surface` reference pointers. These still need to be added for the interface to work correctly. In addition, there seems to be an older bug related to the `haveElectrons=False` option: candidates are not deactivated in `EMPairProduction` when secondaries are not sampled. After applying the corresponding fix, everything else works as intended.

spectrum_comparison

In the upper figure, I simulated photons between `Emin = 10 * GeV` and `Emax = 10 * PeV` using both `TabularPhotonField` and `TabularSpatialPhotonField` with a `1/r` scaling. I then compared the result from the spatial photon field to the expected spectrum obtained from the calculated interaction rate with the same scaling applied.

distance_absorption_check

In the second test, I injected monoenergetic `10 * TeV` photons and compared the absorption obtained with both modules to the value expected from the interaction rates. The results are consistent in this case as well.

I have also prepared a few Jupyter notebooks demonstrating how to generate grids for tabulated photon fields. I plan to add these in a separate pull request.

JulienDoerner (changes requested):
Hey @GDMarco, 

Thanks for your PR and adding to CRPropa. I think this will be a good extension of the framework. 
I still have three points which should be addressed before merging. 

1. With the general structure of your `PhotonBackground`, you introduce a new coordinate system that is somehow different than usual. All x-coordinates are inverted $x = -x$. This will start a left-handed system, which we do not want, and it complicates any kind of extension for custom photon fields. 
I would suggest changing the input data files to already account for the right geometry and use the regular coordinates instead. 

2.  Your example notebook does not run at the moment. On the one hand, there are the problems with the `Surface` and on the other hand, there are some typos and inconsistencies in the code. Also, you do not have any resulting plots in the version you committed. I would suggest re-running this and also including a final analysis. In this case, a user can already see what the final result could be. I would also suggest to add the information how to download the needed data and where exactly they have to be added

3. Warnings from the `EMPairProduction`: I tried to reproduce your example notebook and run the setup as stated there. In this case I get a lot of warnings about the tabulated ranges and the sample values: 
```
2026-06-05 08:30:48 [WARNING] EMPairProduction: Energy 1.53528e+14 eV is not in tabulated range
2026-06-05 08:30:48 [WARNING] EMPaiProduction: Sampled s value is the lowest tabulated value, which is not physical. The index j will be set to 1 to avoid division by zero.
2026-06-05 08:30:48 [WARNING] EMPaiProduction: Sampled s value is the lowest tabulated value, which is not physical. The index j will be set to 1 to avoid division by zero.
2026-06-05 08:30:48 [WARNING] EMPairProduction: Energy 2.24826e+13 eV is not in tabulated range
2026-06-05 08:30:48 [WARNING] EMPairProduction: Energy 5.96921e+13 eV is not in tabulated range
2026-06-05 08:30:48 [WARNING] EMPaiProduction: Sampled s value is the lowest tabulated value, which is not physical. The index j will be set to 1 to avoid division by zero.
2026-06-05 08:30:48 [WARNING] EMPairProduction: Energy 3.24214e+14 eV is not in tabulated range
2026-06-05 08:30:48 [WARNING] EMPairProduction: Energy 2.42937e+13 eV is not in tabulated range
2026-06-05 08:30:48 [WARNING] EMPaiProduction: Sampled s value is the lowest tabulated value, which is not physical. The index j will be set to 1 to avoid division by zero.
2026-06-05 08:30:48 [WARNING] EMPairProduction: Energy 1.50853e+14 eV is not in tabulated range
2026-06-05 08:30:48 [WARNING] EMPaiProduction: Sampled s value is the lowest tabulated value, which is not physical. The index j will be set to 1 to avoid division by zero.
2026-06-05 08:30:48 [WARNING] EMPairProduction: Energy 1.23812e+14 eV is not in tabulated range
2026-06-05 08:30:48 [WARNING] EMPairProduction: Energy 2.42169e+13 eV is not in tabulated range
``` 

Maybe you should recalculate the data in a broader energy range or check if the warnings can be suppressed if they are intended. Now, the number of warnings will create a massive slowdown due to a lot of printing. 


Also, there are some merge conflicts which have to be resolved.

JanNiklasB (approved):

include/crpropa/Candidate.h
```diff
@@ -66,15 +75,27 @@ class Candidate: public Referenced {
 	/**
 	 Creates a candidate, initializing the Candidate::source, Candidate::created,
 	 Candidate::previous and Candidate::current state with the argument.
+	 @param state  ParticleState for source, created, previous and current. Makes copies.
 	 */
 	Candidate(const ParticleState &state);
 
+	/** Checks if particle is still active */
 	bool isActive() const;
+	/** Sets particle active or unactive, in unactive state it is ignored by all modules */
```
JulienDoerner: This statement is not true. 
Only the next repetition of the module list will be avoided, but all modules can work on already deactivated particles.

include/crpropa/module/EMDoublePairProduction.h
```diff
@@ -23,53 +24,74 @@ namespace crpropa {
...
 	void setLimit(double limit);
-
+	
 	/** Apply thinning with a given thinning factor
 	 * @param thinning factor of thinning (0: no thinning, 1: maximum thinning)
 	 */
 	void setThinning(double thinning);
 	
+	/** Apply a surface that confine the position dependent photon field region.
+	 * @param surface closed surface to confine the grid to be  uploaded
```
JulienDoerner: Just codewise, this has not to be a closed surface. It can also be a plane.

include/crpropa/Geometry.h
```diff
@@ -27,11 +26,13 @@ class Surface : public Referenced {
...
 	 */
 	virtual double distance(const Vector3d& point) const = 0;
-	/**  Returns the normal to the surface at a point. Negative on the one side,
+	/** Returns the normal to the surface at a point. Negative on the one side,
 	 positive on the other. For closed surfaces it is negative on the inside.
 	 @param point	vector corresponding to the point to which compute the normal vector
 	 */
 	virtual Vector3d normal(const Vector3d& point) const = 0;
+	/// Checks whether the given point is inside of the surface
+	virtual bool isInside(const Vector3d& point) const = 0;
```
JulienDoerner: I do not see the need for this function, the distance < 0 is already checking whether the position is inside a closed surface. And in the case of non-closed surfaces like the plane, an inside definition does not make sense.

test/testInteraction.cpp
```diff
@@ -710,6 +711,15 @@ TEST(EMPairProduction, allBackgrounds) {
 	em.setPhotonField(ebl);
 	ref_ptr<PhotonField> urb = new URB_Protheroe96();
 	em.setPhotonField(urb);
+
+	std::string dir = getDataPath("") + "Scaling/ISRF_Freudenreich98";
+	if (!is_directory(dir)) {
+		std::cout << "Photon background tables not available" << std::endl;
```
JulienDoerner: The ISRF model will not be a default data object for all users. Therefore, it will, in most of the cases, give this error message. This message should be improved to state that this is somehow expected, or remove the test of non-default data.

test/testPythonExtension.py
```diff
@@ -1,385 +1,384 @@
...
+		H2Density = 1
+		nucleonDensity = 0.5
+		massDensity = CustomMassDensity(density, HIDensity, HIIDensity, H2Density, nucleonDensity)
+		pos = crp.Vector3d(1, 0, 0)
+		self.assertEqual(density, massDensity.getDensity(pos))
+		self.assertEqual(HIDensity, massDensity.getHIDensity(pos))
+		self.assertEqual(HIIDensity, massDensity.getHIIDensity(pos))
+		self.assertEqual(H2Density, massDensity.getH2Density(pos))
+
+	def testCustomPhotonField(self):
```
JulienDoerner: This test exists twice and therefore one of them will not be executed.

test/testPythonExtension.py
```diff
@@ -1,385 +1,384 @@
...
+				return self.isPositionDependent
+
+		photonDensity = 10
+		photonField = CustomPhotonField(photonDensity)
+		energy = 10*crp.GeV
+		z = 0
+		self.assertEqual(photonDensity, photonField.getPhotonDensity(energy, z))
+		self.assertEqual('testCustomPhotonField', photonField.getFieldName())
+		self.assertEqual(True, photonField.hasRedshiftDependence(), photonField.hasPositionDependence())
+		self.assertEqual(True, photonField.hasPositionDependence(), photonField.hasPositionDependence())
```
JulienDoerner: This one should return a `False`. Also the arguments are not correct. The third one should be the message sent.

JanNiklasB:
I addressed these changes in this [Branch](https://github.com/JanNiklasB/CRPropa3/tree/PhotonFieldPR).
I think it would be best if you @JulienDoerner  directly merge those smaller changes.

JulienDoerner (approved):
Thanks, @JanNiklasB, for addressing these comments, and @GDMarco, for all your work in preparing the PR. 
From my point of view, this PR is now ready for merging. I will approve it now, but leave it here for a few more days until we merge. Maybe someone else wants to comment.

PR #537, opened 2026-04-01
Many bug and warning fixes

JanNiklasB:
Hey,

with this PR I want to introduce the new CXX standard 23 to CRPropa, it enables us to use more modern code in future additions which should make it easier to contribute.
Furthermore CXX23 introduces the new deducing this which enables us to use CRTP (Curious Reoccurring Template Patterns) instead of vtables while keeping the fundamental structures of CRPropa, this is important for an eventual GPU extension.

I also tried to tackle many warnings that arose in the past, in the process of that I discovered some major bugs which I also fixed here (see comments in code for more details):
- `type`  instead of `to_type` in `Variant.cpp`
- missing return in `Variant.cpp`
- deprecated / missing `ENDDO` statements in `sophia_interface.f` 
- broken doxygen commands
- depricated `sprintf` replaced with `snprintf`
- missing include path in `FindGooglePerfTools.cmake`
- missing brackets in `Random.cpp`
- not initialized `constScaleBendover` in preconstructor
- `extends` never set in `PeriodicMagneticField::setExtends`
- never set `Bphi` in `CMZField.cpp` if r<=r1
- issue between `clang` and `gcc` when formatting `uint64_t`, now using macro `"%10" PRIu64 "\t"` which set it automatically
- added more modern build tool `ninja` to workflows

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -3,7 +3,8 @@ cmake_minimum_required(VERSION 3.14)
 project(CRPropa Fortran C CXX)
 set(CRPROPA_RELEASE_VERSION 3.2.1+)  # Update for new release
 
-set(CMAKE_CXX_STANDARD 11)
+set(CMAKE_CXX_STANDARD 23)
+set(CMAKE_CXX_STANDARD_REQUIRED True)
 set(CMAKE_MODULE_PATH "${CMAKE_SOURCE_DIR}/cmake" ${CMAKE_MODULE_PATH})
 
 set(CRPROPA_EXTRA_SOURCES)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -3,8 +3,7 @@ cmake_minimum_required(VERSION 3.14)
 project(CRPropa Fortran C CXX)
 set(CRPROPA_RELEASE_VERSION 3.2.1+)  # Update for new release
 
-set(CMAKE_CXX_STANDARD 23)
-set(CMAKE_CXX_STANDARD_REQUIRED True)
+set(CMAKE_CXX_STANDARD 11)
 set(CMAKE_MODULE_PATH "${CMAKE_SOURCE_DIR}/cmake" ${CMAKE_MODULE_PATH})
 
 set(CRPROPA_EXTRA_SOURCES)
--- a/include/crpropa/AssocVector.h
+++ b/include/crpropa/AssocVector.h
@@ -122,8 +122,7 @@ namespace Loki
         typedef typename Base::const_reverse_iterator const_reverse_iterator;
 
         class value_compare
-            : public std::binary_function<value_type, value_type, bool>
-            , private key_compare
+            : private key_compare
         {
             friend class AssocVector;
 
--- a/include/crpropa/AssocVector.h
+++ b/include/crpropa/AssocVector.h
@@ -46,12 +46,12 @@ namespace Loki
 
     namespace Private
     {
-        template <class Value, class C>
+        template <class Value, class C, class K>
         class AssocVectorCompare : public C
         {
-            typedef std::pair<typename C::first_argument_type, Value>
+            typedef K first_argument_type;
+            typedef std::pair<first_argument_type, Value>
                 Data;
-            typedef typename C::first_argument_type first_argument_type;
 
         public:
             AssocVectorCompare()
@@ -98,10 +98,10 @@ namespace Loki
     >
     class AssocVector
         : private std::vector< std::pair<K, V>, A >
-        , private Private::AssocVectorCompare<V, C>
+        , private Private::AssocVectorCompare<V, C, K>
     {
         typedef std::vector<std::pair<K, V>, A> Base;
-        typedef Private::AssocVectorCompare<V, C> MyCompare;
+        typedef Private::AssocVectorCompare<V, C, K> MyCompare;
 
     public:
         typedef K key_type;
--- a/include/crpropa/AssocVector.h
+++ b/include/crpropa/AssocVector.h
@@ -110,14 +110,14 @@ namespace Loki
 
         typedef C key_compare;
         typedef A allocator_type;
-        typedef typename A::reference reference;
-        typedef typename A::const_reference const_reference;
+        typedef typename A::value_type& reference;
+        typedef const typename A::value_type* const_reference;
         typedef typename Base::iterator iterator;
         typedef typename Base::const_iterator const_iterator;
         typedef typename Base::size_type size_type;
         typedef typename Base::difference_type difference_type;
-        typedef typename A::pointer pointer;
-        typedef typename A::const_pointer const_pointer;
+        typedef typename A::value_type* pointer;
+        typedef const typename A::value_type* const_pointer;
         typedef typename Base::reverse_iterator reverse_iterator;
         typedef typename Base::const_reverse_iterator const_reverse_iterator;
 
--- a/libs/sophia/sophia_interface.f
+++ b/libs/sophia/sophia_interface.f
@@ -2568,45 +2568,55 @@ C...Calculate maximum weight for ND-particle decay
 
 C...generation of the masses, compute weight, if rejected try again
 240   RORD(1) = 1.D0
-      DO 260 IL1=2,ND-1
-      RSAV = RNDM(0)
-      DO 250 IL2=IL1-1,1,-1
-      IF(RSAV.LE.RORD(IL2))   GOTO 260
-250     RORD(IL2+1)=RORD(IL2)
+      DO IL1=2,ND-1
+        RSAV = RNDM(0)
+        DO IL2=IL1-1,1,-1
+          IF(RSAV.LE.RORD(IL2))   GOTO 260
+          RORD(IL2+1)=RORD(IL2)
+        ENDDO
 260     RORD(IL2+1)=RSAV
+      ENDDO
       RORD(ND) = 0.D0
       WT = 1.D0
-      DO 270 IL=ND-1,1,-1
-      PV(IL,5)=PV(IL+1,5)+P(IL,5)+(RORD(IL)-RORD(IL+1))*(PV(1,5)-PS)
-270   WT=WT*PAWT(PV(IL,5),PV(IL+1,5),P(IL,5))
+      DO IL=ND-1,1,-1
+        PV(IL,5)=PV(IL+1,5)+P(IL,5)+(RORD(IL)-RORD(IL+1))*(PV(1,5)-PS)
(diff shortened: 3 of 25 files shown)


include/crpropa/AssocVector.h
```diff
@@ -46,12 +46,12 @@ namespace Loki
 
     namespace Private
     {
-        template <class Value, class C>
+        template <class Value, class C, class K>
         class AssocVectorCompare : public C
         {
-            typedef std::pair<typename C::first_argument_type, Value>
+            typedef K first_argument_type;
+            typedef std::pair<first_argument_type, Value>
```
JanNiklasB, line 53: `first_argument` is removed in CXX23 so we need an additional template argument, the rest of the code is adjusted, user code is not affected as far as I see

include/crpropa/AssocVector.h
```diff
@@ -110,20 +110,19 @@ namespace Loki
 
         typedef C key_compare;
         typedef A allocator_type;
-        typedef typename A::reference reference;
-        typedef typename A::const_reference const_reference;
+        typedef typename A::value_type& reference;
+        typedef const typename A::value_type* const_reference;
```
JanNiklasB, line 114: `reference` and `const_reference` are removed in CXX23, `value_type` is equivalent

include/crpropa/AssocVector.h
```diff
@@ -110,20 +110,19 @@ namespace Loki
...
-        typedef typename A::const_pointer const_pointer;
+        typedef typename A::value_type* pointer;
+        typedef const typename A::value_type* const_pointer;
         typedef typename Base::reverse_iterator reverse_iterator;
         typedef typename Base::const_reverse_iterator const_reverse_iterator;
 
         class value_compare
-            : public std::binary_function<value_type, value_type, bool>
-            , private key_compare
+            : private key_compare
```
JanNiklasB, line 125: `binary_function` is not needed in newer CXX versions, its functionality is now done automatically and therefore it was removed

libs/sophia/sophia_interface.f
JanNiklasB, line 1: `ENDDO` is the more modern approach, the old one also works but throws warnings

src/magneticField/turbulentField/PlaneWaveTurbulence.cpp
```diff
@@ -59,7 +59,7 @@
 
 #ifdef ENABLE_FAST_WAVES
 #include <immintrin.h>
-#include <memory.h>
+#include <memory>
```
JanNiklasB, line 62: `<memory>` is more general, we had vocally reported issues in the past and `<memory>` seemed to fix them

src/magneticField/CMZField.cpp
```diff
@@ -192,7 +192,7 @@ Vector3d CMZField::getMCField(const Vector3d& pos) const {//Field in molecular c
         Bphi = - B1*exp(-z*z/Hc/Hc)*R/r;
     }
     else{
-        - B1*exp(-z*z/Hc/Hc)*R/r1*(3*r/r1- 2*r*r/r1*r1);
+        Bphi = - B1*exp(-z*z/Hc/Hc)*R/r1*(3*r/r1- 2*r*r/r1*r1);
```
JanNiklasB, line 195: was forgotten

src/magneticField/MagneticField.cpp
```diff
@@ -26,7 +26,7 @@ Vector3d &PeriodicMagneticField::getExtends() {
 	return extends;
 }
 
-void PeriodicMagneticField::setExtends(const Vector3d &origin) {
+void PeriodicMagneticField::setExtends(const Vector3d &extends) {
```
JanNiklasB, line 29: propably a typo

src/module/TextOutput.cpp
```diff
@@ -166,103 +167,104 @@ void TextOutput::process(Candidate *c) const {
...
+		p += std::snprintf(buffer + p, buffersize - p, "%8.5E\t",
 				c->getTime() / timeScale);
 
 	if (fields.test(RedshiftColumn))
-		p += std::sprintf(buffer + p, "%1.5E\t", c->getRedshift());
+		p += std::snprintf(buffer + p, buffersize - p, "%1.5E\t", c->getRedshift());
 
 	if (fields.test(SerialNumberColumn))
-		p += std::sprintf(buffer + p, "%10lu\t",
+		p += std::snprintf(buffer + p, buffersize - p, "%10" PRIu64 "\t",
```
JanNiklasB, line 187: `"%10" PRIu64 "\t"` is the same as `"%10lu\t"` in case of `gcc` and the same as `"%10llu\t"` in case of `clang` (that was a problem before)

src/Random.cpp
```diff
@@ -282,7 +282,8 @@ uint64_t Random::randInt64()
 {
 	int64_t a = randInt();
 	int64_t b = randInt();
-	return (b + a << 32);
+	// a is shifted to the left to create a proper 64 bit integer
+	return (b + (a << 32));
```
JanNiklasB, line 286: same as the old one, but does not throw a warning

src/Random.cpp
```diff
@@ -390,11 +391,11 @@ void Random::initialize(const uint32_t seed) {
...
 	uint32_t *p = state;
 	int i;
-	for (i = N - M; i--; ++p)
+	for (i = +N - M; i--; ++p)
 		*p = twist(p[M], p[0], p[1]);
 	for (i = M; --i; ++p)
-		*p = twist(p[M - N], p[0], p[1]);
-	*p = twist(p[M - N], p[0], state[0]);
+		*p = twist(p[+M - N], p[0], p[1]);
+	*p = twist(p[+M - N], p[0], state[0]);
```
JanNiklasB, line 398: enum-enum-conversion was removed, the `+` fixes that

src/Variant.cpp
```diff
@@ -668,6 +668,7 @@ Variant Variant::fromString(const std::string& s, Type t) {
 			std::string s;
 			v.push_back(s);
 		}
+		return Variant(v);
```
JanNiklasB, line 671: was missing for some reason

src/Variant.cpp
```diff
@@ -1010,7 +1011,7 @@ Variant::operator const std::vector<Variant>&() const {
 
 #define INT_FUNCTION(to_type, fun, to)                                                                                 \
 	to Variant::fun() const {                                                                                          \
-		switch (type) {                                                                                                \
+		switch (to_type) {                                                                                                \
```
JanNiklasB, line 1014: fatal typo

CMakeLists.txt
```diff
@@ -3,7 +3,8 @@ cmake_minimum_required(VERSION 3.14)
 project(CRPropa Fortran C CXX)
 set(CRPROPA_RELEASE_VERSION 3.2.1+)  # Update for new release
 
-set(CMAKE_CXX_STANDARD 11)
+set(CMAKE_CXX_STANDARD 23)
+set(CMAKE_CXX_STANDARD_REQUIRED True)
```
JanNiklasB: Throws an error during configure step if the cmake standard is not met instead of some undefined behavior during build time, that should help the user to determine potentially outdated CXX standards

rafaelab:
Hi @JanNiklasB 

Thanks for the PR.

* `AssocVector` is actually not needed, and neither is `ref_ptr`. A good redesign (e.g., CRPropa 4) would get rid of them and use the smart pointers from C++. The same is true for variants, since newer C++ standards already handle them quite well.

* Ninja is NOT available by default in some systems, so this might be adding more complexity to CRPropa and making it harder for some people to install, especially in clusters.

* You say you introduced C++23, but I don't see this strict requirement in cmakelists. If you say C++23 thinking of CRTP only, CRTP is a pattern applicable to other C++ standards. If C++23 it becomes a strict requirement, it might be problematic for older computing clusters. Therefore, while I'm totally in favour of updating CRPropa and use ≥ C++20, I have mixed feelings about doing that now instead of investing in a redesign.

JanNiklasB:
Hey,
thanks for the quick answer.

- I would also be in favor of a redesign regarding AssocVector and ref_ptr for CRPropa4. But it feels like CRPropa 4 is still not that close, or is there something planned for the close future?

- True, Ninja is not a system default, but it is easily installable over pip or local package managers like conda. 
Since it speeds up the compilation process of our workflows and is as easily available as numpy which we already require I thought it would be better to use it in the workflows, but I could of course reverse that change.

- I was mainly thinking about "deducing this" or [Explicit object member functions](https://en.cppreference.com/w/cpp/language/member_functions.html#Explicit_object_member_functions) which where introduced in CXX23.
I thought setting `set(CMAKE_CXX_STANDARD 23)` would be enough to make the requirement, we certainly can use the new features with this flag.
I'm only working on newer clusters, but it seems to me that when working on clusters it is common to use package managers like conda, which normally enable the possibility of higher CXX standards despite an older system.
Or would that be a hardware issue?
However, I would be ok to reduce the CXX standard to 20 for now, as long as we get rid of the deprications.

lukasmerten:
Hi @JanNiklasB 

I agree with @rafaelab that we should be careful with backward compatibility on older compute clusters. At a minimum we should include a warning in the next v3.3 release notes that we will increase the minimum required version in the future. In this way, people will have a relatively recent official release as a fall back option, when their cluster does not support cxx20/23.

How far CRPropa v4 is in the future depends in the end on us as the developers and can/should probably discussed in the next in person meeting.

For ninja, I guess that people will not have issues to install that if they manage to install numpy. However, I suggest to wait with that until after the v3.3 release.

JanNiklasB:
Ok, I still would like to at least get rid of the warnings and the few typo caused bugs, should I modify this PR or make a new one without the changes to `CMAKE_CXX_STANDARD` and workflows?

lukasmerten:
Fair enough. Fixing the typos and small bugs sounds good to me. 
Do what ever is easier for you: updating this PR or creating a new one does not matter to me.

JanNiklasB:
I reversed the CXX23 requirement and ninja from the worklows, and changed the title accordingly.

JanNiklasB:
I am not sure why the ubuntu64 / 64x test failed, it happened in `testInteraction` in line 533 where the secondary size was equal to zero, but since this was no problem in the other tests and there is a probability* (even though it is small) that no secondary is generated in this test, I think it is fine.

*:
`ElasticScattering.cpp` line 104-108:
```
// check for interaction
Random &random = Random::instance();
double randDist = -log(random.rand()) / rate;
if (step < randDist)
	return;
```

lukasmerten:
I re-ran the test and it passed just fine. I'll probably have time to check and approve the PR this afternoon.

lukasmerten (approved):

PR #540, opened 2026-04-14
Relative path fix

JanNiklasB:
Hi,

this PR fixes the mentioned issue in #523 by changing all paths that were not already defined relative to `CMAKE_CURRENT_SOURCE_DIR` and `CMAKE_CURRENT_BINARY_DIR` accordingly (if needed).
Furthermore, it makes the variable `CRPROPA_EXTRA_INCLUDES` to a cache variable which exposes it to main projects which otherwise wouldn't be able to deduce the location of library includes automatically.
`CRPROPA_EXTRA_INCLUDES` is marked as advance variable so it does not appear in the ccmake without pressing "t".


rafaelab:
Hi @JanNiklasB 

The changes `#include "header.h"` to `#include "<header.h>"` work just fine, but they go against common C++ conventions. That is fine for external header, but that is uncommon for "owned" codes.

I don't know if I completely understood the point of the PR. 
A simpler solution would be to install the external libraries with `make_install`. 
Then it would work out of the box, right? I've done that in my own branch, and it works just fine both in my laptop and in the cluster.

JanNiklasB:
Hi

Changes like  `#include "header"` to `#include <header>` where only something I wanted for usage of binder, a tool to automatically generate pybind11 code for larger projects. I intended to not include that change in this PR but I forgot to remove it, sorry.

If you would manually install crpropa and then reference to the install folder there is no problem, but if you include crpropa without installing over `add_subdirectory` it would not be properly installed and can not be included in the top level project like the installed version.

An example for this would be https://gist.github.com/JanNiklasB/ca187681f446d323bdc773c425d2e57f .

rafaelab:
There are still a few `<header>` that should be reverted back to the original.
If I understood everything correctly, this PR should be just a change to CMakeLists.txt, and another to `src/Common.cpp`.

It runs just fine on my laptop with manual installation, so after things are reverted back to the original, it could be merged. 
Maybe it is easier to open another PR with the specific changes, based on the current version, to keep the git history clean.

JanNiklasB:
I reverted back all unrelated changes.
I think to keep the git histoy clean it would be the easiest to just squash the commits during the merge

JulienDoerner (approved):

PR #542, opened 2026-05-26
Anaconda Workflows

JanNiklasB:
Hi,

this PR introduces a workflow that automatically generate and publish conda packages to anaconda.org.
To make this work I added the folder `conda-bld` where all necessary files are included.
Furthermore the workflow itself is added with `.github/workflows/create_conda_package.yml`
Some additional changes are:
- moved `#define NPY_NO_DEPRECATED_API NPY_1_7_API_VERSION` to the correct position in `Vector3.h` so we avoid the warning.
- added limitation to `testInteraction.cpp` in the `TEST(SynchrotronRadiation, PhotonEnergy)` test so we won't need more memory then the github jobs can provide (More memory is needed during package creation)
- I noticed that when moving the tests during package building the recently from me added absolute paths for the `add_test` statements are expected to be relative by ctest, so I changed those lines back. This should not cause any problems when CRPropa3 is included as subproject, since ctest just uses the `subdirs("CRPropa3")` command which points to the nested CRPropa3 binary folder.
- Added `!conda-bld/build.sh` to `.gitignore`

Otherwise some package features that would be nice are still missing, but can be added later:
- OSX support for macs -> For this I would need a way to debug path problems that occur on OSX, but for that I need a machine to test it on, which is currently not available to me.
- Better SIMD support -> The SIMD instruction are compiled on the machine the package is created, we are using `fma+avx` currently, which is supported on a wide range of x64 CPUs, but the `native` flag could mean better performance on newer machines. However, this would require us to compile the SIMD code on the host machine and not the compile machine. For that I still need to find a good solution other then to completely compile CRPropa on package installation.

To make the workflow work some credentials need to be added to `CRPropa/CRPropa3`:
- ANACONDA_ACCOUNTNAME
- ANACONDA_TOKEN

The token needs to be generated on anaconda.org (either @JulienDoerner or I can do this), it then needs to be added with the above name under the github secrets section.
The account name is just `crpropa`


JulienDoerner (approved):
Hey, 

This PR looks good for me. I would suggest adding a manual deployment of the action as well.

PR #544, opened 2026-05-27
Bugfixes workflows

JanNiklasB:
Hey,

@JulienDoerner and I worked on some bugfixes for the current workflows:
- splitted conda workflows in two files to be able to restart them separately
- fixed `.github/workflows/create_coverage_report.yml` workflow and updated to ubuntu24
- fixed `.github/workflows/create_documentation.yml` workflow and updated to ubuntu24
- fixed `.github/workflows/deploy_new_documentation.yml` workflow and updated to ubuntu24
- removed `REQUIRED` flag from python since we can compile without it
- added extra python hints so cmake is able to find virtual envs and conda environments first
- added `--repeat until-pass:3` to all ctest (was addressed in #456). Since we checking the success of stochastic processes it is possible that some tests fail randomly. With `--repeat until-pass:3` we can counter that behavior by making it extremely unlikely tests fail 3 times in a row.

Links to successfully run workflows:
- [create-documentation](https://github.com/JanNiklasB/CRPropa3/actions/runs/26526591326/job/78131787825)
- [create-coverage-report](https://github.com/JanNiklasB/CRPropa3/actions/runs/26514666007/job/78088314917)

### DIFF
--- a/.github/workflows/create_conda_package_arm.yml
+++ b/.github/workflows/create_conda_package_arm.yml
@@ -13,17 +13,12 @@ on:
   release:
     types: [published]
   workflow_dispatch:
-    branches:
-      - master
-      - releases/**
   
 
 jobs:
   arm:
     if: github.repository=='CRPropa/CRPropa3'
     runs-on: ubuntu-22.04-arm
-    strategy:
-      fail-fast: false
     steps:
       - name: Checkout repository
         uses: actions/checkout@v4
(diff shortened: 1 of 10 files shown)


JanNiklasB:
The last commit adds hints for `find_package(python)` that enables cmake to use the python version in virtual environments and conda environments first if they are active.
The line `set(Python_FIND_STRATEGY "LOCATION")` is only needed for us since we use an older cmake standard, in cmake standards larger then 3.15 it is set by default.

lukasmerten (approved):

PR #547, opened 2026-06-12
Improved handling of SWIG in cmake file and added proper stubs

JanNiklasB:
Hi,

this PR is an improvement on our current swig implementation.
I changed the manual swig command target to the from swig provided `swig_add_lib`.
With this cmake is now able to append additional flags, includes, etc. if needed and can decide if a rebuild of our swig interface is necessary.
Furthermore, due to something I do not fully understand, swig now provides us a much better `crpropa.py` file that has properly documented functions, this enables us with the help of `pybind11-stubgen` to generate python stubs from this file which will result much better linting.


JanNiklasB:
It should also now be possible to create the crpropa conda package with python3.14, I appended one building example.
[build.log](https://github.com/user-attachments/files/28887322/build.log)

lukasmerten (approved):

PR #545, opened 2026-05-28
Update of Installation.md

JanNiklasB:
With the now officially provided conda package and the upcoming CRPropa3.3 release it is time to update the installtion.md.
I restructured the instruction with a first part explaining how to install the conda package.
Then I explain the different requirements and cmake flags if the user want to build from source.
I further list the packages required for a global installation on Debian/Ubuntu, Fedora, Archlinux and if conda is used.
For MacOS I used the old description since I have no way of checking if it still works (which I also mention in the instruction)
For Windows I just say that we do not support that currently.
Finally I end with a Notes section where I mention some common problems and some notes to additional installation steps like the stubgen over pybind11-stubgen .

PR #548, opened 2026-06-12
Macos conda fixes + new workflow

JanNiklasB:
Hi,

In this PR I want to introduce fixes for macos which will enable the compilation of CRPropa3 in a conda environment on macos without additional complications (see #545 for install instructions).

Since this enables us to create builds in conda environments, we can also now create conda packages for macos,
for this I added the corresponding workflow (see [here](https://github.com/JanNiklasB/CRPropa3/actions/runs/27436345597/job/81099161125) for a test run [macos action](https://github.com/JanNiklasB/CRPropa3/actions/workflows/create_conda_package_macos.yml) ).

The fixes enable us to use C++11 now in the test. I adjusted the test even though #535 will introduce the newer C++17 standard, but this way we have at least some version that supports C++11 on macos.

This PR also depends on #547 since I needed a way to add the special linking flags for macos.
(This was very hard to figure out, [this saved me in the end](https://github.com/Mizux/cmake-swig/blob/8e6e6e43610a67d8e4c29cb85a6d7f2dde56f8ec/Foo/python/CMakeLists.txt#L18).

Here is another [build.log](https://github.com/user-attachments/files/28893781/build.log) that shows the building process on a x64 macos machine, so even though we do not have a github runner for that it is possible to build on those machines


CMakeLists.txt
```diff
@@ -290,13 +290,13 @@ if(APPLE OR USE_ABSOLUTE_RPATH)
   set(CMAKE_INSTALL_RPATH_USE_LINK_PATH TRUE)
   set(ABSOLUTE_RPATH "${CMAKE_INSTALL_PREFIX}/lib${LIB_SUFFIX}")
   if(NOT IS_ABSOLUTE ${ABSOLUTE_RPATH})
-    set(ABSOLUTE_RPATH ${CMAKE_CURRENT_BINARY_DIR}/${ABSOLUTE_RPATH})
+    set(ABSOLUTE_RPATH ${CMAKE_CURRENT_DIR}/${ABSOLUTE_RPATH})
```
JanNiklasB: This should be `${CMAKE_BINARY_DIR` not `${CMAKE_CURRENT_DIR}`

lukasmerten:
Without going into any details, this looks good to me.
However, we should probably switch to a newer version of the checkout-action 

> Warning: Node.js 20 actions are deprecated. The following actions are running on Node.js 20 and may not work as expected: actions/checkout@v4, actions/upload-artifact@v4. Actions will be forced to run with Node.js 24 by default starting June 16th, 2026. Node.js 20 will be removed from the runner on September 16th, 2026. Please check if updated versions of these actions are available that support Node.js 24. To opt into Node.js 24 now, set the FORCE_JAVASCRIPT_ACTIONS_TO_NODE24=true environment variable on the runner or in your workflow file. Once Node.js 24 becomes the default, you can temporarily opt out by setting ACTIONS_ALLOW_USE_UNSECURE_NODE_VERSION=true. For more information see: https://github.blog/changelog/2025-09-19-deprecation-of-node-20-on-github-actions-runners/

lukasmerten (approved):

PR #554, opened 2026-06-22
fix and enhancement to secondary propagation in ModuleList

### LINKED ISSUE #553: secondariesFirst not working with Source or CandidateVector
the secondariesFirst parameter does not actually propagate secondaries at each step if run is called with a Source or CandidateVector

teeteg:
Hello,

this pull request aims at solving some performance problems with the secondary handling in the `run` function of the `ModuleList`.

- The `secondariesFirst` parameter is now correctly passed on to the call for the candidate, allowing to actually propagate secondaries at each step while using a `Source` or `CandidateVector`.
(closes #553)

- Secondaries are now submitted as OpenMP tasks, allowing multiple threads to work on the secondaries of one primary. This is allows for an equal cpu load for cases of inhomogeneous secondary creation (e.g. cascades).

- A new parameter `waitForSecondaries` is introduced to control if a thread with `secondariesFirst=true` should wait for secondary completion (mimicking the previous intended behaviour) or should continue propagating the primary. The difference to `secondariesFirst=false` then boiling down to secondaries being submitted on the fly or as a badge at the end.

- During propagation with `secondariesFirst=true` the number of already submitted/propagated secondaries is now internally saved with the `taskedSecondaries` parameter. The for-loop over the secondaries will now only loop over new secondaries, avoiding the need to loop over all retired pointers each step. This is necessary for a correct distribution of work over the tasks but should also reduce unnecessary calculations for primaries that have created many secondaries over their lifespan. 
However this removes the possibility of revisiting old secondaries. As `secondariesFirst` is not working anyway, this shouldn't really be impactful.

During my testing these changes lead to either an equal or a better performance.  Constructing a single primary secondary cascade shows that all cores are used properly.
Especially the fixed `secondariesFirst` (in combination with `waitForSecondaries=true`) lead to general improvements, which might have something to do with memory/cache locations.
Exact gains are of course dependent on simulation and machine.

Furthermore, all local tests are still running successfully. Testing on my simulations showed no change in physics outcome.


JanNiklasB (approved):
It looks good to me, the required OMP should be available by default in any gcc that also has C++17 which will be the minimum C++ standard we support with 3.3.

JulienDoerner (approved):
Thanks @teeteg for this good improvement. This will help to speed up simulations.

PR #555, opened 2026-06-24
Plugin example rework and workflow conditions

JanNiklasB:
Hi,

with this PR I want to update the current plugin template.
This includes the following changes:
- updated and more detailed cmake file with apple fixes
- update project structure
- infos and examples on how to use namespaces and tests
- updated Readme
- option for builtin crpropa
- Findcrpropa.cmake
- workflow to test plugin-template

Futhermore I added more conditions when to run the workflows with the Idea to only run them when a related file is changed.

The large number of files comes mostly because I also added gtest to the plugin-template folder


plugin-template/CMakeLists.txt
```diff
@@ -1,61 +1,292 @@
-cmake_minimum_required(VERSION 2.8.12)
+cmake_minimum_required(VERSION 3.14)
 
-# Change name of the library to something meaningful!
-set(PLUGINNAME "myPlugin")
-project(${PLUGINNAME} C CXX)
+# set the version to whatever you want
+# also change the name
+project(myPlugin VERSION 1.0.0 LANGUAGES Fortran C CXX)
```
JulienDoerner, line 5: Why do we also need Fortran support in the Plugin?

plugin-template/CMakeLists.txt
```diff
@@ -1,61 +1,292 @@
...
+if(USE_CRPROPA_BUILTIN)
+	if(NOT crpropa_POPULATED)
+		# needed to fetch builtin version	
+		include(FetchContent)
+		
+		# pulls repository specified and can even be pinned on specific commit
+		FetchContent_Declare(
+			crpropa
+			GIT_REPOSITORY "https://github.com/CRPropa/CRPropa3.git"
+			# GIT_TAG f5cf064cf60b8a4771deab6d963b01d44df13c3f
```
JulienDoerner, line 50: Is this an old version that can be removed? Otherwise, it should be commented on why this is here.

.github/workflows/test_examples.yml
```diff
@@ -1,5 +1,25 @@
 name: crpropa-example-test
-on: [push, pull_request]
+on:
+  push:
+    paths:
+      - ".github/workflows/test_examples.yml"
+      - "**.cpp"
+      - "**.h"
+      - "**.py"
+      - "**.i"
```
JulienDoerner, line 9: Those paths do not check whether the example notebook (`doc/pages/example_notebooks/**.iypnb`) are changed.

JulienDoerner (commented):
Hey @JanNiklasB 

Thanks for the work you have done. 
I left some small comments that should be addressed. 

Besides this, I would like to ask you to add your changes to the `CHANGELOG.md` and add the new test to the `README.md` to indicate the status.

plugin-template/CMakeLists.txt
JanNiklasB: We don't, I just copied the languages from the main `CMakeLists.txt` and forgot to remove Fortran

plugin-template/CMakeLists.txt
JanNiklasB: With the GIT_TAG option you can specify the commit and therefore the branch you want to use. I will add a comment here for a better description and I think it would be good to mention this in the Readme

plugin-template/CMakeLists.txt
JanNiklasB: Ok, even though we do not need it on linux kernels we do "need" it on macos (see failing test). There is the option to remove the `-gfortran` flag, however to ensure full compatibility it is still needed.
Furthermore, to include it in our plugin shows the plugin developer how to handle compilation on macos.

JanNiklasB:
Sorry, I just realize that the conda workflows are reporting they are running successfully even though they are not building the packages sometime. For that reason I splittet them to easier restart them, but I think I found a way to properly report if they are failed and no package could be uploaded, I will test it and add the changes to this PR since it already touches the workflows.
I comment again to signal that I was successfully, sorry again for the inconvinience

JanNiklasB:
The workflow is now combined again and has a check included that checks if any package exists and throws an error if not, that should help us to notice if any packages were not built.

.github/workflows/test_examples.yml
JulienDoerner: These are still missing. I guess this is related to your force push, which overwrote your previous commit.

JulienDoerner: Oh, I see it is in the next line.

JulienDoerner (approved):
This looks good for me now. There is only one merge conflict left.

JanNiklasB:
Ok, now every merge conflict should be resolved, and `testMomentumDiffusion()` contains
```py
if __name__ == '__main__':
	unittest.main()
```
again.

PR #556, opened 2026-06-25
Expanding nuclear limits to above iron mass (A> 56)

mohller:
Implementing super-heavy nuclei to allow the code to use the updated data files. This PR is directly connected to PRs https://github.com/CRPropa/CRPropa3-data/pull/26 and https://github.com/CRPropa/CRPropa3-data/pull/27.

### New features:
* Extended nuclear coverage in `PhotoDisintegration` and `NuclearDecay` from Fe-56 (Z≤26, N≤30) to Pb-214 (Z≤82, N≤132). Table sizes grow from 837 to 11,039 placements (although the actual content of these tables is less, typically ~10%).
* Added named constants `NUCLEAR_ZMAX` (82), `NUCLEAR_NMAX` (132), and `NUCLEAR_NSTRIDE` (133) to `Common.h`, replacing the hard-coded values in previous versions (26, 56) scattered across `PhotoDisintegration`, `NuclearDecay`, and `ParticleMass`.
* `PhotoDisintegration` has an optional argument to load superheavy-nucleus rates and branching tables (covering A≥56 up to Pb) via the `superheavy` flag. When `true`, the files `rate_<field>_superheavy.txt` and `branching_<field>_superheavy.txt` are loaded instead of the standard ones. By default the flag is set to false.

### Summary of changes in class or function calls:
* `PhotoDisintegration` constructor gains an optional fourth argument: `bool superheavy = false`.
* `PhotoDisintegration::setPhotonField()` gains an optional second argument: `bool superheavy = false`.


src/module/PhotoDisintegration.cpp
```diff
@@ -17,18 +18,23 @@ const double PhotoDisintegration::lgmin = 6;  // minimum log10(Lorentz-factor)
...
-PhotoDisintegration::PhotoDisintegration(ref_ptr<PhotonField> f, bool havePhotons, double limit) {
-	setPhotonField(f);
+PhotoDisintegration::PhotoDisintegration(ref_ptr<PhotonField> f, bool havePhotons, double limit, bool superheavy) {
+	setPhotonField(f, superheavy);
 	this->havePhotons = havePhotons;
 	this->limit = limit;
 }
 
-void PhotoDisintegration::setPhotonField(ref_ptr<PhotonField> photonField) {
+void PhotoDisintegration::setPhotonField(ref_ptr<PhotonField> photonField, bool superheavy) {
```
JulienDoerner, line 27: I would suggest implementing an independent set-function for the `bool superheavy` instead of adding this here. This would match the procedure in the code elsewhere.

JulienDoerner (changes requested):
Hey @mohller, 

This is a nice PR. 

You should add your changes to the `CHANGELOG.md` to keep track of all changes.
Also, you should adjust the test in `test/testInteraction.cpp` to test the new superheavy particles.  This will, of course, only work if your PR in the data is merged and the installation is updated with the new data file. I would suggest keeping this PR open until this has happened.

lukasmerten:
I have merged the connected PR in crpropa-data. However, I currently have no time to create, check and upload the new data tarball before next week. So if anyone is faster than that, please go ahead.

JulienDoerner:
I have started the calculation of the data tarball. If this is done, I will upload the data and change the data download in this PR.

JulienDoerner:
The data are now available, and I adjusted the data download.

mohller:
The requested tests were implemented and verified locally. However, an update to the photodisintegration data (see https://github.com/CRPropa/CRPropa3-data/pull/29) is needed in order for the superheavy nuclei to pass, as the current data version doesn't include this fix.

src/module/PhotoDisintegration.cpp
mohller: Could you point to an example in the code where such procedure is followed? I am not familiar with cases where such a change in the interaction module would be desirable. In this case, the superheavy flag implies a complete change in the data file, making this a rather substantial change in the instance rather than just switching its internal behavior or logic.

test/testInteraction.cpp
```diff
@@ -508,6 +564,162 @@ TEST(PhotoDisintegration, interactionTag) {
 	EXPECT_TRUE(pd.getInteractionTag() == "myTag");
 }
 
+// PhotoDisintegration - Superheavy extension ---------------------------------
+namespace {
```
JulienDoerner: here is no need for this namespace, which is also unnamed.

test/testInteraction.cpp
```diff
@@ -508,6 +564,162 @@ TEST(PhotoDisintegration, interactionTag) {
...
+	EXPECT_DOUBLE_EQ(E_before, c.current.getEnergy());
+}
+
+
+TEST(PhotoDisintegration, superheavy_lead) {
+	// Test if a Pb-208 nucleus photo-disintegrates at least once over 1 Gpc
+	// using the superheavy CMB tables.  A, Z, and energy conservation are
+	// verified across all interactions in the step.
+	// This test can stochastically fail.
+	if (!superheavyFilesPresent())
```
JulienDoerner: This files will always be present as they are included in the default CRPropa download. We do not check data availability in other modules as well. 
I would suggest removing this condition

test/testInteraction.cpp
```diff
@@ -508,6 +564,162 @@ TEST(PhotoDisintegration, interactionTag) {
...
+	// proton number conserved
+	EXPECT_EQ(82, Z);
+	// energy conserved (EXPECT_NEAR: multiple interactions accumulate ~1e-12 J rounding)
+	EXPECT_NEAR(2000 * EeV, E, 1e-9);
+}
+
+TEST(PhotoDisintegration, superheavy_allHeavyIsotopes) {
+	// With superheavy=true, processing any nucleus in Z=27..NUCLEAR_ZMAX must
+	// not crash, even for isotopes whose slots are empty in the rate table.
+	if (!superheavyFilesPresent())
```
JulienDoerner: same as before

JulienDoerner (changes requested):
Hey @mohller 
I have checked your new tests. With my newly build data (https://github.com/JulienDoerner/CRPropa3-data/actions/runs/28922077586) based on your new PR in CRPropa3-data, all tests are running. 
Once the PR is merged there, we have to adjust the data version, but I can take care of this.

JulienDoerner (approved):
I have updated the data file. Now the tests are running for my local setup.

lukasmerten:
@mohller are we good to go or is anything still missing?

mohller:
> @mohller are we good to go or is anything still missing?

All complete from my side :)

PR #558, opened 2026-06-29
Reducing Conda Package Size

JanNiklasB:
Hey,

@JulienDoerner  and I noticed that the size of our conda package was each around 45 MB.
This is in itself not much, but since we generate 15 packages with each release we are at 675 MB for each version.
Anaconda gives us I think 5 GB of storage for free, so in a few versions we reach that limit.

The size of 45 MB is caused by the contained data files, without them we are at around 3 MB per package and 45 MB per Version. So consequently I changed the conda package so the data files are not included in the package itself, but rather downloaded and extracted during the conda install process.

The change for `Version.h` is needed since otherwise I have no way of knowing which data files version is needed at the conda install step, since I do not have access to the `CMakeLists.txt` file anymore.

For a successful test run see https://github.com/JanNiklasB/CRPropa3/actions/runs/28377415325


JanNiklasB:
I was informed by @JulienDoerner  that we get a warning when installing our conda package. This was caused by line 2 in `deactivate.sh` which accidentally was `unset $CRPROPA_DATA_PATH_CONDA_BACKUP` instead of `unset CRPROPA_DATA_PATH_CONDA_BACKUP` .

I needed to do a force push since my previous fix did not include a unset for `CRPROPA_DATA_PATH` if it was empty before activating the environment, it does now.

JulienDoerner (approved):
I have tested it again, and I am able to build the package with all. Now all warnings are gone on the mac. I think this PR can be merged.

PR #559, opened 2026-06-30
Removing Galactic Magnetic Lenses

JulienDoerner:
In this PR the Galalctic Magnetic Lenses are removed. 
They are deprecated since our last release, `3.2.2`, and can be removed for `3.3`. 
This also removes the part in the documentation and the connected external libs `eigen` and `healpix` which we provided and have only been used for the lenses.


JanNiklasB:
Hey,

since you removed eigen we also need to remove the following occurences in `doc/pages/Installation.md`
- Under "CMake Flag Documentation" we need to remove the entry for `EIGEN_PATH` and `INSTALL_EIGEN`
- Under the "Conda" section `eigen` needs to be removed from the packages

I also found mentions in 'doc/pages/AdditionalRessources.rst' and `doc/pages/howto_cite_crpropa.rst` but I am not sure if they should be kept.

In `conda-bld/build.sh` line 26 (`-DINSTALL_EIGEN=OFF \`) 

In `conda-bld/meta.yaml` line 34 and 44 (`- eigen`)

JulienDoerner:
I have adjusted the scripts in `conda-bld` as suggested.

rafaelab (approved):
Thanks, @JulienDoerner .
It works just fine.

About @JanNiklasB's comment, I suppose there are no further issues with the conda compatibility.  
In any case, I think this this PR should be merged before #558 .

PR #561, opened 2026-07-01
Position-dependent photon field: fix node-file loading and add example notebooks

Grindegreen:
### General description

Follow-up to #535. After the `<filesystem>` -> `kiss` migration the spatially dependent photon field stopped finding its per-node files: `list_directory` returns bare filenames, but `TabularSpatialPhotonField` and `InteractionRatesPositionDependent` passed those straight to the file open, so `TabularSpatialPhotonField(name)` and the position-dependent `EMPairProduction` rates raised `could not open ..._node_...` even though the files were present.

This PR restores the loading and adds two example notebooks showing how to build and use such a field.

### Technical details

- `PhotonBackground.cpp` and `InteractionRates.cpp`: prepend the directory with `concat_path(dir, filename)` before opening (the same idiom `getDataPath` uses), instead of relying on the bare name from `list_directory`. This restores the pre-migration behaviour, where `directory_iterator.path().string()` gave the full path.
- Added `doc/pages/example_notebooks/photonfield_grid/`: `01_field_generation` (build and verify a 1/r-scaled field) and `02_position_dependent_propagation` (propagate a power-law beam through it, plus a grid-sampling convergence check), linked from `extending_crpropa.rst`.
- Both notebooks require the CRPropa3-data repository (like the custom-photon-field example), so they are excluded from `test_examples.yml`.
- Updated the changelog and the AdditionalResources page.

### Testing

Verified against the analytic expectation: central-vs-edge density ratio = 26.0 and the interaction-rate ratios equal f(r) at each node.


rafaelab:
Thanks, @Grindegreen.
This works for me.

I will wait until the end of the week before merging, in case the fix breaks what was working for other people.
@GDMarco: does this break anything for you?

rafaelab (approved):

PR #562, opened 2026-07-03
Expose interaction rates for electromagnetic processes

rafaelab:
This PR implements an explicit `getRate` method for all electromagnetic interactions.
This is similar to the `lossLength` implementation for photohadronic processes.
Ultimately, it allows users to directly access interaction rates for these processes, which are private in the current implementation.

This implementation supports an external plugin that requires this functionality.


Grindegreen:
Hey @rafaelab, looks good to me overall. I only added one small documentation-related suggestion about moving the “future release” note into a separate issue for better visibility.

lukasmerten (approved):

PR #564, opened 2026-07-07
Plugin improvements

rafaelab:
This PR improves the plugin examples, offering more information and flexibility for users developing their own plugins.

- Provide an advanced CMake finder for CRPropa.
- Expose and reorganise external libraries shipped with CRPropa (e.g., gtest, zstream-cpp, etc), allowing developers to directly link their plugins to them.
- Rename `USE_CRPROPA_BUILTIN` in plugin's CMakeLists to `USE_OWN_CRPROPA` to avoid confusion with SWIG's "builtin" feature.
- Minor bug fix related to the handling of variants in SWIG (this problem appears in the context of plugins).


plugin-template/CMakeLists.txt
```diff
@@ -1,171 +1,218 @@
...
+set(${pluginName}_EXTRA_SOURCES)
+
+# source files for compilation
+set(SOURCE_FILES
+	${CMAKE_CURRENT_SOURCE_DIR}/src/plugin.cpp
+)
+file(GLOB ${pluginName}_HEADERS 
+	"${CMAKE_CURRENT_SOURCE_DIR}/include/${pluginName}/*.h"
+)
+
```
JanNiklasB: `${pluginName}_HEADERS` and `SOURCE_FILES` are never used after this.

plugin-template/CMakeLists.txt
```diff
@@ -1,171 +1,218 @@
...
+	if(Python_FOUND)
+		message(STATUS "Python version: ${Python_VERSION}")
+		message(STATUS "Python executable: ${Python_EXECUTABLE}")
+		message(STATUS "Python headers: ${Python_INCLUDE_DIRS}")
+		message(STATUS "Python library: ${Python_LIBRARY}")
+		message(STATUS "Python site-packages: ${Python_SITE_PACKAGES}")
+	endif(Python_FOUND)
+
+	if(Python_NumPy_FOUND)
+		set(CMAKE_SWIG_FLAGS -DWITHNUMPY ${CRP})
```
JanNiklasB: I can't find where `{CRP}` is defined
```suggestion
		set(CMAKE_SWIG_FLAGS -DWITHNUMPY)
```

plugin-template/CMakeLists.txt
```diff
@@ -1,171 +1,218 @@
...
+		list(APPEND CRPropa_SWIG_DEFINES -DWITHNUMPY)
```
JanNiklasB: `CRPropa_SWIG_DEFINES` is never used in this project, it is already defined in CRPropa's CMakeLists.txt
```suggestion
```

plugin-template/CMakeLists.txt
```diff
@@ -176,119 +223,173 @@ if(ENABLE_PYTHON)
...
+			set_property(TARGET ${pluginName}-swig APPEND PROPERTY LINK_FLAGS "-flat_namespace -undefined suppress")
+		endif(APPLE)
+
+		# # if you prefer PyBind (instead of SWIG)
+		# find_program(PYBIND11_STUBGEN_EXECUTABLE NAMES "pybind11-stubgen")
+		# if(PYBIND11_STUBGEN_EXECUTABLE)
+		# 	message(STATUS "Found pybind11-stubgen at ${PYBIND11_STUBGEN_EXECUTABLE}")
+		# 	install(CODE "execute_process(COMMAND ${PYBIND11_STUBGEN_EXECUTABLE} --ignore-all-errors -o ${Python_SITELIB} ${pluginName})")
+		# 	install(CODE "execute_process(COMMAND echo -- Generated python stubs over pybind11-stubgen for ${pluginName})")
+		# endif(PYBIND11_STUBGEN_EXECUTABLE)
```
JanNiklasB: We also have this in the main `CMakeLists.txt`, it works with swig generated interface files. The comment from me `# if you prefer PyBind (instead of SWIG)` is just misleading, sorry.
If pybind11-stubgen is not found (we do not require it) then the stubs are not build, but crpropa still works without them.

So I think this should be included again.
However, it should be moved after the other installs so all needed libraries are installed when this command is called.

plugin-template/CMakeLists.txt
```diff
@@ -176,119 +223,173 @@ if(ENABLE_PYTHON)
...
-# the testing part. You should create proper tests for your code here:
+# the testing part, where tests for your code should be added
 if(ENABLE_TESTING)
 	enable_testing()
 
+	# temporary workaround for newer clang versions due to its handling of unwinding
+	#   see: https://github.com/google/googletest/issues/3062
+	if(CMAKE_CXX_COMPILER_ID STREQUAL "Clang" OR CMAKE_CXX_COMPILER_ID STREQUAL "AppleClang")
+		if (CMAKE_CXX_COMPILER_VERSION VERSION_GREATER_EQUAL 15)
+			add_definitions(-DCRPROPA_TESTS_SKIP_EXCEPTIONS)
```
JanNiklasB: CRPropa already defines this, but the plugin could define something similar 
```suggestion
			add_definitions(-D${pluginName}_TESTS_SKIP_EXCEPTIONS)
```

plugin-template/CMakeLists.txt
```diff
@@ -176,119 +223,173 @@ if(ENABLE_PYTHON)
...
+		endif (CMAKE_CXX_COMPILER_VERSION VERSION_GREATER_EQUAL 15)
+	endif(CMAKE_CXX_COMPILER_ID STREQUAL "Clang" OR CMAKE_CXX_COMPILER_ID STREQUAL "AppleClang")
+
+	# googletest (provided, see https://code.google.com/p/googletest/wiki/FAQ
+	#     Why is it not recommended use a pre-compiled copy of Google Test?)
```
JanNiklasB: Since we now do use precompiled gtests I would also remove this comment now

CMakeLists.txt
JanNiklasB: Could you move this codeblock
```cmake
if(ENABLE_PYTHON AND Python_FOUND)
  find_program(PYBIND11_STUBGEN_EXECUTABLE NAMES "pybind11-stubgen")
	if(PYBIND11_STUBGEN_EXECUTABLE)
		message(STATUS "Found pybind11-stubgen at ${PYBIND11_STUBGEN_EXECUTABLE}")
		install(CODE "execute_process(COMMAND ${PYBIND11_STUBGEN_EXECUTABLE} --ignore-all-errors -o ${Python_SITELIB} crpropa)")
    install(CODE "execute_process(COMMAND echo -- Generated python stubs over pybind11-stubgen for crpropa)")
	endif()
endif()
```
to line 600 after the other install commands?
I noticed that on first install it causes an install error since `libcrpropa.so` is not yet installed.

plugin-template/README.md
```diff
@@ -1,25 +1,29 @@
 # Plugin template
 
-To create your own CRPropa modules in C++ we recommend using plugins. Plugins are small programs that can be installed as a separate modules. 
+To create your own CRPropa modules in C++ we recommend using plugins. 
+Plugins are programs that can be installed as a separate modules. They allow users to extend CRPropa capabilities without having to modify and reinstall CRPropa.
```
JanNiklasB:
```suggestion
Plugins are programs that can be installed as separate modules. They allow users to extend CRPropa capabilities without having to modify and reinstall CRPropa. 
```

plugin-template/CMakeLists.txt
```diff
@@ -1,171 +1,218 @@
...
+	if(NOT CRPropa_POPULATED)
 		include(FetchContent)
 		
 		# pulls repository specified and can even be pinned on specific commit
 		FetchContent_Declare(
 			crpropa
 			GIT_REPOSITORY "https://github.com/CRPropa/CRPropa3.git"
-			# # The GIT_TAG option enbales you to specify a commit hash or specific tag, 
-			# # see https://cmake.org/cmake/help/latest/module/ExternalProject.html#git
+			# # The GIT_TAG option enbales you to specify a commit hash or specific tag, see:
```
JanNiklasB:
```suggestion
			# # The GIT_TAG option enables you to specify a commit hash or specific tag, see:
```

plugin-template/CMakeLists.txt
```diff
@@ -1,171 +1,218 @@
...
 )
-# include any other needed include, the standard ${CMAKE_CURRENT_SOURCE_DIR}/include folder was included above
-target_include_directories(${CMAKE_PROJECT_NAME} PUBLIC ${${CMAKE_PROJECT_NAME}_EXTRA_INCLUDES})
-# link crpropa and other needed libraries
-target_link_libraries(${CMAKE_PROJECT_NAME} ${${CMAKE_PROJECT_NAME}_EXTRA_LIBRARIES})
 
-# specify installation instructions
-install(TARGETS ${CMAKE_PROJECT_NAME} DESTINATION "${CMAKE_INSTALL_PREFIX}/lib")
-install(DIRECTORY "${CMAKE_CURRENT_SOURCE_DIR}/include/" DESTINATION "${CMAKE_INSTALL_PREFIX}/include" FILES_MATCHING PATTERN "*.h")
+# the plugin's own headers plus any extra includes (e.g. CRPropa's)
```
JanNiklasB:
```suggestion
# the plugin's own headers plus any extra includes (e.g. CRPropa's CMakeLists.txt)
```

CMakeLists.txt
```diff
@@ -251,33 +259,54 @@ endif(ENABLE_HDF5)
...
+#   if("${isSystemDir}" STREQUAL "-1")
+#     message(STATUS "Use absolute RPATH ${ABSOLUTE_RPATH}")
+#     set(CMAKE_INSTALL_RPATH "${CMAKE_INSTALL_RPATH}:${ABSOLUTE_RPATH}")
+#   endif("${isSystemDir}" STREQUAL "-1")
+# endif()
+
+# # Fix for linker error on mac, Issue: #147
+# if(APPLE)
+#   add_definitions(-arch x86_64)
+# endif(APPLE)
```
JanNiklasB: If the lines below solve the issues that previously were solved by the above RPATH comment, then we can delete the comment I think.

JanNiklasB (changes requested):
Hey,

those are very nice improvements, I like especially the improved `FindCRPropa.cmake`.

I just noticed some minor things plus an issue in the python stubs generation, for which `libcrpropa.so` can not be found on first install since it is installed after the stubs command is called (this was caused by one of my PRs), this can easily be fixed by moving the code block (see comments) after the other installs.

With Regards,
Jan-Niklas

rafaelab:
Thanks, @JanNiklasB.
I've implemented all your fixes, I guess
I was just a bit unsure about the `PYBIND_STUBGEN` part because this type of customisation should probably be optional, not the default. But you are right that it is harmless, so I brought it back.

Everything worked on my laptop and on the cluster. Since you seem to have a different setup, can you double check that you can install it?

CMakeLists.txt
JanNiklasB: This is moved now, but the original codeblock (line 547-552) still exists.

JanNiklasB:
I encountered an error while test building the conda package (maybe we should add a test that checks if the conda workflows run on every PR).
It was caused by using absoulte paths in `add_test`, where somehow the function added `testName` after `COMMAND`  as absolute path, it can be fixed by adding a `./` before `${testName}`.
I will add more detailed comments directly in code.

CMakeLists.txt
```diff
@@ -555,126 +560,101 @@ endif(ENABLE_PYTHON AND Python_FOUND)
...
-  add_test(testAdiabaticCooling testAdiabaticCooling)
+  enable_testing()
 
-  add_executable(testCandidateSplitting ${CMAKE_CURRENT_SOURCE_DIR}/test/testCandidateSplitting.cpp)
-  target_link_libraries(testCandidateSplitting crpropa gtest gtest_main pthread ${COVERAGE_LIBS})
-  add_test(testCandidateSplitting testCandidateSplitting)
+  function(addTest testName)
+    add_executable(${testName} ${CMAKE_CURRENT_SOURCE_DIR}/test/${testName}.cpp)
+    target_link_libraries(${testName} PRIVATE crpropa gtest gtest_main pthread ${COVERAGE_LIBS})
+    add_test(NAME ${testName} COMMAND ${testName})
```
JanNiklasB:
```suggestion
    add_test(NAME ${testName} COMMAND ./${testName})
```

CMakeLists.txt
```diff
@@ -555,126 +560,101 @@ endif(ENABLE_PYTHON AND Python_FOUND)
...
   # python tests
   if(ENABLE_PYTHON AND Python_FOUND)
-    CONFIGURE_FILE(${CMAKE_CURRENT_SOURCE_DIR}/test/testSimulationExecution.py ${CMAKE_CURRENT_BINARY_DIR}/testSimulationExecution.py COPYONLY)
-    add_test(testSimulationExecution ${Python_EXECUTABLE} testSimulationExecution.py)
 
-    CONFIGURE_FILE(${CMAKE_CURRENT_SOURCE_DIR}/test/testDiffusionSDE.py ${CMAKE_CURRENT_BINARY_DIR}/testDiffusionSDE.py COPYONLY)
-    add_test(testDiffusionSDE ${Python_EXECUTABLE} testDiffusionSDE.py)
+    function(addTestPy testName)
+      configure_file(${CMAKE_CURRENT_SOURCE_DIR}/test/${testName}.py ${CMAKE_CURRENT_BINARY_DIR}/${testName}.py COPYONLY)
+      add_test(NAME ${testName} COMMAND ${Python_EXECUTABLE} ${CMAKE_CURRENT_BINARY_DIR}/${testName}.py)
```
JanNiklasB:
```suggestion
      add_test(NAME ${testName} COMMAND ${Python_EXECUTABLE} ./${testName}.py)
```

JanNiklasB (changes requested):

plugin-template/CMakeLists.txt
```diff
@@ -1,171 +1,217 @@
...
-		NumPy
-	)
-
 	if(Python_FOUND)
-		# If we found python we also search for SWIG
-	 	find_package(SWIG 3.0 REQUIRED)
-		include(UseSWIG)
-		add_definitions(-D${CMAKE_PROJECT_NAME}_HAVE_PYTHON)
-		list(APPEND ${CMAKE_PROJECT_NAME}_SWIG_DEFINES -D${CMAKE_PROJECT_NAME}_HAVE_PYTHON)
+		find_package(SWIG 3.0 REQUIRED)
```
JanNiklasB: in line 83 there we already search for swig

plugin-template/CMakeLists.txt
```diff
@@ -176,119 +222,174 @@ if(ENABLE_PYTHON)
...
 			else()
 				message(STATUS "  development libraries: NOT found!")
-			endif()
+			endif(Python_Development_FOUND)
+
 			# numpy is required, otherwise Python_FOUND = false
 			message(STATUS "NumPy: Found!")
 			message(STATUS "  headers: ${Python_NumPy_INCLUDE_DIRS} (version ${Python_NumPy_VERSION})")
-		endif()
+		endif(NOT USE_OWN_CRPROPA)
```
JanNiklasB: from line 88 to 103 we already print the infos about the python and numpy version

plugin-template/CMakeLists.txt
```diff
@@ -176,119 +222,174 @@ if(ENABLE_PYTHON)
...
 
-		
 		#  use Python_INSTALL_PACKAGE_DIR if provided; otherwise, install in Python_SITELIB
 		set(Python_INSTALL_PACKAGE_DIR "${Python_SITELIB}" CACHE PATH "folder in which the python package is installed")
 		message(STATUS "  package install directory: ${Python_INSTALL_PACKAGE_DIR}")
 
-		set(CMAKE_SWIG_FLAGS -DWITHNUMPY ${CRP})
-		list(APPEND ${CMAKE_PROJECT_NAME}_SWIG_DEFINES -DWITHNUMPY)
-		list(APPEND ${CMAKE_PROJECT_NAME}_SWIG_DEFINES -doxygen)
+		set(CMAKE_SWIG_FLAGS -DWITHNUMPY)
```
JanNiklasB: The swig flag is also set in line 97 for numpy

plugin-template/CMakeLists.txt
```diff
@@ -176,119 +222,174 @@ if(ENABLE_PYTHON)
...
+
 			# numpy is required, otherwise Python_FOUND = false
 			message(STATUS "NumPy: Found!")
 			message(STATUS "  headers: ${Python_NumPy_INCLUDE_DIRS} (version ${Python_NumPy_VERSION})")
-		endif()
+		endif(NOT USE_OWN_CRPROPA)
 
-		
 		#  use Python_INSTALL_PACKAGE_DIR if provided; otherwise, install in Python_SITELIB
 		set(Python_INSTALL_PACKAGE_DIR "${Python_SITELIB}" CACHE PATH "folder in which the python package is installed")
```
JanNiklasB: `set(Python_INSTALL_PACKAGE_DIR "${Python_SITELIB}" CACHE PATH "folder in which the python package is installed")` is also set in line 106

plugin-template/CMakeLists.txt
```diff
@@ -1,171 +1,217 @@
...
+		message(STATUS "Python version: ${Python_VERSION}")
+		message(STATUS "Python executable: ${Python_EXECUTABLE}")
+		message(STATUS "Python headers: ${Python_INCLUDE_DIRS}")
+		message(STATUS "Python library: ${Python_LIBRARY}")
+		message(STATUS "Python site-packages: ${Python_SITE_PACKAGES}")
+	endif(Python_FOUND)
+
+	if(Python_NumPy_FOUND)
+		set(CMAKE_SWIG_FLAGS -DWITHNUMPY)
+		include_directories(${Python_NumPy_INCLUDE_DIRS})
```
JanNiklasB: I am not sure if we need this global include since `${Python_NumPy_INCLUDE_DIRS}` is included directly to the swig target in line 272

plugin-template/CMakeLists.txt
```diff
@@ -176,119 +222,174 @@ if(ENABLE_PYTHON)
...
+		find_library(GTEST_MAIN_LIBRARY NAMES gtest_main HINTS ${CRPropa_LIBRARY_DIR} ${CRPropa_LIBRARY_DIR}/lib)
+		set(GTEST_LIBRARIES ${GTEST_LIBRARY} ${GTEST_MAIN_LIBRARY})
+	endif(USE_OWN_CRPROPA)
+
 	# first add the cpp tests:
 	add_executable(testPluginCPP ${CMAKE_CURRENT_SOURCE_DIR}/test/testPlugin.cpp)
-	target_link_libraries(testPluginCPP ${CMAKE_PROJECT_NAME} gtest gtest_main pthread)
+	target_include_directories(testPluginCPP PRIVATE ${GTEST_INCLUDE_DIR})
+	target_link_libraries(testPluginCPP ${pluginName} ${GTEST_LIBRARIES} pthread)
 	add_test(testPluginCPP testPluginCPP)
```
JanNiklasB: Maybe we should add the new functions from the main CMakeLists.txt here as well, that way a new plugin devloper can just use these

plugin-template/CMakeLists.txt
```diff
@@ -176,119 +222,174 @@ if(ENABLE_PYTHON)
...
+	if(USE_OWN_CRPROPA)
+		# gtest is built together with CRPropa (see FetchContent above); use its targets directly
+		set(GTEST_INCLUDE_DIR ${CRPropa_SOURCE_DIR}/libs/gtest/include)
+		set(GTEST_LIBRARIES gtest gtest_main)
+	else()
+		# headers sit under the CRPropa include dir (gtest/), libraries next to (install) or under lib/ (build tree) libcrpropa
+		set(GTEST_INCLUDE_DIR ${CRPropa_INCLUDE_DIR})
+		find_library(GTEST_LIBRARY NAMES gtest HINTS ${CRPropa_LIBRARY_DIR} ${CRPropa_LIBRARY_DIR}/lib)
+		find_library(GTEST_MAIN_LIBRARY NAMES gtest_main HINTS ${CRPropa_LIBRARY_DIR} ${CRPropa_LIBRARY_DIR}/lib)
+		set(GTEST_LIBRARIES ${GTEST_LIBRARY} ${GTEST_MAIN_LIBRARY})
```
JanNiklasB: @JulienDoerner noticed that if gtest is not found it might be a good idea to install it for the use (over a fetch_content block like with crpropa). This would make the plugin compatible with older CRPropa versions.

JanNiklasB (changes requested):
Sorry for another request. I updated one of my own plugins with the new template and noticed some other things.
- Duplicates in python parts (s. comments)
- Missing compability with older CRPropa versions due to not exposing gtest, could be solved by adding a fetch_content gtest block if no gtest can be found (noticed by @JulienDoerner )

I think this is now everything since my largest plugin seems to work.

CMakeLists.txt
rafaelab: > It was caused by using absoulte paths in add_test, where somehow the function added testName after COMMAND as absolute path, it can be fixed by adding a ./ before ${testName}.

That is something I didn't anticipate.
`./${testName}` assumes that the test’s working directory is the directory containing the executable. That is often true, but not always. Perhaps doing takes away the user's ability to really choose where the tests go. 

I think we should be careful. Isn't this something that can be configured in the conda installation (e.g., by settng `WORKING_DIRECTORY`)?

JanNiklasB: Conda creates a placeholder path that should be used which is later replaced by the correct path. However, the `CTestTestfile` would not point to the placeholder path but to the binary path and is therefore not replaced by conda. That is why the current workaround was to use relative paths instead.

If we set the working directory dependent on a variable which is set in `conda-bld/build.sh` manually to the target test directory in the conda prefix we might be able to use absolute paths even in a conda package.

rafaelab:
- About the `./` for tests, I don't think it is an acceptable idea for an installation. It fixes conda and breaks custom installations. However, claude suggested me something that seems to be working: `$<TARGET_FILE:${testName}>`. Make sure it also work with conda.

- About gtest, I now replaced only the testing / gtest block according to a preference list:  prefer an existing CMake target; try CMake’s `find_package(GTest)`; then try CRPropa’s installed/bundled libraries, and only then fetch googletest.
(This was a bit tricky to do in a consistent way. If you have a better suggestion, let me know).

JanNiklasB:
`$<TARGET_FILE:${testName}>` would just find the path to the targeted binary (s. [cmake-generator-expressions](https://cmake.org/cmake/help/latest/manual/cmake-generator-expressions.7.html#target-dependent-expressions) ).
So in our case it should be no different from using `${CMAKE_CURRENT_BINARY_DIR}`.

My fix in mind would be something like:

Make the path dependent on a variable which defaults to `${CMAKE_CURRENT_BINARY_DIR}`:
```cmake
set(CRPROPA_TESTS_PATH ${CMAKE_CURRENT_BINARY_DIR} CACHE PATH "")  # rename to anything more appropriate
mark_as_advanced(CRPROPA_TESTS_PATH)
add_test(NAME ${testName} COMMAND ${CRPROPA_TESTS_PATH}/${testName})
# and for python
add_test(NAME ${testName} COMMAND ${Python_EXECUTABLE} ${CRPROPA_TESTS_PATH}/${testName}.py)
```
then change the variable manually in `conda-bld/build.sh` (last line):
```sh
cmake .. -G Ninja \
	-DCMAKE_PREFIX_PATH="${PREFIX}" \
	-DPython_EXECUTABLE="${PYTHON}" \
	-DPython_NumPy_INCLUDE_DIR="${NUMPY_INCLUDE_DIR}" \
	-DPython_INCLUDE_DIR="${PYTHON_INCLUDE_DIR}" \
	-DPython_INSTALL_PACKAGE_DIR="${SP_DIR}" \
	-DCMAKE_INSTALL_PREFIX="${PREFIX}" \
	-DBUILD_DOC=OFF \
	-DDOWNLOAD_DATA=OFF \
	-DENABLE_COVERAGE=OFF \
	-DENABLE_GIT=ON \
	-DENABLE_HDF5=ON \
	-DENABLE_OPENMP=ON \
	-DENABLE_PYTHON=ON \
	-DENABLE_QUIMBY=OFF \
	-DENABLE_SWIG_BUILTIN=ON \
	-DENABLE_TESTING=ON \
	-DFAST_WAVES="${FAST_WAVES}" \
	-DOMP_SCHEDULE=dynamic \
	-DSIMD_EXTENSIONS="${SIMD_EXTENSIONS}" \
	-DUSE_ABSOLUTE_RPATH=ON \
	-DCRPROPA_TESTS_PATH="${PREFIX}/share/crpropa/test/"
```

rafaelab:
I think this version is cleaner, @JanNiklasB .
The conda build seems to work.

The only thing I'm unsure about is the plugin's CMakeLists correctly finding gtests for older CRPropa versions.
I tried a few different things, some worked, some didn't, and the one that hasn't given me problems is the current one.

plugin-template/CMakeLists.txt
```diff
@@ -1,294 +1,400 @@
...
+		else()
+			include(FetchContent)
+			FetchContent_Declare(
+				googletest
+				GIT_REPOSITORY https://github.com/google/googletest.git
+				# GIT_TAG v1.14.0
+			)
+			FetchContent_MakeAvailable(googletest)
+			set(${pluginName}_GTEST_LIBRARIES gtest gtest_main)
+		endif(TARGET GTest::gtest AND TARGET GTest::gtest_main)
```
JanNiklasB: Beginning from line 310: I think `${pluginName}_GTEST_INCLUDE_DIRS` is never defined here.

plugin-template/CMakeLists.txt
```diff
@@ -1,294 +1,400 @@
...
+			file(WRITE "${CRPropa_PYTHON_TEST_PACKAGE_DIR}/__init__.py" "from .crpropa import *\n")
+			add_test(
+				NAME moveCRPropaPlugin
+				COMMAND ${CMAKE_COMMAND} -E copy
+					"${CRPropa_BINARY_DIR}/crpropa.py"
+					"$<TARGET_FILE:crpropa-swig>"
+					"$<TARGET_FILE:crpropa>"
+					"${CRPropa_PYTHON_TEST_PACKAGE_DIR}"
+			)
+			set_tests_properties(testPluginPy PROPERTIES DEPENDS moveCRPropaPlugin)
```
JanNiklasB: The dependency on the plugin files should be added inside the addTestPy function I think, then it would be ensured that every python test has access to the crpropa python package

JanNiklasB (commented):
I think your way of finding gtest should in principle work in any case, only the includes might not be set every time (s. comment).

Furthermore, the dependencies of the python tests on the test that copies the crpropa python files should probably be placed inside the addtestPy function so that the dependency is set for all potential tests.

plugin-template/CMakeLists.txt
```diff
@@ -1,294 +1,408 @@
...
+
+	# Reuse CRPropa's own gtest if at all possible; only if that fails, look for the system's gtest (or download it)
+	set(${pluginName}_GTEST_LIBRARIES)
+	set(${pluginName}_GTEST_INCLUDE_DIRS)
+
+	# message(STATUS "======== ${CRPropa_LIBRARY_DIR} ========")
+
+	# install CRPropa together with plugin
+	if(USE_OWN_CRPROPA)
+		set(${pluginName}_GTEST_INCLUDE_DIRS "${CRPropa_INCLUDE_DIR}")
```
JanNiklasB:
```suggestion
		set(${pluginName}_GTEST_INCLUDE_DIRS "${crpropa_SOURCE_DIR}/libs/gtest/include")
```

plugin-template/CMakeLists.txt
```diff
@@ -1,294 +1,408 @@
...
+		set(${pluginName}_GTEST_LIBRARIES "${CRPropa_LIBRARY_DIR}/libgtest.a")
```
JanNiklasB:
```suggestion
		set(${pluginName}_GTEST_LIBRARIES gtest gtest_main)
```

JanNiklasB (commented):
The tests are only failing because of typos in line 302 and 303 (see suggestions)
Other then that it now looks good to me.

plugin-template/CMakeLists.txt
rafaelab: If the system has gtest, then there might be version conflicts. This must be explicit.

rafaelab:
I'm still working on my last attempt, @JanNiklasB .
If it doesn't work, I won't try to make gtest work with older CRPropa versions.
I will let you know when it's done

rafaelab:
This version (current) is almost my last attempt, @JanNiklasB. I will just fix one more thing based on the outcome of our discussion.

As you can see, one of the tests fail because of the pybind11-stubgen (in OSX only).  I had originally removed all references to pybind from CMakeLists.txt because it does not pertain to CRPropa (at the moment). In my opinion, it should not even have  been included in the cmake file. 
It is one more thing to support and it makes no difference for the science output of most users.

However, since it is already there, it can be offered as an option. Naturally, it should only be activated by the users who really want it and be kept off by default. That can be done with a cmake flag `ENABLE_PYBIND11_STUBGEN`.  
It keeps the code clean from a feature that is not needed by most users and developers.

The question then becomes: should it be part of the conda release? To me this is a clear no, especially because it is making a test fail. It will just increase the size of the package and increase the runtime of the workflows.
If it is removed from meta.yaml, the conda package will likely work just fine. 

The way it currently is, it is acting as a strong dependence and I really don't see the practical advantage of having it in the default installation and conda release.


I am adding this to the PR, since it is needed to make the tests pass, but I'd like to hear from you first.

JanNiklasB:
It is not failing because of `pybind11-stubgen`, what is failing is the python import. This is caused by your removal of
```
  if(APPLE)
    set_target_properties(crpropa-swig PROPERTIES
      INSTALL_RPATH "@loader_path;@loader_path/../../${PYTHON_PROJECT}/.libs"
    )
    set_property(TARGET crpropa-swig APPEND PROPERTY
      LINK_FLAGS "-flat_namespace -undefined suppress"
    )
  endif()
```
(Compare #548 )

`pybind11-stubgen` is not a requirement we do but an option since we do not built any stubs when it is not available.
In the conda package the stubs take only minimal space (~ 600KiB in raw form, much less in compressed package form, compare that to 6.3 MiB for our required python files and the 3MiB total size of our package (header and libraries not yet included)).

It enables the linting which is typically done by pylance and makes our documentation available without looking it up per `help()` function or by searching through the manual, so even if it is not scientific it is still very useful for most users.

rafaelab:
Hm... Then I misunderstood the intent. 
I thought the goal was more to prepare for your own Pybind extensions rather than for improved documentation.
In this case, you are right, it is can be very useful.

In any case, I will add it as optional, but keep the default to `On`.

JanNiklasB:
I see why it seemed that it is needed for pybind11 extensions (it's in the name after all). Since swig does not generate proper modern stub files I searched for a third party alternative but it seems pybind11-stubgen is also the best option for swig...

I like adding a option, I was hesitant with that since I did not want to create the implication that we would require `pybind11-stubgen` for CRPropa to work, but we also do not require `fftw` and `hdf5` for CRPropa to work and still have the option.

Was that the last thing you wanted to add? If so I would approve it from my side.

rafaelab:
Yes, I got confused because I know you were working on pybind wrappers, so I thought you wanted to add something in the current version to prepare for that. Sorry about that.
I think an option set to true by default is harmless and gives more control to the users.

From my side, I'm done with this PR if all tests pass.

Thanks for the review, @JanNiklasB!

JanNiklasB (approved):

PR #565, opened 2026-07-07
Added new test_conda_package.yml workflow

JanNiklasB:
Hi,

this PR adds a new test that only starts on PRs (since for most changes our other tests are enough) and tests if the conda packages would compile. The test can also be started manually on every fork and branch.
It compiles all conda packages without uploading them.

### DIFF
--- a/.github/workflows/test_conda_package.yml
+++ b/.github/workflows/test_conda_package.yml
@@ -8,6 +8,7 @@ on:
       - "**.py"
       - "**.i"
       - "**/CMakeLists.txt"
+      - "conda-build/*"
       - "!plugin-template/**"
       - "!doc/**"
   workflow_dispatch:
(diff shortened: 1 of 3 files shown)


.github/workflows/test_conda_package.yml
```diff
@@ -0,0 +1,120 @@
+name: test_conda_package
+on:
+  pull_request:
+    paths:
```
JulienDoerner, line 4: This does not check if we change the building for the conda package. I would also add  `conda-build/*`

.github/workflows/test_conda_package.yml
```diff
@@ -0,0 +1,120 @@
...
+      - ".github/workflows/test_conda_package.yml"
+      - "**.cpp"
+      - "**.h"
+      - "**.py"
+      - "**.i"
+      - "**/CMakeLists.txt"
+      - "!plugin-template/**"
+      - "!doc/**"
+  workflow_dispatch:
```
JulienDoerner, line 14: Maybe you could add an option to select whether we want to upload the binary of the conda package as an upload artifact. In this case, the user can build their own conda package and add this later to their personal conda without doing the compilation on their own machines.

JulienDoerner (commented):
Thanks @JanNiklasB, this is a good improvement. By now, we only notice problems with the package creation after the merge to the master.

JulienDoerner (approved):

PR #569, opened 2026-07-13
Fixing missing RPATH for unix swig interface files

### LINKED ISSUE #568: Missing RPATH for `_crpropa.so` on unix systems
When installing CRPropa in a custom directory that is not included in any path variable the swig interface file is not able to find the install path.

This is caused by never handing that path over to the swig library.

The fix is to add 
```cmake
set_target_properties(crpropa-swig PROPERTIES
   BUILD_RPATH "$<TARGET_FILE_DIR:crpropa>"
   INSTALL_RPATH "${CMAKE_INSTALL_FULL_LIBDIR}"
)
```
to the `CMakeLists.txt` file to guarantee that `libcrpropa.so` can be found.

An example for a failing plugin compilation with custom CRPropa install directory can be seen [here](https://github.com/JanNiklasB/CRPropa3/actions/runs/29270506577/job/86886715057)

JanNiklasB:
Hi,

this PR attempts to fix #568 by adding the explicit RPATH to the swig interface.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -532,6 +532,11 @@ if(ENABLE_PYTHON AND Python_FOUND)
       INSTALL_RPATH "@loader_path;${CMAKE_INSTALL_FULL_LIBDIR}"
     )
     set_property(TARGET crpropa-swig APPEND PROPERTY LINK_FLAGS "-flat_namespace -undefined suppress")
+  else()
+    set_target_properties(crpropa-swig PROPERTIES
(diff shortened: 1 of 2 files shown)


JulienDoerner:
You use two different RPath variables in the plugin-template (`"${CMAKE_INSTALL_PREFIX}/lib"`) and the main CMakeLists (`"${CMAKE_INSTALL_FULL_LIBDIR}"`). 

Is there a special reason why they should differ? Otherwise, I would suggest making it consistent.

JulienDoerner (approved):

PR #574, opened 2026-07-22
Explicit Time Steps

JanNiklasB:
Hi,

this is the first of two connected PRs to introduce explicit time integration and velocities smaller then c.

# Motivation

This PR starts with changing the lengths step integration to time step integration.
Untill now, we used length steps to propagate out particles, which is equivalent to time steps when assuming a constant velocity, in out case the vacuum lightspeed c.
Since I plan to introduce variable velocities in the second PR those two quantities are not the same anymore, and to avoid dividing by v=0 at any given time to convert to time steps, we will need to track the time steps instead and convert to length steps by multiplying the time step with the velocity instead.

# Changes

- Changed all `*CurrentStep` and `*NextStep` calls so time steps are handed over
- Changed doxygen comments from length step to time step where it is appropriate
- Added overloads for constructors of propagators so the default is still length steps and the time steps have a different order of arguments. Lengths steps are converted in the constructor to time steps by dividing by `c_light`
- Added `MaximumTime` break condition
- Added `ObserverSpacialEvolution` to observer length steps
- Changed `ObserverTimeEvolution` so it accepts time steps (this and the previous point can be a Problem since they break the interface, I think it is more intuitively to have the `ObserverTimeEvolution` take times by default, but we could preserve the old behavior of using length steps.
- Added some new tests and adjusted old ones

## Example Notebooks


JanNiklasB:

JanNiklasB:
I am now able to reopen this, sorry for the disruption

PR #575, opened 2026-07-22
Change `ParticleState::getVelocity()` to represent velocities smaller `c_light`

JanNiklasB:
Hi,

this is the second of two connected PRs to introduce explicit time integration and velocities smaller then c.

# Motivation

This PR continues based on #574 with the introduction of velocities smaller then `c_light`.
Velocities smaller then `c_light` will enable us to do better tracking of the distance the particle travels together with the the in #574 introduced explicit time steps. It will also enable us to in the future introduce modules that utilize those velocities, like Propagators that account for electric fields and can perform rest acceleration.

# Low Velocities and other Modules

One point that is and was always concerning when thinking about introducing velocities smaller then `c_light` is that we might suggest all CRPropa modules would work at any given speed, even though most modules assume relativistic speeds to work properly.
I think that is not a new problem we are facing, since you were always able to propagate particles in low energy regimes without any warning what so ever and would still violate the relativistic assumptions those module make.
To now accurately represent the velocity as smaller then `c_light` only makes it more noticeable, but not introduce a new Problem.
Since it is now more noticeable we could throw a warning as soon as someone wants to use any module with small enough velocities.

I would hope that the possibility to use lower velocities motivates people to expand the existing module that assume the relativistic limit to also work at low energies, until then, I would require our users to think about the energy ranges the modules are applicable for as they are required to do now.

# Changes

- Particle state functions now consider velocities lower then `c_light`, especially `getVelocity`
- Adjusted propagators to account for v=0
- Added some tests for low velocities

## Example Notebooks


JanNiklasB:

JanNiklasB:
I am now able to reopen this PR, sorry for the disruption

PR #576, opened 2026-07-22
Electric Fields and Boris Push

JanNiklasB:
Hi,

this is the third of three connected PRs to introduce explicit time integration, velocities smaller then c and a boris push that supports electric fields.

# Motivation

This PR introduces electric fields and expands `PropagationBP` with the ability to utilized those fields and is based on #575 and #574.

# Changes

- Expansion of `PropagationBP` with additional overloads for electric fields and preserved old behavior when not electric field is present.
- Added setter for mass and charge to be able to set partially ionized particles if needed
- Added tests for electric fields
- Added electric field base class and uniform electric field

## Example Notebooks


JanNiklasB:

PR #578, opened 2026-07-22
Fix for particle mass 

### LINKED ISSUE #577: Particle Mass not found for protons
While running our example notebooks, I noticed some unexpected warnings. 
I have reduced the problem to this minimal example 

```py 
from crpropa import * 


sim = ModuleList() 
prop = DiffusionSDE(UniformMagneticField(Vector3d(1,0,0)*muG), 1e-4, 1 * pc, 1*kpc, 1)
sim.add(prop)
sim.add(MaximumTrajectoryLength(1 * Mpc))

s = Source() 
s.add(SourceParticleType(nucleusId(1,1)))
s.add(SourceIsotropicEmission())
s.add(SourceEnergy(1 * TeV))

sim.run(s, 10_000)
```

I get the following warning via KISS: 

```
[WARNING] nuclearMass: nuclear mass not found in the mass table for A = 1, Z = 1. Approximated value used A * amu - Z * m_e instead.
```

This has been introduced with the heavy nuclei in PR #556. 
I will work on a fix, but release 3.3 is definitely broken, and we have to work on a short-term patch.

JulienDoerner:
This PR fixes #577 by adding an additional check if data are initialized to prevent the new assignment of the data table.

### DIFF
--- a/src/ParticleMass.cpp
+++ b/src/ParticleMass.cpp
@@ -22,6 +22,8 @@ struct NuclearMassTable {
 	}
 
 	void init() {
+		if(initialized)
+			return;
 		std::string filename = getDataPath("nuclear_mass.txt");
 		std::ifstream infile(filename.c_str());
 


rafaelab:
I'm impressed that such a simple fix solves the issue, but it seems to be the case.

rafaelab (approved):

Issue #587, opened 2026-08-28
Possible double redshift scaling in EMInverseComptonScattering::process()


inr-otf:
**Title:** Possible double redshift scaling in `EMInverseComptonScattering::process()`

While cross-checking an electromagnetic-cascade calculation with CRPropa 3.3.1, we noticed what appears to be a double application of the redshift scaling in `EMInverseComptonScattering`.

In the version we tested, `getRate(E, position, z)` already applies the redshift-dependent scaling factor. However, `process()` calls `getRate(...)` and then appears to multiply the resulting rate by the redshift scaling once again.

Removing the second multiplication noticeably improves an independent 100-TeV line-injection cross-check:

* CRPropa 3.2.2 / gamma-Cascade: \(1.027 \pm 0.038\)
* locally patched CRPropa 3.3.1 / gamma-Cascade: \(1.050 \pm 0.023\)

The local patch simply removes the second redshift-scaling multiplication in `EMInverseComptonScattering::process()`.

We also encountered the `secondariesFirst=True` traversal problem, but this appears to be the same issue already reported in #583, so I am not reporting it here.

Could you please check whether the second redshift scaling in `EMInverseComptonScattering::process()` is intentional?

— Sergey Troitsky

lukasmerten:
Thanks for reporting this.

An issue might have been introduced with PR #562; @rafaelab can you please double check your implementation to expose the getRate() function?

rafaelab:
Fixed with #588.

PR #588, opened 2026-08-28
fix double redshift correction; issue #587

rafaelab:
This PR fixes a small bug introduced by PR #562, solving issue #587.
The bug was a line of code scaling the interaction rate with the redshift that should have been removed with the PR.


JulienDoerner (approved):
All EM processes are now consistently implemented.

